Issue #15, opened 2019-03-05
Generalize attestation support


g2flyer:
Generalize attestation support to cover linkable and unlinkable EPID as well as DCAP/3rd-party attestation

g2flyer:
Priorities:
- short-term: support for both linking types of EPID
- mid-term: DCAP support so it will work in cloud settings

bvavala:
epid-based attestations work with DCAP.
Additional work is required to support ECDSA.

bvavala:
Moving this issue to a later project

mbrandenburger:
As SGX SDK 2.6 is required for DCAP see this comment:

(Note, though, PR #168 only updates the docker FPC CC _runtime_ image to SDK 2.6, i.e., we also have to update the docu correspondingly to capture the build time.  (although docu update we might probably best combine with dcap update as it simplifies the install procedure, see the [corresponding change in PDO](https://github.com/hyperledger-labs/private-data-objects/pull/229/files)

_Originally posted by @g2flyer in https://github.com/hyperledger-labs/fabric-private-chaincode/issues/112#issuecomment-533341498_

PR #20, opened 2019-04-01
doc on how to make it work with linkable epid credentials

g2flyer:
This is first step to address issue #15.   Right now it's only documentation. Ideally it would be set as a configuration parameter but for that we should settle first on whether to keep ercc as a normal chaincode (and figured out a way to either replace or restrict decorators to pass sensitive parameters), move back to system chaincode or integrate it even differently (e.g., like private data). Alternatively, we could also just make it a build parameter?

### DIFF
--- a/fabric/README.md
+++ b/fabric/README.md
@@ -46,7 +46,12 @@ Place your client certificate and your SPID in the ``ias`` folder.
     echo 'YOURSPID' | xxd -r -p > /path-to/fabric/sgxconfig/ias/spid.txt
 
 We currently make use of `unlinkable signatures` for the attestation, thus, when registering with the IAS please choose
-unlinkable signatures.
+unlinkable signatures.  In the case you prefer linkable attestation,
+e.g., because you already have linkable IAS EPID credentials, change
+in line 217 of [../ecc_enclave/sgxcclib/sgxcclib.c](../ecc_enclave/sgxcclib/sgxcclib.c)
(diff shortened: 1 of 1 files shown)


fabric/README.md
```diff
@@ -46,7 +46,12 @@ Place your client certificate and your SPID in the ``ias`` folder.
     echo 'YOURSPID' | xxd -r -p > /path-to/fabric/sgxconfig/ias/spid.txt
 
 We currently make use of `unlinkable signatures` for the attestation, thus, when registering with the IAS please choose
-unlinkable signatures.
+unlinkable signatures.  In the case you prefer linkable attestation,
+e.g., because you already have linkable IAS EPID credentials, change
+in line 217 of [../ecc_enclave/sgxcclib/sgxcclib.c](../ecc_enclave/sgxcclib/sgxcclib.c)
+the constant `SGX_UNLINKABLE_SIGNATURE` to `SGX_LINKABLE_SIGNATURE`,
+recompile ercc, configure your IAS settings as above with your
```
mbrandenburger: it should be ecc_enclave/ecc instead or ercc

g2flyer: @mbrandenburger  Fixed to point to ecc and ecc_enclave an added corresponding links.

PS: i noticed you added an explicit merge and also prior PRs have mostly an additional empty merge commit. In PDO we try to minimize commits by rebasing and potentially squashing commits (e.g., i did incorporate the fix now in the original commit). We might at some point discuss our strategy to commits (e.g., best-practices) to get a consistent git log?

PR #24, opened 2019-04-10
support operations behind proxy

g2flyer:


mbrandenburger:
@bvavala Can you please check this CR since you are also living behind a proxy.

PR #27, opened 2019-04-15
Update docu

mbrandenburger:
Based on the feedback from @piachristel (Christina Mueller) here some documentation improvements. Thanks Christina!

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,19 +95,26 @@ deploying and running an example chaincode.
 ## Requirements
 
 * CMake 3.5.1 or higher
-* Go 1.11 or higher
-* Install Fabric v1.4 https://github.com/hyperledger/fabric 
-* Install the Linux SGX SDK v2.4 https://github.com/intel/linux-sgx 
-* You also need to install Intel SGX SSL https://github.com/intel/intel-sgx-ssl
+* Go 1.11.x or higher
+* Hyperledger Fabric v1.4 https://github.com/hyperledger/fabric 
+
+    If you are new to Fabric, we recommend the Fabric documentation as your starting point. You should start with
+    [installing](https://hyperledger-fabric.readthedocs.io/en/release-1.4/prereqs.html) Fabric dependencies and setting
+    up your [development environment](https://hyperledger-fabric.readthedocs.io/en/release-1.4/dev-setup/build.html).
+    
+* SGX SDK v2.4 for Linux https://github.com/intel/linux-sgx 
+* SSL for SGX SDK v2.4.1 https://github.com/intel/intel-sgx-ssl (we recommend using OpenSSL 1.1.0j)
 * (for hardware-mode SGX build) credentials for IAS, read [here](fabric#intel-attestation-service-ias)
 
 ### SGX SDK and SSL
 
-The the chaincode envlave and the trusted ledger enclave require SGXSSL.  See
-README in project root fore more details. Intel SGX SSL
+The the chaincode envlave and the trusted ledger enclave require SGX SSL.  See
+README in project root for more details. Intel SGX SSL
 https://github.com/intel/intel-sgx-ssl
 
-The makefiles are configured with meaningful defaults for variables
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -95,10 +95,15 @@ deploying and running an example chaincode.
...
-* Install the Linux SGX SDK v2.4 https://github.com/intel/linux-sgx 
-* You also need to install Intel SGX SSL https://github.com/intel/intel-sgx-ssl
+* Go 1.11.x or higher
+* Hyperledger Fabric v1.4 https://github.com/hyperledger/fabric 
+
+    If you are new to Fabric, we recommend the Fabric documentation as your starting point. You should start with
+    [installing](https://hyperledger-fabric.readthedocs.io/en/release-1.4/prereqs.html) Fabric dependencies and setting
+    up your [development environment](https://hyperledger-fabric.readthedocs.io/en/release-1.4/dev-setup/build.html).
+    
+* SGX SDK v2.4 for Linux https://github.com/intel/linux-sgx
```
g2flyer: btw: you hard-coded now 2.4 in a patch for docker-sgx as that didn't seem to work anymore with 2.5. However, docker-sgx now fixed the docker file such that master with your patch does _not_ work anymore.  Maybe for now easiest would be to update docu to say one should check out commit is08c2457 for docker-sgx (bruno tested that config).  For more long-term i propose we eventually dump docker-sgx and go with our own.  We could already use the one from PDO but probably better would be to do the switch once i have a docker image for DCAP (which still supports epid) which i plan to do for PDO and try to do it also in a way that it would be easy shareable/copyable for TCC ...

mbrandenburger: Thank you. Let's create a separate issue for this problem. I also agree that we should maintain our own dockerfile and thereby remove the dependency of another project. 
Updating to SGX 2.5 is yet another TODO.
Can you approve and merge?

README.md
```diff
@@ -95,10 +95,15 @@ deploying and running an example chaincode.
...
-* You also need to install Intel SGX SSL https://github.com/intel/intel-sgx-ssl
+* Go 1.11.x or higher
+* Hyperledger Fabric v1.4 https://github.com/hyperledger/fabric 
+
+    If you are new to Fabric, we recommend the Fabric documentation as your starting point. You should start with
+    [installing](https://hyperledger-fabric.readthedocs.io/en/release-1.4/prereqs.html) Fabric dependencies and setting
+    up your [development environment](https://hyperledger-fabric.readthedocs.io/en/release-1.4/dev-setup/build.html).
+    
+* SGX SDK v2.4 for Linux https://github.com/intel/linux-sgx 
+* SSL for SGX SDK v2.4 https://github.com/intel/intel-sgx-ssl
```
g2flyer: maybe we should switch sgxssl to v2.4.1? And as discussed during call propably mention to use latest version of openssl 1.1.0 (greater-or-equal 1.1.0j)?

tlcc_enclave/README.md
```diff
@@ -11,9 +11,12 @@ the blockchain state.
...
 ## Start with generating proto parser
 
-We use *nanopb*, a lightweight implementation of Protocol Buffers, inside the
-ledger enclave to parse blocks. Install nanopb by following the instruction on
-http://github.com/nanopb/nanopb
+We use *nanopb*, a lightweight implementation of Protocol Buffers, inside the ledger enclave to parse blocks of
+transactions. Install nanopb by following the instruction below. For more detailed information consult the official
+nanopb documentation http://github.com/nanopb/nanopb.
+
+[comment]: # (note by christina: I would rather write: "Install nanopb by the following instructions. Because one does not have to go through the whole README of
```
g2flyer: the comment seems obsolete now?  (But i agree with rewording: official README is actually more confusing but it does _not_ make obvious the actual steps you have to do ...)

mbrandenburger: ok, removed the comment

README.md
mbrandenburger: Done!

PR #33, opened 2019-04-17
fix "Public key not valid (Point not on curve)" error

bvavala:
tlcc now checks whether the local attestation (from C code) has failed and outputs an error message.
Without the check, in the case of attestation error, tlcc still passes the parameters ecc, which parses the enclave's public key and fails because the key is likely not valid.

### DIFF
--- a/tlcc/enclave/enclave_stub.go
+++ b/tlcc/enclave/enclave_stub.go
@@ -19,6 +19,8 @@ package enclave
 import (
 	"unsafe"
 
+    "fmt"
+
 	"github.com/hyperledger/fabric/common/flogging"
 	"github.com/hyperledger-labs/fabric-secure-chaincode/ecc/crypto"
 )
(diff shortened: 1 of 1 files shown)


tlcc/enclave/enclave_stub.go
```diff
@@ -19,6 +19,8 @@ package enclave
 import (
 	"unsafe"
 
+    "fmt"
+
```
mbrandenburger: no empty line

mbrandenburger: can you run goimport please

tlcc/enclave/enclave_stub.go
```diff
@@ -101,11 +103,14 @@ func (e *StubImpl) GetLocalAttestationReport(targetInfo []byte) ([]byte, []byte,
 	defer C.free(targetInfoPtr)
 
 	// call enclave
-	// TODO read error
-	C.tlcc_get_local_attestation_report(e.eid,
+    ret := C.tlcc_get_local_attestation_report(e.eid,
```
mbrandenburger: formatting; run gofmt to fix that

tlcc/enclave/enclave_stub.go
```diff
@@ -101,11 +103,14 @@ func (e *StubImpl) GetLocalAttestationReport(targetInfo []byte) ([]byte, []byte,
...
 		(*C.target_info_t)(targetInfoPtr),
 		(*C.report_t)(reportPtr),
 		(*C.ec256_public_t)(pubkeyPtr))
+    if ret != 0 { // 0 is SGX_SUCCESS
```
mbrandenburger: formatting; run gofmt to fix that

bvavala:
replaced by #40

PR #40, opened 2019-04-19
fix "Public key not valid (Point not on curve)" error

bvavala:


bvavala:
I don't get your comment.
The enclave's logger (I think you refer to PERR/golog) is not used here.
The return statement simply creates the error message which [tlcc#77](https://github.com/hyperledger-labs/fabric-secure-chaincode/blob/d446ec4dc2f2fe063d92e2cef8113c4727ba8976/tlcc/tlcc.go#L77) expects and logs/dumps as wanted.

mbrandenburger:
in `enclave_stub.go` there is just one logger instance called `tl-enclave`. See code
```
var logger = flogging.MustGetLogger("tl-enclave")
```

So far this logger is used exclusively by the method `golog` to print enclave logging statements. 
```
//export golog
func golog(str *C.char) {
```
`enclave_stub.go` is otherwise log free and all errors are just returned.

I suggest that we don't use the tl-enclave logger to perform any other logging in `enclave_stub.go` as this may result in difficulties reading the logs later. Therefore, I also suggest that we use the `tlcc` logger in `tlcc.go` to output this error before we return an error to the client (as shim error).

I guess the idea behind his wired `tl-enclave` logging construct is not ideal and may be improved.

bvavala:
> This looks good to me. Just a minor. Let's not use the logger here as it is used for the enclave logging only; just return the error is sufficient here. If you want to add an error log do so in `tlcc.go`
> 
> ```
> return nil, nil, fmt.Errorf("error requesting local attestation from tlcc")
> ```
> 
> What do you think?

Alright, now I see you were referring to the "previous" line.
Fixed!

mbrandenburger:
Awesome thanks!

PR #41, opened 2019-04-22
Fix return statements in auction chaincode

### LINKED ISSUE #23: Inappropriate return statement in auction demo example
In my opinion, some return statements in the fabric-secure-chaincode/ecc_enclave/enclave/auction/auction_cc.cpp  are not correct. For example in the following section of the `auction_submit` method (line 115-118):
```
 if (auction_bytes_len == 0) {
        LOG_DEBUG("AuctionCC: Auction does not exist");
        return AUCTION_ALREADY_EXISTS; 
}
```
Shouldn't it return a AUCTION_DOES_NOT_EXIST? Which would imply the definition of an appropriate #define alias.

piachristel:
Signed-off-by: Christina Müller <muellerchristina@bluewin.ch>
fixes #23

### DIFF
--- a/ecc_enclave/enclave/auction/auction_cc.cpp
+++ b/ecc_enclave/enclave/auction/auction_cc.cpp
@@ -26,8 +26,10 @@
 #define OK "OK"
 #define AUCTION_DRAW "DRAW"
 #define AUCTION_NO_BIDS "NO_BIDS"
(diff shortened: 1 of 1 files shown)


piachristel:
@mbrandenburger Where shall I add the description and the link?
Because the link is already contained in the very first comment on the top of this history?

mbrandenburger:
you are right! everything is here in the PR history! :D merged

PR #46, opened 2019-05-04
Fixing typos

mbehlendorf:
Just fixed some small typos which i noticed while working with your project.


mbrandenburger:
Thanks! Glad for every fixed typo! I guess there many more to be found! :)

Issue #54, opened 2019-05-09
Refactor NanoPB dependencies


mbrandenburger:
make NANOPB include directly from external source (via $NANOPB_PATH so we don't need any cp (or related, currently missing, clean target)?

update READMEs to match above changes ..

this issue was part of #38

mbrandenburger:
This is addressed by #52

mbrandenburger:
refactored with #52

Issue #38, opened 2019-04-18
Global Makefile


g2flyer:
- [x] build system with a top-level make and a make file in each module. More details in #56 
  - build
  - test
  - check
  - integration (end-to-end-test (just calling run-auction initially?))
  - clean

- [x] config.mk file with global variables like
  - GO
  - GOFLAGS

- [x] code checking target. More details in #55 

- [x] refactoring of ecc (SDKization)
  - renaming of component names from trusted to secure, e.g.,  tlcc -> slcc
  - ecc only includes SDK (with a library)
  - auction goes to examples/auction
    - fabric/sgxconfig might also move to examples?
      - we could also replace all /path/to/fabric by $(FABRIC_SRC_ROOT)/.. with FABRIC_SRC_ROOT as $(SCC_ROOT)/../../hyperledger/fabric/ (ie., assume a single root in GOPATH and then all paths to fabric can be done in relative paths ..
    => delegated to issue #69 

- [x] Nanopb refactoring. More details in #55 

- [x] update READMEs to match above changes ..

PR #58, opened 2019-05-13
Nanopb

mbrandenburger:
This PR addresses #54 and is also part of #38.

This PR is based on #52 and should only be merged once its parent is done.


mbrandenburger:
Thanks, I will squash 'Refactor NanoPB dependencies'  into #52.

Issue #55, opened 2019-05-09
Add code check target to build system


mbrandenburger:
formatting check, referring to golinter from fabric; potentially separate comment which auto-indents potentially also clangformat related target for C code

update READMEs to match above changes ..

This is needed for CI as well.

This links back to #38 

- [x] code check for GO

- [x] code check for c/cpp #76 

- [ ] spellchecker (optional)

- [x] license #73

mbrandenburger:
#61 adds code checks for go.

mbrandenburger:
#84 adds code checks for cpp

PR #61, opened 2019-05-13
Code checks

mbrandenburger:
This addresses #55 and adds Fabric's go linter to the project. 

Moreover, proper code checks for c/cpp components are needed as well. @bvavala suggested using clang-format for this as PDO also uses it. I think we should adopt the coding c/cpp coding standards used in PDO as this would exchanging code among much cleaner. WDYT?

Note that another PR is required to fix the issues detected by the checks, that is, when running `make check` on the current code it may fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,15 @@
+TOP = .
+include $(TOP)/build.mk
+
 SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc
 
-all clean test:
+all build test clean :
 	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@;)
+
+checks: linter
+
+linter:
+	@echo "LINT: Running code checks.."
+	@./scripts/golinter.sh
+	@./scripts/cpplinter.sh
+
--- a/build.mk
+++ b/build.mk
@@ -4,11 +4,11 @@ include $(TOP)/config.mk
 -include $(TOP)/config.override.mk
 
 .PHONY: all
-#all: build check test integration
-all: build test
+#all: build checks test integration
+all: build checks test
 
 .PHONY: build
-#.PHONY: check
+.PHONY: checks
 .PHONY: test
 #.PHONY: integration
 .PHONY: clean
--- a/ecc/crypto/ecdsa.go
+++ b/ecc/crypto/ecdsa.go
@@ -113,7 +113,7 @@ func EnclavePk2ECDSAPK(input []byte) (*ecdsa.PublicKey, error) {
 		return nil, fmt.Errorf("Public key not valid (Point not on curve)")
 	}
 
-	return &ecdsa.PublicKey{curve, x, y}, nil
+	return &ecdsa.PublicKey{Curve: curve, X: x, Y: y}, nil
 }
 
 // MarshallEnclavePk converts sgx format Big endian to DER-encoded PKIX format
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -2,3 +2,11 @@ SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc
 
 all clean test:
 	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@;)
+
+checks: linter
+
+linter:
+	@echo "LINT: Running code checks.."
+	@./scripts/golinter.sh
```
g2flyer: Given that we realistically require fabric source already checked out, couldn't we just reference the scripts from the fabric distro. Or did you have to change the scripts so you have to include them here?   If not, we can just reference them via ../../hyperledger/fabric.

Apropos fabric dependencies: Unless we make FABRIC_PATH environment variable mandatory, we should also change tlcc_enclave/generate_protos.sh correspondingly along the lines of below patch

```
diff --git a/tlcc_enclave/generate_protos.sh b/tlcc_enclave/generate_protos.sh
index 3d6e40f..7de9fe9 100755
--- a/tlcc_enclave/generate_protos.sh
+++ b/tlcc_enclave/generate_protos.sh
@@ -2,7 +2,7 @@

 # set -eux

-FABRIC_PATH=${FABRIC_PATH-~/fabric}
+FABRIC_PATH=${FABRIC_PATH-../../../hyperledger/fabric}
 NANOPB_PATH=${NANOPB_PATH-~/nanopb}

 PROTOC_OPTS="--plugin=protoc-gen-nanopb=$NANOPB_PATH/generator/protoc-gen-nanopb"
```

make checks currently fails for me with the following. Is this expected? (I guess that's what #62 supposedly fixes?)
```
> LINT: Running code checks..
> Checking with gofmt
> Checking with goimports
> Checking for golang.org/x/net/context
> Checking with go vet
> # github.com/hyperledger-labs/fabric-private-chaincode/ecc
> ecc/enclave_chaincode_test.go:223:13: result of fmt.Errorf call not used
> ecc/enclave_chaincode_test.go:228:13: result of fmt.Errorf call not used
> ecc/enclave_chaincode_test.go:235:13: result of fmt.Errorf call not used
> Makefile:9: recipe for target 'linter' failed
```

Lastly, goimports is not part of go tools, so we probably should also mention in README that you have to do a .'go get golang.org/x/tools/cmd/goimports'?

build.mk
```diff
@@ -4,11 +4,11 @@ include $(TOP)/config.mk
 -include $(TOP)/config.override.mk
 
 .PHONY: all
-#all: build check test integration
-all: build test
+#all: build checks test integration
+all: build checks test
```
g2flyer: This made me realize that you actually do _not_ include build.mk from the top level.  This means that on the top-level 'checks' is _not_ executed by when you just run 'make' or 'make all'.   Also the build target is currently not recursively triggered.  So i suggest we add also following patch:

```
diff --git a/Makefile b/Makefile
index 809dd37..317636e 100644
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
+TOP = .
+include $(TOP)/build.mk
+
 SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc

-all clean test:
+all build clean test:
        $(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@;)

 checks: linter

```

mbrandenburger: thanks! added

Makefile
mbrandenburger: 1. Yes, we could indeed try to reuse `golinter.sh` from Fabric and call it directly. However, if we maintain or own we can easily change it according to our needs. I don't have a strong opinion here. This just works :)

2. yes, code check fails at the moment. #62 fixes that. Looking forward to the c/cpp code checks.

3. I added it to the linter script.

PR #70, opened 2019-05-14
fix make docker step

bvavala:
This fixes PR #52 -- thanks Michael for suggesting it.
The 'make docker' fails due to the symbolic links in ecc/enclave.
Additionally, the 'make debug' contains an incorrect reference to an environment variable.

### DIFF
--- a/ecc/Makefile
+++ b/ecc/Makefile
@@ -43,7 +43,7 @@ stress:
 .PHONY: debug
 debug:
 	$(LD_LIBRARY_PATH) $(GO) test -c
-	$(LD_LIBRARY_PATH) sgx-gdb ecc.test -d $GOROOT
+	$(LD_LIBRARY_PATH) sgx-gdb ecc.test -d $(GOROOT)
 
 clean: docker-clean
 	$(GO) clean
@@ -52,7 +52,7 @@ clean: docker-clean
 #	rm enclave/mrenclave.go
 
 docker:
-	$(DOCKER) build -t $(DOCKER_IMAGE) .
+	$(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..
 
 docker-run:
 	$(DOCKER) run \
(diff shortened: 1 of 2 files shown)


ecc/Makefile
```diff
@@ -52,7 +52,7 @@ clean: docker-clean
 #	rm enclave/mrenclave.go
 
 docker:
-	$(DOCKER) build -t $(DOCKER_IMAGE) .
+	$(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..
```
g2flyer: this target will fail after a clean as DOCKER-IMAGE is empty. I suggest adding below patch which help to quickly diagnose the failure:
```
diff --git a/ecc/Makefile b/ecc/Makefile
index 7faef50..d2c0700 100644
--- a/ecc/Makefile
+++ b/ecc/Makefile
@@ -52,6 +52,7 @@ clean: docker-clean
 #      rm enclave/mrenclave.go

 docker:
+       @if [ -z "${DOCKER_IMAGE}" ]; then echo "ERROR: you have to run_action.sh first before invoking this target"; exit 1; fi
        $(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..

 docker-run:
```

bvavala: I have 7 other fixes for the clean part.
Happy to list this as the 8th fix for the next PR.

g2flyer: > I have 7 other fixes for the clean part.
> Happy to list this as the 8th fix for the next PR.

Clean is not the only cause, initial build of course is another one.  But if it's easier to add it to another PR i'm fine as well, just seemed like a good patch to eventually get in

bvavala: This is not executed in the initial build (I think!). 
You mean if you run it without first making Fabric create the container?

g2flyer: i think this target is (should) never automatically triggered by another target (only place i expect is as part of first version integration test).   So, yes, your second line was the scenario i was the referring to. But as said, i'm fine if we add it to a separate PR :-)

mbrandenburger:
Thanks bruno!

PR #73, opened 2019-05-15
Add missing license identifier

### LINKED ISSUE #72: Add missing license identifier
some source files are missing license identifiers.

I propose to follow the Fabric way and use SPDX licence identifiers. They are SHORT, easy to use, machine-readable and language neutral.

Here an example
```
#
```

Moreover, we can add fabric's license checking mechanism to our checks target.

mbrandenburger:
This fixes #72 and adds missing license headers.
It also adds Fabric's license check to the
make checks target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
 TOP = .
 include $(TOP)/build.mk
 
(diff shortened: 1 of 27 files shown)


Makefile
```diff
@@ -6,7 +10,11 @@ SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc
 all build test clean :
 	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@;)
 
-checks: linter
+checks: license linter
+
+license:
+	@echo "License: Running licence checks.."
+	@sh ${GOPATH}/src/github.com/hyperledger/fabric/scripts/check_license.sh
```
g2flyer: Calling it via ${TOP}/../../hyperledger/fabric/scripts/check_license.sh seem a bit more robust (i.e., it will also work when you have multiple GOPATH components)?  But i have no strong opinion on this ...

mbrandenburger:
Yes, I guess there are also some missing Intel copyrights already; can you do a pass and add a commit here or reuse mine and also sign-off?! Let's not over complicate it. @bvavala Can you do this?

bvavala:
Nice check.
Agreed, let's not over complicate. We'll update the header as we touch the files.

Issue #71, opened 2019-05-14
cleanup of 'make clean'


bvavala:
-ercc force docker remove
-force rm in ecc_enclave
-force rm sgxcclib
-ecc docker stop check container exists
-docker rmi fails (requires 1 argument)
-tlcc rm enclave/ ... force it
-tlcc rm trustedledger ... force it
-check docker_image in ecc (patch below)
diff --git a/ecc/Makefile b/ecc/Makefile
index 7faef50..d2c0700 100644
--- a/ecc/Makefile
+++ b/ecc/Makefile
@@ -52,6 +52,7 @@ clean: docker-clean
 #      rm enclave/mrenclave.go

 docker:
+       @if [ -z "${DOCKER_IMAGE}" ]; then echo "ERROR: you have to run_action.sh first before invoking this target"; exit 1; fi
        $(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..

 docker-run:

mbrandenburger:
I dont think we should output an error saying one needs to run the auction script first. A more generic error message would be better. Like "ERROR: no docker image for ${ECC_NAME} found."

mbrandenburger:
Is this complete with PR #75 ? Can we close it?

bvavala:
Yes -- completed in PR #75

PR #75, opened 2019-05-16
Fix for PR#73 and minor clean up

bvavala:
-conditional docker command executions
-force file rm
-typos in commands
It addresses #71

### DIFF
--- a/ecc/Makefile
+++ b/ecc/Makefile
@@ -6,13 +6,11 @@ TOP = ..
 include $(TOP)/build.mk
 
 LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:./enclave/lib
-PEER_NAME ?= dev-jdoe
 CC_NAME ?= ecc
 
 VSCC_SRC = vscc/ecc_validation_plugin.go vscc/ecc_validation_logic.go
 VSCC_OUT = ecc-vscc.so
 
-DOCKER_CONTAINER = $(PEER_NAME)-$(CC_NAME)-0
 DOCKER_IMAGE ?= $$(docker images | grep -- -$(CC_NAME)- | awk '{print $$1;}')
 DOCKER_CONTAINER_ID?=$$(docker ps | grep -- -$(CC_NAME)- | awk '{print $$1;}')
 
@@ -59,9 +57,10 @@ clean: docker-clean
 
 docker:
 	@if [ -z "${DOCKER_IMAGE}" ]; then\
-		echo "\033[0;31mERROR: no $(CC_NAME) docker image found\033[0m";\
+		echo "\033[0;31mERROR: cannot override $(CC_NAME) docker image - not found\033[0m";\
 		exit 1;\
 	fi
+	echo "\033[1;33mWARNING: overriding $(CC_NAME) docker image\033[0m"
 	$(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..
 
 docker-run:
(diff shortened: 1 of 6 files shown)


tlcc_enclave/generate_mrenclave.sh
```diff
@@ -13,7 +13,7 @@ tmp_name=tmp_enc_hash
 go_name=mrenclave.go
 
 cd $build_dir
-sgx_sign gendata -enclave enclave.so -config $enclave_dir/Enclave.config.xml -out $tmp_name
+sgx_sign gendata -enclave enclave.so -config $enclave_dir/enclave.config.xml -out $tmp_name
```
g2flyer: Good catch, but the bigger problem seems that we do no catch any errors and so it was happily running so far (just producing only an empty mrenclave).   Easiest fix for that seems to be o just add a 'set -e' in the beginning?

Also, given that the generate_mrenclave.sh is the same in ecc_enclave and tlcc_enclave, maybe we should move that to a common bin directory? 

Thinking of it, the method here also seems rather brittle, the way we do it in pdo is more robust:
- add to sgx_sign sign the '-dumpfile "${SIGNED_ENCLAVE_METADATA}"' option and then
- extract it with a more understandable like the ' VAR_MRENCLAVE=$$(perl -0777 -ne 'if (/metadata->enclave_css.body.enclave_hash.m:([a-fx0-9 \n]+)/) { $$eh = $$1; $$eh=~s/0x| |\n//g; $$eh=~tr/a-z/A-Z/; print "$${eh}\n"; }' $(CONTRACT_ENCLAVE_MRENCLAVE_META_FILE))' in target '$(CONTRACT_ENCLAVE_MRENCLAVE_C_FILE)' in pservice/Makefile

mbrandenburger: This deserves a separate PR! I created an issue for that #77

ecc/Makefile
```diff
@@ -72,9 +78,9 @@ docker-run:
 		chaincode
 
 docker-stop:
-	-docker stop $(DOCKER_CONTAINER)
-	-docker rm $(DOCKER_CONTAINER)
+	-if [ ! -z "$(DOCKER_CONTAINER_ID)" ]; then docker stop $(DOCKER_CONTAINER); fi
```
mbrandenburger: this should be `docker stop $(DOCKER_CONTAINER_ID);`, right?

also DOCKER_CONTAINER is not really useful anymore, I think DOCKER_CONTAINER_ID is much better.

I suggest let's remove DOCKER_CONTAINER and in docker-run target we just use CC_NAME.

ecc/Makefile
```diff
@@ -56,6 +58,10 @@ clean: docker-clean
 #	rm enclave/mrenclave.go
 
 docker:
+	@if [ -z "${DOCKER_IMAGE}" ]; then\
+		echo "\033[0;31mERROR: no $(CC_NAME) docker image found\033[0m";\
```
mbrandenburger: This error message seems to be not super intuitive. As `make docker` creates a docker image for CC_NAME, the error tells the user that here is no such image. mhhh.

I suggest the following.

If there is no ercc docker image yet, output a warning (maybe in yellow if you want) and create the image with some default name which can be set by the user.

If the docker image exists, maybe also output a warning that we override an existing image.

WDYT?

ecc/Makefile
bvavala: agreed

ecc/Makefile
bvavala: I changed the message.
For the rest I am not sure what the "current" use of that would be. I would postpone it to a separate PR that addresses this.

mbrandenburger: ok for me

mbrandenburger:
Thanks Bruno!

Issue #68, opened 2019-05-14
Unit test cleanup


g2flyer:

mbrandenburger:
- ERCC attestation test is failing
- ERCC attestation test should mock connection to real IAS
- ERCC attestation tests need some more love; however, this package will be factored anyway, so we address full coverage testing later

mbrandenburger:
ECC_ENCLAVE no test at all :/ tests needed for ...
- enclave/shim.cpp
- enclave/enclave.cpp 
- enclave/crypto.cpp
- sgxcclib/sgxcclib.c

enclave/auction will be moved somewhere with #69.

mbrandenburger:
ECC need some more tests ...

- crypto/ecdsa.go 
- vscc/ecc_validation_logic.go

Moreover, `enclave_chaincode_test.go` shoudl also use a mock for enclave; this also allows to test ecc logic without having the enclave binaries compiled yet. However, `enclave_stub_test.go` should test with an actual enclave binary including. This may make tests in ecc_enclave redundant. @g2flyer WDYT?

mbrandenburger:
Utils has no tests :/

mbrandenburger:
tlcc_enclave is missing many chances to have unit tests ... 

There is a test which loads a set of blocks (created by an ordering service) and processes them. However, the test is not triggered with `make test`

mbrandenburger:
tlcc comes with basic unit tests.

`enclave/enclave_stub_test.go` is not invoked yet with test target.

mbrandenburger:
closed with #78

PR #78, opened 2019-05-17
Unit test cleaning

mbrandenburger:
This addresses #68.


ecc/enclave_chaincode_test.go
```diff
@@ -118,7 +108,8 @@ func TestEnclaveChaincode_Invoke_Auction(t *testing.T) {
 
 	th.CheckInvoke(t, stub, create_args)
 
-	executor := executor.NewConcurrent("Client", 16)
+	// note that shim.MockStub is not thread-safe! Therefore we better use a single executor here
```
g2flyer: In this context, interesting asplos'19 paper: https://blog.acolyer.org/2019/05/17/understanding-real-world-concurrency-bugs-in-go/ ...

ecc/Makefile
```diff
@@ -37,8 +37,8 @@ vscc-plugin: $(VSCC_SRC)
 	$(LD_LIBRARY_PATH) $(GO) build -o $(VSCC_OUT) -buildmode=plugin $^
 
 # TODO fix test and enable this target
```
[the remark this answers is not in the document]
mbrandenburger: ack!

ecc/enclave_chaincode_test.go
mbrandenburger: nice morning read

ercc/attestation/provider.go.template
```diff
@@ -0,0 +1,33 @@
+/*
```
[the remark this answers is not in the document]
mbrandenburger: yes, during review of the simualtion PR I did not consider that testing and developing of a component becomes cumbersome with that template approach. I still think Brunos approach, using the templates, is nice and we should use it in the way I propose here.

What are the additional checks that are missing now in mock?

I will clarify in the commit msg.

mbrandenburger:
> also get TONS (thousands!) of "ERROR [Enclave] VIOLATIONoh oh! cmac does not match!" in TestEncalveCHaincode_Invoke_auction and ..EncryptInvoke. If this is expected, we should at least put a printf before (and after given the amount of messages) saying that this is expected, also, if it is expected, do there have to be thaaaat many of them?

well ... the reason is that there is no tlcc that produces a proper cmac! :D thus verification fails. I guess the test needs some love. I will add a comment in the readme for the moment.

g2flyer:
> > also get TONS (thousands!) of "ERROR [Enclave] VIOLATIONoh oh! cmac does not match!" in TestEncalveCHaincode_Invoke_auction and ..EncryptInvoke. If this is expected, we should at least put a printf before (and after given the amount of messages) saying that this is expected, also, if it is expected, do there have to be thaaaat many of them?
> 
> well ... the reason is that there is no tlcc that produces a proper cmac! :D thus verification fails. I guess the test needs some love. I will add a comment in the readme for the moment.

Hmm, i think adding the warning to the makefile and/or test itself would be more useful than to the README

g2flyer:
if you run make test from top-level after make clean (or fresh checkout) it will fail in ecc as templates are not instantiated.  Maybe easiest could be to add a dependency test: build in build.mk?

mbrandenburger:
> Hmm, i think adding the warning to the makefile and/or test itself (e.g., at the end) would be more useful than to the README. That said, from what you say, actually a verification fails: So why does the test still succeed?

It seems that ecc_enclave does not handle verification errors correctly at the moment, instead of aborting the chaincode execution and returning with an error, the current code just outputs this error message and continues. I suggest to tackle this in a separate PR that refactors ecc_enclave and also adds proper tlcc mocks that produces cmacs. WDYT?

g2flyer:
> > Hmm, i think adding the warning to the makefile and/or test itself (e.g., at the end) would be more useful than to the README. That said, from what you say, actually a verification fails: So why does the test still succeed?
> 
> It seems that ecc_enclave does not handle verification errors correctly at the moment, instead of aborting the chaincode execution and returning with an error, the current code just outputs this error message and continues. I suggest to tackle this in a separate PR that refactors ecc_enclave and also adds proper tlcc mocks that produces cmacs. WDYT?

Sounds like a plan. Assuming that that PR will be fixed sooner than later, i guess we can also just drop the warning (in README or elsewhere)?

ecc_enclave/enclave/shim.cpp
```diff
@@ -49,7 +49,10 @@ void get_state(const char* key, uint8_t* val, uint32_t max_val_len, uint32_t* va
         sgx_sha256_msg(val, *val_len, &state_hash);
     }
 
-    if (check_cmac(key, NULL, &state_hash, &session_key, &cmac) == 0) {
+    if (check_cmac(key, NULL, &state_hash, &session_key, &cmac) != 0) {
+        LOG_ERROR("Enclave: VIOLATION!!! Oh oh! cmac does not match!");
```
bvavala: can we just remove this for now and add an issue?

ecc_enclave/enclave/shim.cpp
```diff
@@ -49,7 +49,10 @@ void get_state(const char* key, uint8_t* val, uint32_t max_val_len, uint32_t* va
         sgx_sha256_msg(val, *val_len, &state_hash);
```
bvavala: also, should we change the get_state to return an error?

mbrandenburger: I agree, eventually yes! I create a issue for that

ecc_enclave/enclave/shim.cpp
mbrandenburger: ok, let's tackle this with #83

mbrandenburger:
> Maybe easiest could be to add a dependency test: build in build.mk?
done

mbrandenburger:
WIP: still have issues when using SIM mode

g2flyer:
btw: couldn't/shouldn't we have a default definition of  FABRIC_PATH (to ${GOPATH}/src/github.com/hyperledger/fabric), e.g., in config.mk?

mbrandenburger:
> Consistently fails for me when running 'make', 'make test' after 'make clean' or 'make build' (with slight variation of vet error message). This happens though only in HW-mode, in SIM-mode it seems to work (and also has now much more sane output ... :-)
> 
> go build -o ercc-vscc.so -buildmode=plugin vscc/ercc_validation_plugin.go vscc/ercc_validation_logic.go
> go build -o ercc-decorator.so -buildmode=plugin attestation/ias_credentials/decoration.go
> go test -v ./... # github.com/hyperledger-labs/fabric-private-chaincode/ercc
> vet: ./provider.go:7:6: GetVerifier already declared through dot-import of package attestation ("github.com/hyperledger-labs/fabric-private-chaincode/e
> rcc/attestation") FAIL github.com/hyperledger-labs/fabric-private-chaincode/ercc [build failed] === RUN TestAttestation Running Suite: Attestation Suite ================================ Random Seed: 1560876147 Will run 6 of 6 specs ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒ Ran 6 of 6 Specs in 0.015 seconds SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 0 Skipped --- PASS: TestAttestation (0.02s) PASS ok github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation 0.074s ? github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation/ias_credentials [no test files] ? github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation/mock [no test files] ? github.com/hyperledger-labs/fabric-private-chaincode/ercc/cmd [no test files] testing: warning: no tests to run PASS ok github.com/hyperledger-labs/fabric-private-chaincode/ercc/vscc 0.065s [no tests to run] Makefile:52: recipe for target 'test' failed make[1]: *** [test] Error 2 make[1]: Leaving directory '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v1.4/src/github.com/hyperledger-labs/fabric-private-chaincode/ercc' Makefile:11: recipe for target 'test' failed make: *** [test] Error 2

mhhh ... I can not reproduce this error. Can you double check that `ercc/attestation` does not contain a provider.go file anymore when you run make in `ercc`? Not that provider.go is moved from `ercc/attestation` to `ercc`. The error you observe is suspicious.

mbrandenburger:
> btw: couldn't/shouldn't we have a default definition of FABRIC_PATH (to ${GOPATH}/src/github.com/hyperledger/fabric), e.g., in config.mk?

yeah I think this is a good idea

g2flyer:
> mhhh ... I can not reproduce this error. Can you double check that `ercc/attestation` does not contain a provider.go file anymore when you run make in `ercc`? Not that provider.go is moved from `ercc/attestation` to `ercc`. The error you observe is suspicious.



Lots of debugging later, i still have the problem but it's rather bizarre:  the problem is not, as originally stated, related to SGX_MODE but to a particular platform. On a separate machine with pretty much same setup it works. And it all boils down to vet, e.g., 'go vet provider.go in ercc fails on one machine as above but not on the other.  I've tried different go versions (1.11.5 vs 1.12.4), cleaning out various caches from go which i didn't know before, removing some vet version which were lingering around in GOPATH to make sure the one from (implicit) GOROOT is used, analyzing go vet -x  output.  Everything looks same on both sides, but evidentally is no :-(   I'm at a loss.  That said, it seems dot imports might go even away, or at least are a quirk -- https://github.com/golang/go/issues/29326 -- and there is really any need to use dot-imports.  When i apply following patch it works fine for me:

```
diff --git a/ercc/provider.go.template b/ercc/provider.go.template
index 78cc1c8..a9259ba 100644
--- a/ercc/provider.go.template
+++ b/ercc/provider.go.template
@@ -9,30 +9,30 @@ package ercc
 #ifdef USE_SGX_HARDWARE_MODE

 import (
-       . "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation"
+       "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation"
 )

-func GetVerifier() Verifier {
-       return &VerifierImpl{}
+func GetVerifier() attestation.Verifier {
+       return &attestation.VerifierImpl{}
 }

-func GetIAS() IntelAttestationService {
-       return NewIAS()
+func GetIAS() attestation.IntelAttestationService {
+       return attestation.NewIAS()
 }

 #else // USE_SGX_HARDWARE_MOD

 import (
-       . "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation"
-       . "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation/mock"
+       "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation"
+       "github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation/mock"
 )

-func GetVerifier() Verifier {
-       return &MockVerifier{}
+func GetVerifier() attestation.Verifier {
+       return &mock.MockVerifier{}
 }

-func GetIAS() IntelAttestationService {
-       return &MockIAS{}
+func GetIAS() attestation.IntelAttestationService {
+       return &mock.MockIAS{}
 }

 #endif

```

mbrandenburger:
mh super wired issue here .... I am ok with removing dot imports.

PR #80, opened 2019-06-04
Remove duplicated fabric script

g2flyer:
* call fabric script directly from fabric instead of a (modified) copy &
   add dependency so make linter runs properly even when running after
   checkout


Makefile
```diff
@@ -16,8 +16,11 @@ license:
 	@echo "License: Running licence checks.."
 	@${GOPATH}/src/github.com/hyperledger/fabric/scripts/check_license.sh
 
-linter:
+linter: gotools build
```
mbrandenburger: I can see the need of having the build target as an requirement to successfully run linting as many generated files are missing.

However, I have a strong feeling that linter should not have this dependency for the following reasons:

- this target should be fast
- build triggers the recreation of our fabric-ccenv-sgx docker image
- I believe generated files, such as `mrenclave.go` should have no need for linting.

g2flyer: well, the problem right now is that unless the files are generated, it will fail and it's not easy to see how the script could be changed to handle this.  The best alternative which comes in mind is to drop the dependency and maybe just add an warning/echo (conditional on failure) that you might have to run 'make build' first?

g2flyer: BTW: i don't completely agree with your 'generated have no need for linting'. At least if the source to the generation is so close to go as in our use-cases it seems to me it should apply there as well (or otherwise it would not be clear to me why we would care for other files but not for these)?

g2flyer:
> also, in `golinter.sh`, does it make sense to read the path from $FABRIC_PATH and if not exists use this as default?! WDYT?

You mean replace the sym-link with  a wrapper script which looks at FABRIC_PATH and otherwise use relative path?  But you still would have to solve the problem on how to invoke it that it treats the local FPC rather than the fabric files? (Maybe a source might work?).   That said, given that we rely also on other places on fabric being in the same GOPATH-root, it seems to me the symlink should work for now?

mbrandenburger:
>  That said, given that we rely also on other places on fabric being in the same GOPATH-root, it seems to me the symlink should work for now?

ok, let's keep it like that!

Makefile
mbrandenburger: ok, let's quickly recap how one would use linter:

1. CI - when used by CI we can make sure that build target runs before - so no problem (BTW that means we need to add `checks` to the `all` target.

2. Developer/contributor - before committing code to this project the code should pass the `checks` target, I can imaging that `make checks` is performed a couple of times and thus it should be fast and not annoying. 

I suggest that we add a warning/echo that `make build` must be at least performed once before linting. WDYT?

g2flyer: to play devil's advocate, before commiting code you should run 'make test' which obviously implies a build and ideally also a check?

PR #91, opened 2019-06-19
Revert "Unit test cleaning"

g2flyer:
Reverts hyperledger-labs/fabric-private-chaincode#78

I've tested my patch only for 'make test' in ercc where it worked, turns out the integration test does actually fail with this patch in a 
```
Error: endorsement failure during invoke. response: status:500 message:"Setup failed: Can not register enclave at ercc: Error while retrieving attestation report: report does not contain submitted quote"
```


g2flyer:
Figured out problem: will provide separate PR which provides directly fix ...

Issue #83, opened 2019-06-13
refactor shim and chaincode interface


mbrandenburger:
the methods provided by `shim.h` should be revisited. Error handling is missing, for instance, `getState` does not return an error when a integrity violation is detected.

mbrandenburger:
logging should be revisited as well #88

mbrandenburger:
we also should refactor the chaincode interface, for instance, returning the arguments already as array instead of forcing the chaincode to unmarshal the args using json.

```
invoke(const char *args, uint8_t *response, uint32_t max_response_len, uint32_t *actual_response_len, void *ctx);
```

g2flyer:
We should also clean up shim.h with internal functions not targeted to chaincode developers (e.g., all the rwset functions). They should be moved out of the shim to a separate header-file such that shim.h is exactly and only the complete external programming model.

g2flyer:
If we have unmarshall functions for inputs in shim.h, shouldn't we have corresponding marshall functions for results? Seems strange to have only one but not the other ...

g2flyer:
More long-term, but we probably also want to have some `put_public_state function` as discussed earlier

g2flyer:
Shouldn't `shim.h` also contain some form of interface definition of functions which the chaincode has to implement, i.e.,
```
int invoke(const char* args,
    uint8_t* response,
    uint32_t max_response_len,
    uint32_t* actual_response_len,
    void* ctx)
{
```?

g2flyer:
If we want enable client authentication based on MSP, we might also want to add a corresponding API function

g2flyer:
eventually we want to benefit also from fabric private data collection. 

They definitely can give a performance boost when dealing with large data as the data doesn't have to go all to the orderer.  Enabling this, though, can be done under the cover and does not have to be exposed in the API.

Less clear is whether there are some scenario where we want to expose it to the developer the `{Get,Put}PrivateData*` which are in the normal Go Shim (https://godoc.org/github.com/hyperledger/fabric/core/chaincode/shim).  In their case of course it might be necessary due to some data being shared with chaincodes which are not part of the collection but for us it really shouldn't make a security difference (and hence following occam's razor i would not expose it).  Maybe iff we add a `PutPublic*` it might be slightly different as we would then have a somewhat semi-private setting (though we should then probably call it `PutPublicPrivate` or `PutSemiPublic` rather than the `PutPrivateData` ...

[...]

PR #93, opened 2019-06-20
Separation of shim/enclave from chaincode app

bvavala:


Makefile
```diff
@@ -1,11 +1,12 @@
+# Copyright 2019 Intel Corporation
 # Copyright IBM Corp. All Rights Reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc integration
+SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc private-chaincode-apps integration
```
mbrandenburger: I would rename `private-chaincode-apps` to just `examples`

private_chaincode_apps/CMakeVariables.txt
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
mbrandenburger: please use short SPDX license header

private_chaincode_apps/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+# Copyright 2019 Intel Corporation
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.5.1)
+project(FPC_APPS)
```
mbrandenburger: I have the feeling that FPC_APPS is not the right name here. this folder will just contain a chaincode, which may be part of a "bigger" blockchain application. I still like `examples`

private_chaincode_apps/Makefile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Intel Corporation
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+TOP = ..
+include $(TOP)/build.mk
+
+BUILD_DIR := _build
```
mbrandenburger: does it make sense to have a makefile here? As a developer I would assume that each example chaincode (echo, auction,...) is a self-contained project, maybe similar as the SampleCode in the linux-sgx repo. (https://github.com/intel/linux-sgx/tree/master/SampleCode)

private_chaincode_apps/echo/echo_cc.cpp
```diff
@@ -0,0 +1,261 @@
...
+// implements chaincode logic for invoke
+int invoke(const char* args, uint8_t* response, uint32_t max_response_len,
+    uint32_t* actual_response_len, void* ctx)
+{
+    LOG_DEBUG("EchoCC: +++ Executing chaincode invocation +++");
+    LOG_DEBUG("Args: %s", args);
+
+    std::vector<std::string> argss;
+    // parse json args
+    unmarshal_args(argss, args);
```
mbrandenburger: when I see this, I actually think the shim should provide the args already as array instead of unmarshalling here. I add a note in the shim refactoring issue #83

private_chaincode_apps/echo/echo_json.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <stdint.h>
+#include <string>
+
+typedef struct echo {
+    std::string echo_string;
+} echo_t;
+
+int unmarshal(echo_t* echo, const char* json_bytes, uint32_t json_len);
```
mbrandenburger: maybe json is not even needed in this echo chaincode

private_chaincode_apps/CMakeLists.txt
bvavala: removed

private_chaincode_apps/echo/echo_cc.cpp
bvavala: echo has to be cleaned up in a later PR

private_chaincode_apps/CMakeVariables.txt
bvavala: removed

private_chaincode_apps/Makefile
bvavala: yes, just to loop through the examples.
the cmake has been removed

private_chaincode_apps/echo/echo_json.h
bvavala: probably. but it can be enhanced later

bvavala:
This PR mainly separates the private chaincode applications from the `ecc` component. Important enhancements include:
* the build of `ecc` produces a boilerplate docker image, which is meant to be later extended with the chaincode app
* the auction chaincode is now an example
* an  echo chaincode has been added to the examples
* the integration test now performs two serial chaincode executions (auction and echo respectively)

This PR must be improved as follows (issues will be created):
* the boilerplate docker image should include the `ecc` binary. This is not possible now because compiling `ecc` requires `mrenclave.go`, which includes the hardcoded MRENCLAVE of the enclave, which is only available once the chaincode application is compiled. Solution: the MRENCLAVE value should be read from a file, which is added to the boilerplate image once the application is compiled.
* the `echo` chaincode must be cleaned up. Also, the write to the ledger is not necessary. Also, the echoed string in the response field of the transaction should be checked.

Makefile
```diff
@@ -1,11 +1,12 @@
+# Copyright 2019 Intel Corporation
 # Copyright IBM Corp. All Rights Reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc integration # docs
```
g2flyer: I actually left the docs there commented out on purpose to explicitly show that we do not make docs now.  But admittedly not an essential thing ...

ecc/Dockerfile.boilerplate-ecc
```diff
@@ -13,7 +13,8 @@ RUN mkdir -p ${CC_LIB_PATH}
 ENV SGX_SDK=/opt/intel/sgxsdk
 ENV LD_LIBRARY_PATH=${SGX_SDK}/sdk_libs:${CC_LIB_PATH}
 
-copy ecc/${CC_NAME} ${CC_PATH}/chaincode
-copy ecc_enclave/_build/lib/*.so ${CC_LIB_PATH}/
+# TODO: the boiler plate should include the compiled ecc
```
g2flyer: I thought we wanted to outsource the MRENCLAVE into a text file rather than baking it into ecc (and maybe revisit why what's the point of having MRENCLAVE of fpc chaincode used inside ecc in the first place). 
I guess you dropped that as you run out of time?  If so, we should at least add a corresponding issue (and reference it here)?

ecc/Dockerfile.fpc-app
```diff
@@ -0,0 +1,14 @@
...
+FROM fabric-private-chaincode/boilerplate-ecc
+
+ARG enclave_so_path
+ARG CC_PATH="/usr/local/bin"
+ARG CC_LIB_PATH=${CC_PATH}"/enclave/lib"
+
+RUN test -n "$enclave_so_path" 
+
+copy ${enclave_so_path}/enclave.signed.so ${CC_LIB_PATH}/
+copy ecc/ecc ${CC_PATH}/chaincode
```
g2flyer: Analogous to above boilerplate we might want to add a comment here that this chaincode should eventually go away (and also refer to related issue?)

ecc/Makefile
```diff
@@ -1,11 +1,13 @@
+# Copyright 2019 Intel Corporation
 # Copyright IBM Corp. All Rights Reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
 TOP = ..
 include $(TOP)/build.mk
 
-LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$$(pwd)/enclave/lib
+LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}
```
g2flyer: hmm, if we don't define anything, why not just drop it? Seems rather confusing now (and in general might even be problematic, at least i'm not 100% whether the semantics of LD_LIBRARY_PATH is the same of having it undefined vs defined as empty ...

ecc/Makefile
```diff
@@ -15,19 +17,28 @@ VSCC_OUT = ecc-vscc.so
...
+all: build
+# test target disable because it requires revision
+#all: build test
+build: docker-boilerplate-ecc vscc-plugin
 
 ECC_ENCLAVE_BUILD = ../ecc_enclave/_build
 ERR_MSG = "ecc_enclave build does not exist!"
 
-ecc_enclave:
+ecc_enclave: sym-links
```
g2flyer: I guess all that sym-linking is also due to not-yet removing/outsourcing MRENCLAVE? If so, maybe also add comment as for docker files?

ecc/Makefile
```diff
@@ -1,11 +1,13 @@
...
 # Copyright IBM Corp. All Rights Reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
 TOP = ..
 include $(TOP)/build.mk
 
-LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$$(pwd)/enclave/lib
+LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}
+
```
g2flyer: I think you rely on ENCLAVE_SO_PAT for finding fpc-c chaincode (enclave.so).  Maybe also add a define here so all the parameter for the docker build would be at a single place?  (we might also add a commenting header saying that these are the docker build params?)

fabric/bin/peer.sh
```diff
@@ -36,15 +36,19 @@ handle_chaincode_install() {
 		CC_LANG=$2
 		shift;shift
 		;;
+        -e|--enclavesopath)
```
g2flyer: Why add a new parameter? This is really semantically the -p|--path option for language fpc-c?

examples/echo/echo_json.cpp
```diff
@@ -0,0 +1,28 @@
+/*
```
g2flyer: I don't see that used. I guess this is the beginning of a utility to help verifying also return results from invocations/queries?   If so, this raises a number of questions:
- couldn't (and shouldn't!) this be generic and hence also should be rather in fabric/bin or alike?
- wouldn't that have been easier scripted in python or alike? (i'd venture, extracting the return result from the peer json can be wrritten in a one-liner python or perl command-line invocation?)
- if we don't complete it, maybe better to leave out of this PR and add to one which does it end-to-end? (With right now just creating a corresponding issue?)

ecc/Makefile
```diff
@@ -57,12 +60,24 @@ clean: docker-clean
...
 	@if [ -z "${DOCKER_IMAGE}" ]; then\
-		echo "\033[0;31mERROR: cannot override $(CC_NAME) docker image - not found\033[0m";\
+		echo "\033[0;31mERROR: cannot override $(CC_NAME) docker image (${DOCKER_IMAGE}) - not found\033[0m";\
 		exit 1;\
 	fi
-	echo "\033[1;33mWARNING: overriding $(CC_NAME) docker image\033[0m"
+	echo "\033[1;33mWARNING: overriding $(CC_NAME) docker image (${DOCKER_IMAGE})\033[0m"
 	$(DOCKER) build -t $(DOCKER_IMAGE) -f Dockerfile ..
 
+docker-boilerplate-ecc:
```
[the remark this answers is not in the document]
bvavala: correct

ecc_enclave/enclave/CMakeLists.txt
```diff
@@ -50,37 +55,51 @@ include_directories(
...
 
-add_custom_command(TARGET enclave
+add_custom_command(TARGET ${ENCLAVE_BASE_LIB_NAME}
     POST_BUILD
-    COMMAND ${COMMON_SOURCE_DIR}/enclave/generate_mrenclave.sh ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}
-    COMMAND cp ${CMAKE_BINARY_DIR}/mrenclave ${CMAKE_BINARY_DIR}/lib
-    COMMAND cp ${CMAKE_BINARY_DIR}/enclave.signed.so ${CMAKE_BINARY_DIR}/lib
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-    COMMENT "Create enclave mrenclave")
+    #<<<<<<< Updated upstream
```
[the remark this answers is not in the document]
bvavala: this is outdated

integration/echo_test.sh
```diff
@@ -0,0 +1,100 @@
...
+SCRIPTDIR="$(dirname $(readlink --canonicalize ${BASH_SOURCE}))"
+FPC_TOP_DIR="${SCRIPTDIR}/.."
+CONFIG_HOME="${SCRIPTDIR}/config"
+
+. ${SCRIPTDIR}/common_utils.sh
+. ${SCRIPTDIR}/common_ledger.sh
+
+CC_ID=ecc
+FPC_APP_NAME=echo
+# TODO: once issue #86 is fixed, change above to ecc_echo_test
```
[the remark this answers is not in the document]
bvavala: moved up

ecc/Dockerfile.boilerplate-ecc
bvavala: right, this is one of the improvements I have mentioned above.
I will convert these to issues once this is merged

ecc/Dockerfile.fpc-app
bvavala: done

ecc_enclave/enclave/CMakeLists.txt
```diff
@@ -36,7 +37,11 @@ add_custom_command(
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     )
 
-add_library(enclave SHARED ${SOURCE_FILES})
+set(ENCLAVE_BASE_LIB_NAME enclave-base)
```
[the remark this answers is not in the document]
bvavala: sounds reasonable, renamed

ecc/Makefile
bvavala: it must not be empty cause the sgx sdk libraries are needed.
however, yes, it can be simply removed (also from hte commands) at this point.
done.

ecc/Makefile
bvavala: done

ecc/Makefile
bvavala: not sure what you mean, but I created an internal variable, now check the value, and clarified what the docker build parameters are

fabric/bin/peer.sh
bvavala: it was not clear at all that -p was there for this.
anyway, fixed

examples/echo/echo_json.cpp
bvavala: the `echo` (as well as the `auction`) cc is not complete. It however performs some operations, and it works now for showing that we can run 2 different cc in serial. That's why it is in this PR, and I will add an issue once merged.

As Marcus pointed out the json machinery here might not be needed, so they might be removed.

ecc/Makefile
g2flyer: My comment was in the context of the three definitions CC_NAME, NET_ID and PEER_ID which are following at the location i commented:  These three variables define almost but not quite the "api" you would use for 'docker-fpc-app'.  The almost being that you also will have to define that path to the code which i thought is ENCLAVE_SO_PATH.   Hence my suggestion to also add a ?= define or alike for ENCLAVE_SO_PATH to the other three defines and add a comment above the four of them that this defines the 'make docker-fpc-app' api

examples/echo/echo_json.cpp
g2flyer: oh, i definitely dd not want to pull the echo test from the PR.  My comment purely to the json response processing (in regards of not being echo specific, maybe simpler dealt with a script and better included in this PR only if complete/used (compared to a separate PR which might generalize and also instrument the auction test with checks of return value)

bvavala: the echo test has been cleaned up

bvavala: @g2flyer I think Marcus is waiting for your approval

bvavala:
> This looks good to me! I created a hello world example private chaincode to test this. Works now.
> Thanks for this huge piece of work!


Great!
What was wrong with the initial hello-world implementation?

examples/echo/echo_json.cpp
g2flyer: @bvavala well, i was waiting for you to re-request the review after my initial review. Didn't know when you were done with your changes.  But i assume that is now the case?

bvavala: sorry @g2flyer , there was no button to push for that (I might have done that some days ago).
Anyway, yes, and I have also rebased

g2flyer: actually, there is a button: just click on the "arrow-circle" to the left of the reviewer asking for changes and then s/he will be notified to re-review (that is what marcus has done)

bvavala:
> the new version addresses most of my comments, see though a few additional comments. Primarily i'd suggest to already create all follow up issues as otherwise they might get missed with vacations and al.
> 
> Also, there is a format issue in echo_cc.h as 'make' fails as follows ...
> 
> LINT: Running code checks for Cpp/header files... ERROR in format: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v1.4/src/github.com/hyperledger-labs/fabric-private-chaincode/scripts/../examples/echo/echo_cc.h Format check failed. Run './scripts/cpplinter.sh DO_FORMAT' to fix the format.

added 3 issues, and fixed the cpplinter complaint (I still wonder how I missed that since I have tested)

ecc/Makefile
```diff
@@ -1,68 +1,83 @@
...
-	$(LD_LIBRARY_PATH) $(GO) build
+	$(GO) build
 
 vscc-plugin: $(VSCC_SRC)
-	$(LD_LIBRARY_PATH) $(GO) build -o $(VSCC_OUT) -buildmode=plugin $^
+	$(GO) build -o $(VSCC_OUT) -buildmode=plugin $^
 
-test:
-	$(LD_LIBRARY_PATH) $(GO) $@ -test.v ./...
+#test:
```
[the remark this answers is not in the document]
bvavala: yes, issue #105

ecc/Makefile
```diff
@@ -1,68 +1,83 @@
...
 CC_NAME ?= ecc
 NET_ID ?=dev
 PEER_ID ?=jdoe
 
 VSCC_SRC = vscc/ecc_validation_plugin.go vscc/ecc_validation_logic.go
 VSCC_OUT = ecc-vscc.so
 
-DOCKER_IMAGE ?= $$(docker images | grep -- ${NET_ID}-${PEER_ID}-$(CC_NAME)- | awk '{print $$1;}')
 DOCKER_CONTAINER_ID?=$$(docker ps | grep -- ${NET_ID}-${PEER_ID}-$(CC_NAME)- | awk '{print $$1;}')
+# the following are the required docker build parameters
```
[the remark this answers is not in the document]
bvavala: ignored!

common/enclave/generate_mrenclave.sh
```diff
@@ -27,7 +28,14 @@ cat $base64_out_name
 
 echo "Create go file"
 touch $go_name
-echo "package enclave" > $go_name
+echo "/*" >> $go_name
```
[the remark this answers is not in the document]
bvavala: kind of, didn't think about that.
Much cleaner now!

Makefile
bvavala: oh, did I remove that?! .. restored!

mbrandenburger:
> Great!
> What was wrong with the initial hello-world implementation?

Cannot really say. Maybe a `make clean` already helped!

BTW. I messed up you PR a bit by squashing and not taking care of a proper commit msg. Sorry :/

Issue #96, opened 2019-06-22
Docu update for SDKization


g2flyer:
Bundle all docu update related to the SDKization refactoring and related (e.g., the removal of the explicit docker switcheroo) to after the corresponding code changes are done

mbrandenburger:
- [ ] ECC refactoring #69 

- [ ] remove switcheroo #97

mbrandenburger:
Fix also _Time to get the code_ section.

```
$ git clone https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/hyperledger-labs/fabric-private-chaincode
```

```
$ git clone https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
```

mbrandenburger:
Consolidate READMEs. With the new build system we don't need extra README in each component folder that explains how to build the component. I suggest that we combine them in a single README and only provide extra information if needed.

mbrandenburger:
content for new section FAQ:

- `sudo make` solves issues with misconfigured docker environments but changes `$GOPATH`

https://docs.docker.com/install/linux/linux-postinstall/
https://stackoverflow.com/questions/55843277/cant-run-go-program-using-sudo-command

mbrandenburger:
- Maybe move "Behind a proxy" to FAQ section or closer to IAS section

blockchaindemo:
Add:    sudo snap install --stable --classic protobuf
in Get NanoPB section

g2flyer:
we should  also a simple software development section on the FPC programming model (shim.h et al ...)

mbrandenburger:
> we should also a simple software development section on the FPC programming model (shim.h et al ...)
> 
> We definitely also should document that folks have to use the wrapper scripts from fabric/bin and point out the utility scripts fabric/bin/ledger_init & fabric/bin/ledger_shutdown if users want to interactively play with FPC ..

Let's address this briefly in #129 and provide concise documentation on the programming model later.

g2flyer:
addressed mostly in PR #142 (issue #129) and PR #137 (with increased docu inside shim.h)

PR #95, opened 2019-06-21
update licence headers

mbrandenburger:
Signed-off-by: Marcus Brandenburger <bur@zurich.ibm.com>

PR #97, opened 2019-06-22
Goodbye explicit docker switcheroo

g2flyer:
This PR contains the promised peer wrapper which does automatically a pro-active docker switcheroo iff the chaincode installed is of --lang fpc-c.    integration test is correspondingly automated.  Not touched is fabric/sgxconfig/demo and the docu. The former will be made obsolete by bruno's upcoming PR and the docu we best do after finishing the SDKization which anyway requires a bigger docu revisit, defined as issue #96


docs/fpc-lifecycle.txt
```diff
@@ -0,0 +1,60 @@
+@startuml
```
mbrandenburger: can we have a docs/README.md documenting how to generate these very nice uml diagrams? :)

ercc/attestation/.gitignore
```diff
@@ -0,0 +1,3 @@
+ias.go
```
mbrandenburger: this .gitignore seems like a relict to me. With #68 this has been changed and moved to `ercc/`

fabric/bin/lib/common_ledger.sh
```diff
@@ -7,19 +7,23 @@
 # - CONFIG_HOME is defined
 # - common_utils.sh is loaded
 # optional config overrides
+# - FABRIC_PATH
```
mbrandenburger: you might want to reuse $FABRIC_PATCH from `config.mk`

integration/auction_test.sh
```diff
@@ -19,24 +20,11 @@ auction_test() {
     expect_switcheroo_fail=$1
 
     # install, init, and register (auction) chaincode
-    # install some dummy chaincode (we manually need to create the image)
-    try ${PEER_CMD} chaincode install -n ${CC_ID} -v ${CC_VERS} -p github.com/hyperledger/fabric/examples/chaincode/go/example02/cmd
+    try ${PEER_CMD} chaincode install -l fpc-c -n ${CC_ID} -v ${CC_VERS} -p github.com/hyperledger-labs/fabric-private-chaincode/ecc/
```
mbrandenburger: why not calling it `fpc-cpp`? I think at the moment we can also just call it `fpc` as we do not support any other chaincode language as c/cpp. This may of course change later (far future (next gen release)) with other chaincode languages support. So now I would call it just `fpc`

utils/fabric/get-fabric-container-name.go
```diff
@@ -0,0 +1,27 @@
+// Copyright Intel Corp. 2019 All Rights Reserved.
```
mbrandenburger: ehm picky comment: let's try to be consistent with #95 license header format for go files 💃

utils/fabric/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+GO_CMDS= get-fabric-container-name
+
+build: ${GO_CMDS}
+
+${GO_CMDS}:
+	$(GO) build $@.go
+
+test: build
+#	$(GO) test -v
```
mbrandenburger: Wouldn't it be nice to have a test here? :)

ercc/attestation/.gitignore
g2flyer: this was an artifact from before your attestation-refactoring-disguised as a test-update overlooked in the merge ;-)  will remove ...

utils/fabric/Makefile
g2flyer: the program is so simple and actual semantics is defined by fabric, so not sure what test program i could meaningfully write.  I'll remove the test from the makefile, though

utils/fabric/get-fabric-container-name.go
g2flyer: i thought we want to have short header? In any case, license check passes?

integration/auction_test.sh
g2flyer: cpp sounds to me like c pre-processor but i guess you mean c++?  I realize that we don't currently support any other language but i still think it's useful to mention the language, after all the param is called --lang.  What about fpc-c++  if we want to emphasize the ++ aspect of c?

fabric/bin/lib/common_ledger.sh
g2flyer: Not sure what you are referring to: this is a shell script which should be executable separate from make, so can't use config.mk.  It also defines currently imho sensible defaults consistent with config.mk and i mention it in the comment only as docu that you can override defaults ...

utils/fabric/get-fabric-container-name.go
g2flyer: btw: shouldn't the *.md files also have headers?

g2flyer: will add ...

docs/fpc-lifecycle.txt
g2flyer: will add Makefile and text to README on what to install

utils/fabric/get-fabric-container-name.go
mbrandenburger: argh ... 
I suggest that we do follow the Hyperledger approach and add this to the project main readme

```
Hyperledger Project documentation files are made available under the Creative Commons Attribution 4.0 International License (CC-BY-4.0), available at http://creativecommons.org/licenses/by/4.0/.
```
see https://github.com/hyperledger/fabric

That way I dont see need to put license headers in *.md files.

tlcc_enclave/README.md
```diff
@@ -40,4 +44,4 @@ Note that OPENSSL sometimes complains, here you can just continue debugging.
 
 ## Cleaning
 
-    $ make clean
\ No newline at end of file
+    $ make clean
```
mbrandenburger: what is changing here?

integration/auction_test.sh
```diff
@@ -81,32 +69,13 @@ trap ledger_shutdown EXIT
 
 # 2. First run, this should fail due to docker-switcheroo
 para
-say "FIRST RUN: this should fail due to docker-switcheroo ..."
+say "Run auction test"
```
mbrandenburger: so cool!

integration/auction_test.sh
mbrandenburger: Once we have the SDKization in place, will we the enduser install the chaincode using `-p github.com/hyperledger-labs/fabric-private-chaincode/examples/auction`?

g2flyer: that was more or less the thought with -p. Everything should be in place in this PR and is more or less designed to work with the changes from Bruno's PR in mind

g2flyer:
- Did now remove 'doc' from subfolders in main Makefile as discussed in call
- after cross-checking with Mic i decided to keep explicit attribution in documentation
- replaced the sleeps after chaincode invoke with --waitForEvent (alas there are still some sleeps after install and instantiate where i couldn't immediately see how to get rid of the sleep ...)

CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,7 @@
+<!---
```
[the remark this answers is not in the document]
g2flyer: Cross-checked with Mic: HL definitely wants attribution and he also thinks explicit attribution as for source is the right way to address it.  So i've kept it ...

Makefile
```diff
@@ -5,7 +5,7 @@
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = utils/fabric-ccenv-sgx ercc ecc_enclave ecc tlcc_enclave tlcc integration
+SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc docs integration
```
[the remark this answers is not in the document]
g2flyer: for now commented out docs.  In long run, we should want to build docu as part of overall make and the way to go is to have a docker build image which really would simplify quick get'up'n'running (we have more complex dependencies right now that java and plantuml ...)

Issue #85, opened 2019-06-14
store spid as text


g2flyer:
Remove the requirement to do the xxd hack when creating spid.txt (or at least rename file to spid.bin iff we really insist on this binary format ...)

mbrandenburger:
#100 tackles this issue

mbrandenburger:
fixed with #100

PR #100, opened 2019-06-27
Load SPID as text (#85)

mbrandenburger:
This PR allows to store SPID as text format and performs conversion to
binary format within ercc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ could a mismatch between provided SPID and api key as specified below).
 Place your ias api key and your SPID in the ``ias`` folder as follows:
 
     echo 'YOUR_API_KEY' > ${GOPATH}/src/github.com/hyperledger-labs/fabric-private-chaincode/config/ias/api_key.txt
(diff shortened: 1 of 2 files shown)


mbrandenburger:
> you require now goimports as part of builds, whereas goimports is added only later via target 'gotools' in top-level Makefile. Maybe gotools should be a target unconditionally added as dependency to build target in build.mk?

well my IDE runs goimports. 

> Also, while it works in simulator mode, i couldn't get it to work for either linkable or unlinkable attestation mode as get always a 400 from IAS (and yes, i did update spid.txt :-). Looking at debug output i also see a printed out SPID which is evidentally wrong (but maybe that's before decoding?)

I fixed that by moving the decoding to the decorator for the moment. The problem was, that the decoding was not even called since ERCCstub in ecc also reads from the decorator instead invoking getSPID at ercc. I created an issue for that #100.
This will be fixed with refactoring ercc.

Actually it was signed-off but now DCO checks runs through.

PR #102, opened 2019-07-02
remove double build

bvavala:
`make all` has `build` as a dependency, and... a "recipe", which is the same as `build`.
Consequently, the build procedure was performed twice.
Also the top CmakeList has been removed because it is not used.


bvavala:
done

build.mk
```diff
@@ -8,14 +8,13 @@ include $(TOP)/config.mk
 -include $(TOP)/config.override.mk
 
 .PHONY: all
-#all: build checks test integration
 all: build test checks # keep checks last as license test is brittle ...
+
```
g2flyer: Why are you moving the .PHONY up here?  i think we should keep all phony together. Right now this one stands in middle of two dependency target and kind of shadows the test: build dependency ...

g2flyer: also, what's the logic of keeping .phony here?  if we keep docker, why not integration?

bvavala: if a phony is close to target definition, chances are we look/modify when we look/modify the target. Personally, I did not know about the list, and tomorrow it'll be already forgotten.

the phony integration was commented and I removed the comment. If you want I can put the phony, though in the main makefile where it belongs.

g2flyer: i'm mainly asking for consistency: if we define most phony in build.mk we better do all or always define them in makefile. Most of these phony though really don't matter and obscure more important information, so i think keeping them in build.mk seems to me the preferred way.  But as you notice, i've approved the PR without requiring a change, just an observation ...

Issue #103, opened 2019-07-03
boilerplate image should include the ecc binary


bvavala:
Currently the bilerplate image does not include the ecc binary because it requires mrenclave.go, which is available only after the chaincode enclave is compiled. So ecc is built after the chaincode enclave and then added to the final docker image.

g2flyer:
One way to address that is to load mrenclave from a text-file instead of baking it into the code.  But bigger question for me is why would ecc need mrenclave baked into it? It always can get it at runtime if needed and ecc is not trusted anyway, so it's not really a trust question of baking it in?

mbrandenburger:
this is why ecc has mrenclave baked in:

when a private chaincode is instantiated, mrenclave is written to the ledger as part of the chaincodes genesis block. Thus, this defines the code running in the enclave. At this time there is no enclave spawned which we could ask for its mrenclave.

As you said, ecc is untrusted but the client who instantiates the chaincode "can" verify that the writeset actually contains "the" expected mrenclave before submitting the transaction proposal response to the ordering service. Having a proper endorsement in place would also enable to detect a bad peer writing a "wrong" mrenclave into the chaincode's genesis block.

Our custom vscc for ercc relies on mrenclave in order to verifiy that an enclave is registered running the expected code. However, this is TBD as we may change to normal endorsement model here.

With new lifecycle in place mrenclave should be part of the chaincode definition and thus must be approved by the consortium.

g2flyer:
> this is why ecc has mrenclave baked in:
> 
> when a private chaincode is instantiated, mrenclave is written to the ledger as part of the chaincodes genesis block. Thus, this defines the code running in the enclave. At this time there is no enclave spawned which we could ask for its mrenclave.
> 
> As you said, ecc is untrusted but the client who instantiates the chaincode "can" verify that the writeset actually contains "the" expected mrenclave before submitting the transaction proposal response to the ordering service. Having a proper endorsement in place would also enable to detect a bad peer writing a "wrong" mrenclave into the chaincode's genesis block.
> 
> Our custom vscc for ercc relies on mrenclave in order to verifiy that an enclave is registered running the expected code. However, this is TBD as we may change to normal endorsement model here.
> 
> With new lifecycle in place mrenclave should be part of the chaincode definition and thus must be approved by the consortium.

[...]

PR #106, opened 2019-07-03
ecc docker image built as an actual boilerplate (merge after pr #93 )

bvavala:
addresses #103


scripts/cpplinter.sh
```diff
@@ -16,6 +16,8 @@ EXCLUDED_PATTERNS="\
     ecc_enclave/_build \
     tlcc_enclave/nanopb \
     tlcc_enclave/_build \
+    examples/auction/_build \
```
g2flyer: to keep this file simpler, maybe we could just exclude /_build instead of listing every single directory?

ecc/enclave/enclave_stub.go
```diff
@@ -389,6 +391,22 @@ func (e *StubImpl) Destroy() error {
 	return nil
 }
 
-func (e *StubImpl) MrEnclave() string {
-	return MrEnclave
+// Loads the MrEnclave string from file
+func LoadMrEnclave() error {
```
mbrandenburger: this method does not need to be public?

ecc/enclave/enclave_stub.go
```diff
@@ -389,6 +391,22 @@ func (e *StubImpl) Destroy() error {
...
+	binMrEnclave, err := ioutil.ReadFile("mrenclave")
+	if err != nil {
+		return fmt.Errorf("Error reading MrEnclave from file")
+	}
+	MrEnclave = string(binMrEnclave)
```
mbrandenburger: better return just return mrenclave instead of storing in a public var

ecc/enclave/enclave_stub.go
```diff
@@ -389,6 +391,22 @@ func (e *StubImpl) Destroy() error {
...
+	return nil
+}
+
+func (e *StubImpl) MrEnclave() (string, error) {
```
mbrandenburger: since MrEnclave() is only invoked once during init ... why not reading here from file?

ecc/enclave/enclave_stub.go
```diff
@@ -389,6 +391,22 @@ func (e *StubImpl) Destroy() error {
 	return nil
 }
 
-func (e *StubImpl) MrEnclave() string {
-	return MrEnclave
+// Loads the MrEnclave string from file
+func LoadMrEnclave() error {
+	binMrEnclave, err := ioutil.ReadFile("mrenclave")
```
mbrandenburger: what if the file is empty? or has some invalid mrenclave value?

mbrandenburger:
since Bruno is on vacations, I suggest we close this PR; I take over fix changes, and then we can merge. WDYT?

mbrandenburger:
Ok, we close this PR and create another one.

PR #116, opened 2019-07-30
WIP: Add enclave lifecycle UML diagrams

mbrandenburger:
This adds UML sequence diagrams for the current model and
the new approach for enclave registration as discussed at the
FPC project meeting.


docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+    - ERCC and ECC share endorsement policy 
+    - attestations stored ECC namespace
+  Separate:
+    - flexible endorsement policies (ERCC and ECC can have different)
+    - attestations stored in ERCC namespace;
+end note
+
+entity IAS order 50
+collections Orderer order 70
+collections Peers order 80
```
g2flyer: We really show it only for one peer right now.  At some level i think having at least two peers would help understanding all implications?

On a related note, as key-mgmt is a core part of ERCC, it would be nice to also add these flows?

docs/fpc-registration-old.puml
```diff
@@ -0,0 +1,83 @@
...
+participant ERCC order 40
+entity IAS order 50
+collections Orderer order 70
+collections Peers order 80
+
+note right of Peers
+  **Drawbacks**
+  - Custom validation at __every__ peer
+  - Single-endorser policy only
+  - ERCC must contact IAS (internect connection and IAS credentials needed)
```
g2flyer: Crucially also we currently have no way to securely pass the credentials to ERCC only (i.e., all chaincode running on that peer does receive the credentials ...)

docs/fpc-registration-old.puml
```diff
@@ -0,0 +1,83 @@
...
+participant ECC order 20
+participant ERCC order 40
+entity IAS order 50
+collections Orderer order 70
+collections Peers order 80
+
+note right of Peers
+  **Drawbacks**
+  - Custom validation at __every__ peer
+  - Single-endorser policy only
```
g2flyer: also note a single-endorser policy doesn't really make sense security-wise for these kind of peer-endorsed lifecycle/upgrade events (differently to attestation-based endorsement where a single one is fine as we trust the TEE fully, at least for now ...)

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+  ECC_Manager --> Admin : attestation, ECC_s_pk / error
+  deactivate ECC_Manager
+
+end
+
+group register attestation
+  Admin -> ERCC: register(attestation, ECC_s_pk)
+  activate ERCC
+  ERCC -> ERCC : check attestation signature
+  ERCC -> ERCC : check attestation ECC_s_pk and mrenclave
```
g2flyer: i think you also have to  cc-name/cc-version to that tuple and before verification also read from ledger the ECC record which put in the MRENCLAVE (presumably the first putstate also should have an additional cc-name/cc-version)

docs/fpc-registration-old.puml
```diff
@@ -0,0 +1,83 @@
...
+  - ERCC must contact IAS (internect connection and IAS credentials needed)
+  - for EPID attestation SPID is needed when quote is created (ECC must provide this)
+end note
+
+group instantiation
+   Admin -> ECC : install
+   Admin -> ECC : instantiate (exp.mrenclave)
+   ECC -> ECC : putState(ecc.mrenclave, exp.mrenclave)
```
g2flyer: what is exp.mrenclave? we have to somewhere also commit tlcc.mrenclave but i thought we would do that only during key-mgmt and so would be elsewhere (besides it would then also need a reference to the channel genesis block)

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+
+  ECC_Manager --> Admin : attestation, ECC_s_pk / error
+  deactivate ECC_Manager
+
+end
+
+group register attestation
+  Admin -> ERCC: register(attestation, ECC_s_pk)
+  activate ERCC
+  ERCC -> ERCC : check attestation signature
```
g2flyer: We probably spell out explicitly the IAS root-of-trust as this is probably also something we would want to do in the channel genesis block/definition?

docs/fpc-registration-new.puml
bvavala: definitely, though it would be better to have it in a separate diagram

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+  ECC --> Admin : success / error
+  deactivate ECC
+end
+
+Admin -> Orderer : submit instantiate transaction
+...when next block ready...
+Orderer -> Peers : broadcast block
+Peers -> Peers : Validation
+
+group enclave creation and attestation generation
```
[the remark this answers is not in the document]
bvavala: what's the semantics of approve?
According to https://buildmedia.readthedocs.org/media/pdf/hyperledger-fabric/latest/hyperledger-fabric.pdf the chaincode on a channel is defined by a pair name-version, but then... each organization associates a "possibly different" package ID with it -- since such ID contains the hash of the package, then the package itself can be different.
So,
* MRENCLAVE cannot be part of the package, because it could be different
* it should be part of the chaincode definition, but I am not sure that it can be extended
* it could be part of the cc name, but then the infrastructure should enforce that this is equal to the actual enclave's

docs/fpc-registration-new.puml
g2flyer: sure, didn't necessarily mean the same diagram, just nice to have the "whole" picture of ercc as key-mgmt is a key-dist is a key part of it.  What could be also potentially useful is a simple class diagram of the ERCC (and, maybe ECM) state?

docs/fpc-registration-new.puml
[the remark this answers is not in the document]
g2flyer: The package id is afaict only a local identifier created as you do a peer lifecycle chaincode install.  That said, packages (and other approve params like plugins) could be different in the approve. However, a subsequent commit will only succeed if sufficient approveformyorg do agree on the same content (i.e., chaincode definition). See January lifecycle play back on https://wiki.hyperledger.org/display/fabric/Playbacks and/or look at lifecycle slide Marcus "extracted" from Ale a while back (lifecycle_pres_2019-113-updated.odp) ..

So that's exactly why you want MRENCLAVE in package (and approve) in it as it exactly the place where each organization explicitly defines the chaincode definition (i.e., the content of the local package as well as the other stuff like plugins) it agrees with.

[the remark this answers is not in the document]
bvavala: Maybe you're right - docs are a bit confusing about this point.
Alright, in this case, I interpret the two options you mentioned for MRENCLAVE as follows:
* direct: MRENCLAVE is part of the cc definition, which has to be extended (and this might require to modify Fabric).
* indirect: MRENCLAVE is in the cc itself (e,g., in the shim)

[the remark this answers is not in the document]
g2flyer: I guess in principle you could bake in MRENCLAVE into a shim binary but i don't really see the point, at least from an architectural perspective? The shim doesn't really matter (and i'm not sure whether the shim is really in the measurement for the javascript/typescript version or the java one). Also we really want a fixed shim, so it cannot really contain the MRENCLAVE?

That said, where you thinking of the current state before going to a true chaincode type?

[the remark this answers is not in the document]
bvavala: > I guess in principle you could bake in MRENCLAVE into a shim binary but i don't really see the point, at least from an architectural perspective? The shim doesn't really matter (and i'm not sure whether the shim is really in the measurement for the javascript/typescript version or the java one). Also we really want a fixed shim, so it cannot really contain the MRENCLAVE?

The point is that: if organizations approve the package id, they approve also the MRENCLAVE baked inside.
Thinking about that, arguably this may not be necessary, since the hash of the package should also cover the enclave binary. So it follows that the agreement on the package id, implies agreement on an MRENCLAVE (whatever that is).

> 
> That said, where you thinking of the current state before going to a true chaincode type?

I don't understand the question. Rephrase?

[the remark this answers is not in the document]
g2flyer: > [...] See January lifecycle play back on https://wiki.hyperledger.org/display/fabric/Playbacks and/or look at lifecycle slide Marcus "extracted" from Ale a while back (lifecycle_pres_2019-113-updated.odp) ..

btw: https://hyperledger-fabric.readthedocs.io/en/latest/chaincode4noah.html?highlight=policy gives good discussion of approve/polices and also chaincode definition)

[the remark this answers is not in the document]
g2flyer: > The point is that: if organizations approve the package id, they approve also the MRENCLAVE baked inside.
> Thinking about that, arguably this may not be necessary, since the hash of the package should also cover the enclave binary. So it follows that the agreement on the package id, implies agreement on an MRENCLAVE (whatever that is).

Actually, looking a bit closer at code and docu i'm actually not sure whether the current code of the new lifecycle does bind to the code! I think they did implement now the model where each peer can have a different implementation as long as they agree on other stuff (see link i referenced above). This works fine as long as you only care about integrity (and trust at least as many other peers as required by the endorsement policy quorum). However, in the case of confidentiality, i.e., us, that doesn't work. So we _must_ add MRENCLAVE separatedly!

> > That said, where you thinking of the current state before going to a true chaincode type?
> 
> I don't understand the question. Rephrase?

what i was wondering whether you said that with still the current (hacky/broken) model in mind where we run our chaincode on top of go chaincode (and hence our shim is really just part of the go chaincode and hence was measured in the old model, but seemingly not anymore in the new one, see above)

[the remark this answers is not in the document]
bvavala: We agree that that part is confusing - whether the peers must agree on the package ID or not.

If yes, I think this solves our problems, without the need of adding MRENCLAVE - it implies they agree on that as well.

If not, of course we have a problem and, yes, we must add MRENCLAVE.
Now, if Fabric enables the customization of the chaincode definition and validation, then let's put it in the chaincode definition.
Otherwise, we should not put it in there, and rather implement a separate mechanism.

> (and hence our shim is really just part of the go chaincode and hence was measured in the old model, but seemingly not anymore in the new one, see above)

hmm I am referring to whatever code/data is reacheable from the Init/Invoke functions which the chaincode must implement and thus must be in the chaincode package that is measured.

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+  Admin -> ECC_Manager : create (CC_ID, channel)
+  activate ECC_Manager
+
+  ECC_Manager -> ECC : (CC_ID/Channel).create
+
+  create participant ECC_Enclave order 30 #99FF99 
+  ECC -> ECC_Enclave : create
+   
+  ECC_Manager -> ECC_Enclave : getQuote(SPID)
+  return quote, ECC_s_pk
```
bvavala: this return quote should go from ecc_enclave to the ecc_manager

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,87 @@
...
+    - attestations stored in ERCC namespace;
+end note
+
+entity IAS order 50
+collections Orderer order 70
+collections Peers order 80
+
+group instantiation
+  Admin -> ECC : install
+  Admin -> ECC : instantiate (exp.mrenclave)
```
bvavala: I am puzzled about the `exp.*`. Do we need them?

docs/fpc-registration-new.puml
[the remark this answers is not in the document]
g2flyer: the package id part is actually clear from doc: the id has only local meaning and does not play any relevance in whether a commit succeeds or not.  From what i've seen from protobufs (protos/peer/lifecycle) seem to confirm that in that there is nothing code related in them such as a package id or alike.

regarding reachable it seems to me you are still in the go-chaincode wrapper model: if we implement our own chaincode type we can do whatever we want in the shim and it doesn't have to be in the chaincode package whatsoever: our chaincode package would just be enclave.so?

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,116 @@
...
+  
+  ECC_Enclave -> ECC_Enclave : Generate quote
+  note right
+    Quote includes ... 
+    SPID, CC_name, CC_ver, Org_i,
+    ECC_{S_PK, E_PK}, TLCC_MR,
+    TLCC_{S_PK, E_PK}, CH_id
+  end note
+
+  ECC_Enclave -> ECC_Enclave : Seal state
```
g2flyer: we might want to mention explicity what state is?  Presumably (at least), ECC_{S,E}_SK, TLCC_MRENCLAVE, CH_id?

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,116 @@
...
+  activate ECC_Enclave
+
+  ECC_Enclave -> ECC_Enclave : Generate ECC_{S_PK, S_SK, E_PK, E_SK}
+
+  ECC_Enclave -> ECC : Binding(ECC_{S_PK, E_PK})
+  ECC --> ECC_Enclave : Sig_Org_1(ECC_{S_PK, E_PK})
+
+  ECC_Enclave -> TLCC_Enclave : Binding(ECC_{S_PK, E_PK}) using local attestation
+  note over ECC_Enclave, TLCC_Enclave: establish secure channel with TLCC_Enclave
+  TLCC_Enclave --> ECC_Enclave : TLCC_MR, TLCC_{S_PK, E_PK}, CH_id
```
g2flyer: Not convinced that we need TLCC_{S,E}_PK and want to include them in attestation: clearly for the current session we don't need them, we can do (cheaply) local attestation and then based on DH generate a symmetric session key.   

 To deal with the `different chain-code engaged in common local CC2CC transaction TLCC must be same instance` issue, this key would have to be ephemeral and so each time you restart the peer you would have to update all involved ERCCs entries (which seems a hassle).  As soon as you persist the key, you cannot ensure unique instances (besides adding additional exposures of a long term key to worry about)! 

Even if you re-register,  you still would have to pass TLCC identity explicitly from caller to callee in a CC2CC call so the callee can ensure consistency.  But in that case there is a much easier solution by having TLCC just have an instance-fixed DH-key used for local key exchange you just pass that along from caller to callee?

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,116 @@
...
+
+end
+
+group register attestation
+  
+  Admin -> ECC ++: register(attestation, ECC_s_pk) proposal
+  ECC -> ECC : verify and store
+  note right
+    - check attestation signature: signature verification key defined in genesis block
+    - check attestation ECC_s_pk and mrenclave: mrenclave defined in chaincode definition
```
g2flyer: assuming that we store TLCC_mrenclave in channel definition, we also would have to verify here that what ecc got is indeed the correct one, similarly we will have to verify correctness of CH_id

docs/fpc-registration-new.puml
g2flyer: As most identifiers do not have that "content-addressable" propery, maybe it is also worth stating explicitly that CH_id is hash of channel genesis block and uniquely will  identify channel content (if orderers are trusted)?

docs/fpc-lifecycle-v2.puml
```diff
@@ -0,0 +1,129 @@
...
+	//MRENCLAVE: mrenc_1//
+end note
+
+Admin_Org1 -> Peer1_Org1 : install package
+return packageID: mycc_1:hash
+Admin_Org1 -> Peer2_Org1 : install package
+return packageID
+note right
+	chaincode package identifier
+	is mycc-1:hash
```
g2flyer: i don't think peer will necessarily return the same package id than peer1

docs/fpc-lifecycle-v2.puml
```diff
@@ -0,0 +1,129 @@
...
+Admin_Org3 -> Peer2_Org3 : install package
+return packageID
+
+
+Admin_Org1 -> Orderer : Approve
+note left
+	**Chaincode definition**
+	Name: mycc, Version: 1
+	End Policy: (2of3), Seq 1,
+	Package ID: mycc_1:hash,
```
g2flyer: i don't think there is a package id in the actual chaincode definition which is in the approve/on ledger

docs/fpc-registration-new.puml
```diff
@@ -0,0 +1,116 @@
...
+title Foot Box removed
+
+title Fabric Private Chaincode - Enclave registry - New approach
+
+actor	"Admin_Org1" as Admin order 10
+participant "ECC_Manager (syscc) Peer1_Org1" as ECC_Manager order 20
+participant "ECC Peer1_Org1" as ECC order 30
+
+note right ECC
+  ECC can be **integrated** in ECC or as **seperate** chaincode
```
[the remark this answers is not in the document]
mbrandenburger: - not sure what exactly you mean here with fundamental security problem. The current approach uses a single endorsement policy (for attestation generation) but requires all peers to validate the attestation. Anyway I am open to add here some more information but please suggest by providing a change set?! 💃 

- yes indeed copy/replace bug

docs/fpc-cc-invocation.puml
```diff
@@ -0,0 +1,158 @@
...
+        from sealed state. Remembers TLCC_inst_id.
+      end note
+      return
+    else
+      |||
+
+      ECC1 -> Enclave1 ++: invoke (args_ECC)
+
+      note right
+        Eventually we probably will also first verify signature of proposal?
```
bvavala: yes, if we have to authenticate clients based on MSP, and in this case we can rely on tlcc, which should already have (1) the data/certificates, (2) the MSP logic to validate it.

docs/fpc-validation-msp.puml
```diff
@@ -0,0 +1,142 @@
...
+    group validation & commit
+        ECC1 -> ECC1 : validate block signature
+        loop for each tx in block
+            ECC1 -> ECC1 : validate tx (client) signature
+
+            ECC1 -> ECC1 : validate endorsements policy
+            note right #FFaa00
+                - is endorsing enclave part of Org_i?
+                - has enclave a proper attestation?
+                - is this enclave also registered to
```
bvavala: here we should not overload MSP with operations it is not supposed to do, so checking the registration should be actually cast as membership check, where the membership is maintained by MSP itself.

bvavala:
` docs/initial_arch/`: it is exactly for this reason that github is not the right place for ongoing design.
The final docs should be added as things are implemented -- and ideally keep them up-to-date

mbrandenburger:
> ` docs/initial_arch/`: it is exactly for this reason that github is not the right place for ongoing design.
> The final docs should be added as things are implemented -- and ideally keep them up-to-date

mbrandenburger:
> ` docs/initial_arch/`: it is exactly for this reason that github is not the right place for ongoing design.
> The final docs should be added as things are implemented -- and ideally keep them up-to-date

I don't see you point why github is not the right place to work on this document? As it is a text document it perfect to track changes. I agree that having this `initial_arch` folder to keep the "old"` design is not the "git" way how things are developed, however, I think it is good to keep it as a reference. This allows us aways go back and look at it without using git command.

For the work on the design I can imagine that it would also live very well in a different branch, thereby the current design in that branch does not reflect the state of the implementation of whatever we have in master. Once designs are final, it can be merged to master. That's ok.

mbrandenburger:
@bvavala let's merge it to the new design branch.

PR #122, opened 2019-07-31
Fix concurrency issue in registry

mbehlendorf:
In cases of high load the evaluation of r.index took place after it was incremented by another invocation.
This led to two invocations using the same index and finally a crash.

Now the lock is released after the evaluation of r.index.


g2flyer:
Great find!!

Issue #88, opened 2019-06-15
logging should be in the shim, not in the auction


bvavala:

mbrandenburger:
here we have the reference shim we need to follow. https://github.com/hyperledger/fabric/blob/13447bf5ead693f07285ce63a1903c5d0d25f096/core/chaincode/shim/interfaces_stable.go

I guess what you are supposing is, that we want to extend the shim in a way that the chaincode developer can perform logging easily.

Currently, a chaincode developer can just include `common/logging.h` and use `LOG_DEBUG`, `LOG_INFO`, and `LOG_ERROR`.

g2flyer:
Clearly we don't want to add new bell's and whistle for logging, just (a) i think FPC chaincode should only use services from shim.h when interacting with infrastructure and (b) the logging namespace should be separated between logging done by the shim itself vs logging done by the FPC chaincode.

mbrandenburger:
yes, this sounds right to me

g2flyer:
On a related note: while logging functionality should move to shim.h, shim.h seems to contain a number of function which seem internal and not targeted to chaincode (e.g., all the rwset functions). They should be moved out of the shim.   Also, if we have unmarshall functions for inputs, shouldn't we have corresponding marshall functions for results? Seems strange to have only one but not the other ...

Lastly (but more long-term), we probably also want some put_public_state function?

g2flyer:
Observations on #137 prototypes by Marcus ...
> Seems this is aligned with Fabric 1.4.x. This is good.
>
> In Fabric 2.0, however, chaincode logging is the responsibility by the chaincode developer.
https://hyperledger-fabric.readthedocs.io/en/master/logging-control.html
>
> Since this is not easily possible with FPC I still think its a good idea to provide this logging to the user. I believe we can only reduce the number of different log levels we want to support. Maybe just ERROR | WARNING | INFO | DEBUG ?

We definitely still should have logging-level edl functions as eventually with our own untrusted stub/shim we still want to be able to route (trusted) shim log-messages to peer. Given that we do not really have a full libc available, i think we also still would have to provide some logging functionality to user, but then on untrusted side might just throw it directly to stdout (rather than as right now to logger which then passes it to stdout ..)

g2flyer:
[...]

PR #137, opened 2019-08-13
Shim refactoring, phase 2

g2flyer:
This  PR addresses (most of) the remaining issues from issue #83 -- issue is updated on still (potentially) remaining issues -- but not yet issue #88 (shim.h has now a tentative interface but no implementation yet.).

More specifically:
- factor out shim-internal stuff into separate file (`shim_internals.h`) and consolidate public stuff from `chaincode.h` into `shim.h`, making the latter the sole API for FPC CC programmers.
- adds (simplified) get_creator-like functionality
- adds init behaviour as in standard chaincode
- hides all marshalling and json encoding beyond what you need for peer cli  (i.e., client passes now args as `{"args": ["arg1", "arg2", ...]}` rather than `{"args":["[\"arg1\", \"arg2\", ... ]"]}` ...)

Also contains some other more or less minor changes:
- more context in C logging
- swap linter and licence so that brittle license test is always last
- more sane buffer management in get_state


examples/auction/auction_cc.cpp
```diff
@@ -48,8 +48,18 @@ int invoke(const char* args,
...
     {
-        std::string bidder_name = argss[2];
         int value = std::stoi(argss[3]);
+        std::string bidder_name = argss[2];
+        // TODO: eventually replace bidder_name with get_creator_name but for now
+        //  in our tests we have only one client, so leave passed bidder_name to
+        //  allow for different bidders ...
+        char real_bidder_name_msp_id[1024];
+        char real_bidder_name_dn[1024];
+        get_creator_name(real_bidder_name_msp_id, sizeof(real_bidder_name_msp_id),
```
bvavala: This input is not validated.
Most importantly, I don't see an easy way to do that. Even if you can verify the MSP id and the certificate, it does not mean that you have the creator's name.

g2flyer: Yes, the current implementation is not secure (as is many other things right now :-) and the goal is primarily to fix API.  I should have added a corresponding TODO in shim.ccp, though, which i just did.
Regarding implementation: There is definition no easy-way (or any?) with the current way of piggy-backing on top of the go-shim but with a proper chaincode type we can have access to the signed proposal in the enclave (see above-referenced TODO). The verification of the whole thing is certainly non-trivial but we more or less have to do that anyway inside TLCC so we can just off-load it to TLCC.  Lastly, the creator name is by definition the distinguished name in the certificate, so if we have the certificate, we have the name ...

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
-void put_state(const char* key, uint8_t* val, uint32_t val_len, void* ctx);
+// TODO: Some Meta-questions
+// - should we prefix functions with fpc_ or alike to make them more distinct?
+//   Alternatively, we might go C++ and define a shim object like the go shim
+//   (but then maybe we do not necessarily want to force folks to go to C++ and
+//    still enable C?)
+
+// Function which FPC chaincode has to implement
+// ==================================================
+int invoke(const char* args,
```
mbrandenburger: why moving this from chaincode.h to here?

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
-typedef std::set<std::string> read_set_t;
-typedef std::map<void*, std::pair<read_set_t*, write_set_t*>> context_t;
+typedef void* fpc_ctx_t;
 
-// shim put/get
-void get_state(const char* key, uint8_t* val, uint32_t max_val_len, uint32_t* val_len, void* ctx);
-void put_state(const char* key, uint8_t* val, uint32_t val_len, void* ctx);
+// TODO: Some Meta-questions
+// - should we prefix functions with fpc_ or alike to make them more distinct?
+//   Alternatively, we might go C++ and define a shim object like the go shim
```
mbrandenburger: ack! this would be much closer to the existing go programming model; 
https://godoc.org/github.com/hyperledger/fabric/core/chaincode/shim#ChaincodeStubInterface

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
+//   (but then maybe we do not necessarily want to force folks to go to C++ and
```
mbrandenburger: I believe forcing people to use c++ is ok;

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
+int invoke(const char* args,
+    uint8_t* response,
+    uint32_t max_response_len,
+    uint32_t* actual_response_len,
+    fpc_ctx_t ctx);
+
+/*
+  TODO: compatibility with fabric programming model, we might want to add also 
+
+int init(const char* args,
```
mbrandenburger: yes that would be nice; however, if we are going to support init we may need to reconsider the enclave registration protocol; as we need an registered enclave to perform init; moreover, with the new lifecycle we can also say init is not supported as one can define this in a chaincode definition

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
+//   (hence been better able to exploit the side-channel). Insofar, it doesn't
+//   seem to me worth to expose it, at least until wee have a concrete use-case
+//   requiring it?
+//
+// - '*_semi_public_state*' variant of '*_super_private_state*' where data is
+//   stored unencrypted in a private data collection
+
+// (un)marshalling
+//-------------------------------------------------
 int unmarshal_args(std::vector<std::string>& argss, const char* json_string);
```
mbrandenburger: I have the feeling that the shim should already return unmarshaled args to the chaincode.
removing the args from the invoke and provide something similar to `GetArgs() [][]byte` (https://godoc.org/github.com/hyperledger/fabric/core/chaincode/shim#ChaincodeStubInterface)

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
 int unmarshal_values(
     std::map<std::string, std::string>& values, const char* json_bytes, uint32_t json_len);
 
-// read/writeset
-void register_rwset(void* ctx, read_set_t* readset, write_set_t* writeset);
-void free_rwset(void* ctx);
-read_set_t* get_read_set(context_t* context, void* ctx);
-write_set_t* get_write_set(context_t* context, void* ctx);
+// TODO (maybe): also corresponding marshall functions for return values?
```
mbrandenburger: should we return a proper pb.Response instead of json?

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
+//   - macro vs function?
+//     - function looks kind of cleaner.
+//     - macros would allow passing context such as file & linenumber
+// - implementation
+//   - i would remove the DO_* from logging.h and replace the printfs with a
+//     golog which would pass the level to go. that way normal log-enablement
+//     would work.
+//   - a somewhat orthogonal question is whether in production mode
+//     a chaincode should ever log anything to the outside, but i think
+//     that should be handled on the implementation side hiden from the API here
```
mbrandenburger: agree

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,130 @@
...
+void log_critical(const char* format, ...);
+void log_error(const char* format, ...);
+void log_warning(const char* format, ...);
+void log_notice(const char* format, ...);
+void log_info(const char* format, ...);
+void log_debug(const char* format, ...);
+// TODO
+// - API design questions
+//   - macro vs function?
+//     - function looks kind of cleaner.
```
mbrandenburger: I don't have a preference here.

ecc_enclave/enclave/shim.h
g2flyer: aeh, because i didn't even know of the existance of chaincode.h: none of the examples was using it ...
Seems, though, that only invoke but not inovke_enc would be exposed?  If so, why not move invoke to shim.h and invoke_enc to shim_internals.h?  Do we need two header files exposed to the application?

g2flyer: invoke_enc doesn't even really need a function prototype, in particular not in a separate header file, as it is used and defined in same file, so i just removed chaincode.h ..

ecc_enclave/enclave/shim.h
g2flyer: why is this not supported?  In new lifecycle you can specify init to be optional or not, but i don't see why this would be ruled out?  Obviously you have to run it after enclave registration (and key-management). It's true with current chaincode type it's a bit messy but i think even there it would work: we could (and should) add the setup calls also to the peer instantiate wrapper and then we can call init also from the wrapper. for the go cc of course the init would look then like an invoke but that would be all under the cover?

ecc_enclave/enclave/shim.h
g2flyer: That was also were i was coming for. But after discussion this morning i made executive decision that for now we leave out C++ and stay with C as (a) that's what we already have, (b) sooner or later (though probably rather later than sooner :-) we might go to a different language like WASM and (c) i'm not super-fond of (nor super-proficient in) in C++ ..

ecc_enclave/enclave/shim.h
g2flyer: I was thinking of protobuf as well.  However, after the discussion this morning i think the right way is to provide no marshalling/unmarshalling of parameters at all but provide arguments as an array of strings. This would be fairly easy and more consistent with the other CC apis (e.g., node sdk also handles only string args) and presumably also the client cli ..).

ecc_enclave/enclave/shim.h
sjanakir2015: - First, reg. input.  
Does the arguments have to be in _json_?  Particularly, when that is not the case with Fabric.

case 1:   If json is enforced for arguments, shim could provide unmarshalled args.  The chaincode receives what its client sends.  

case 2:  if json is not enforced for arguments, unmarshalling cannot be done; but can be made available for the following reason. 

For general purpose code, a user can marshall / unmarshall using his choice of library.  Since this is SGX chaincode, the user is, kind of, limited to what libraries he can use.  For this reason, providing an `unmarshall_args` is user friendly.  

- Second, output i.e. result
I dont think we should look for symmetry in this case.  Can we provide a (nice) generalized marshall api in cpp?

ecc_enclave/enclave/shim.h
sjanakir2015: Agree - to provide args as an array of strings. 
- First, reg. input.  
To be consistent with Fabric shim api, arguments can be of `array of byte arrays` or `array of strings`.   Since json is not the only format,  unmarshalling cannot be done; but can be made available for the following reason. 
For general purpose code, a user can marshall / unmarshall using his choice of library.  Since this is SGX chaincode, the user is, kind of, limited to what libraries he can use.  For this reason, providing an `unmarshall_args` is user friendly.  

Reg. output i.e. to provide api to marshall result:
I dont think we should look for symmetry in this case.  Not sure if we can we provide a (nice) generalized marshall api in cpp.

g2flyer: >   To be consistent with Fabric shim api, arguments can be of `array of byte arrays` or `array of strings`.  

That's true for the go shim. The node shim does have both `GetArgs` and `GetStringArgs` but both return a string version!  Anyway, i guess for us in C string vs bytes shouldn't fundamentally be an issue from passing it along.  But if we pass args in invoke (`main`-like) we would have pick one vs. the other.  Short-term i think we would please more with string-variant?

> Since this is SGX chaincode, the user is, kind of, limited to what libraries he can use. 

For marshalling/unmarshalling libraries, i would have expected many should not be too hard to make work with SGX or might work even out of the box. Why do you think it is limited?

> * Reg. output i.e. to provide api to marshall result:
>   I dont think we should look for symmetry in this case.  Not sure if we can we provide a (nice) generalized marshall api in cpp.

Curious, as you seem even with string/byte-arguments, having an additional `unmarshal` function, why wouldn't you see the same value for a `marshall`. After all you probably in the end as likely to send data into the CC as you receive from the CC?   I certainly can see value where json makes live much easier and so providing a library  can make sense but i would argue that in particular marshalling would be interested as for input complex values can be decomposed in the arg list but the return is a single value so you have no choice but encode it if you have something more complicted?

The other question popping up writing this: if we see marshalling only as a convience, then it's also debatable whether it should be in shim or we might not just have a (set of) additional libraries we might eventually offer for convenience (i.e., saving the developer from finding some sgx-enabled library for some given functionality), one being a json encoding/decoding library?

sjanakir2015: > > To be consistent with Fabric shim api, arguments can be of `array of byte arrays` or `array of strings`.
> 
> That's true for the go shim. The node shim does have both `GetArgs` and `GetStringArgs` but both return a string version! Anyway, i guess for us in C string vs bytes shouldn't fundamentally be an issue from passing it along. But if we pass args in invoke (`main`-like) we would have pick one vs. the other. Short-term i think we would please more with string-variant?

I agree
> 
> > Since this is SGX chaincode, the user is, kind of, limited to what libraries he can use.
> 
> For marshalling/unmarshalling libraries, i would have expected many should not be too hard to make work with SGX or might work even out of the box. Why do you think it is limited?
> 
I thought it may be an issue to add  an additional (marshall / unmarshall) library  to make it work with SGX.  Also, as a chaincode developer, it is recommended to keep the code (running inside TEE) to be small. One less library.  I am not sure if it matters. 

> > * Reg. output i.e. to provide api to marshall result:
> >   I dont think we should look for symmetry in this case.  Not sure if we can we provide a (nice) generalized marshall api in cpp.
> 
> Curious, as you seem even with string/byte-arguments, having an additional `unmarshal` function, why wouldn't you see the same value for a `marshall`. After all you probably in the end as likely to send data into the CC as you receive from the CC? I certainly can see value where json makes live much easier and so providing a library can make sense but i would argue that in particular marshalling would be interested as for input complex values can be decomposed in the arg list but the return is a single value so you have no choice but encode it if you have something more complicted?
> 
> The other question popping up writing this: if we see marshalling only as a convience, then it's also debatable whether it should be in shim or we might not just have a (set of) additional libraries we might eventually offer for convenience (i.e., saving the developer from finding some sgx-enabled library for some given functionality), one being a json encoding/decoding library?

Agree completely ! 
My first thought was like-wise:  a client who sends in JSON would like to receive JSON too.  Similar handling.  JSON works well for complex input and output.    I agree with the option of providing a _library_ for json marshall _and_ unmarshall.  
Not as (a method in) api.

ecc_enclave/enclave/enclave.cpp
```diff
@@ -67,14 +69,117 @@ int ecall_bind_tlcc(const sgx_report_t* report, const uint8_t* pubkey)
...
+    uint8_t* response,
+    uint32_t* response_len_out,
+    sgx_ec256_signature_t* signature,
+    shim_ctx_ptr_t ctx)
+{
+    int ret;
+
+    // create Hash <- H(type || args || result || read-write set)
+    // TODO: somewhere there should be matching verification code which originally
+    //    did not have 'type' in hash. However, i couldn't find it (and it also does not fail)
```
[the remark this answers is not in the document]
g2flyer: signature validation is now fixed.   There seem to be a strange "ERROR [Enclave] Ledger: ecc signature validaion failed" in tlcc.  It happens only for auction test and not echo, only for create but also not for all of them (but always for the same one!).  As we ignore signature validation errors in tlcc, you’ll see it only in the log-file, no program/script will fail due to it.  I couldn't figure out what goes wrong but as a side effect there is a bit more complete and systematic logging.  As i eventually realized that this problem already existed in master, i've ignored it for now ...

ecc_enclave/enclave/shim.cpp
```diff
@@ -87,22 +106,41 @@ void put_state(const char* key, uint8_t* val, uint32_t val_len, void* ctx)
     std::string base64 = base64_encode((unsigned char*)cipher, cipher_len);
 
     // write state
-    write_set_t* write_set = get_write_set(&context, ctx);
-    write_set->insert({key, base64});
-    ocall_put_state(key, (uint8_t*)base64.c_str(), base64.size(), ctx);
+    ctx->write_set.insert({key, base64});
```
mbrandenburger: This looks so much better! cool

ecc_enclave/enclave/shim.h
```diff
@@ -6,27 +6,167 @@
...
+//---------------------------
+// invokeChaincode
+// TODO (possible extensions): Currently not supported (but eventually should)
 
-// read/writeset
-void register_rwset(void* ctx, read_set_t* readset, write_set_t* writeset);
-void free_rwset(void* ctx);
-read_set_t* get_read_set(context_t* context, void* ctx);
-write_set_t* get_write_set(context_t* context, void* ctx);
+// logging
```
mbrandenburger: Seems this is aligned with Fabric 1.4.x. This is good.

In Fabric 2.0, however, chaincode logging is the responsibility by the chaincode developer.
https://hyperledger-fabric.readthedocs.io/en/master/logging-control.html

Since this is not easily possible with FPC I still think its a good idea to provide this logging to the user. I believe we can only reduce the number of different log levels we want to support. Maybe just ERROR | WARNING | INFO | DEBUG ?

examples/auction/auction_cc.cpp
```diff
@@ -21,25 +22,96 @@
...
+    LOG_DEBUG("AuctionCC: Args: %s",
+        (argss.size() < 1
+                ? "(none)"
+                : std::accumulate(std::next(argss.begin()), argss.end(), argss[0],
+                      [](std::string a, std::string b) { return (a + std::string(", ") + b); })
+                      .c_str()));
+
+    const char* _auction_house_name = argss[0].c_str();
+    put_state(
+        AUCTION_HOUSE_NAME_KEY, (uint8_t*)_auction_house_name, strlen(_auction_house_name), ctx);
```
mbrandenburger: Great we have an auction house name now!

examples/auction/auction_cc.cpp
```diff
@@ -21,25 +22,96 @@
...
+        {
+            _auction_house_name_buf[ahn_len + 1] = '\0';
+            _auction_house_name = _auction_house_name_buf;
+        }
+    }
+
+    LOG_DEBUG(
+        "AuctionCC: +++ Executing '%s' auction chaincode invocation +++", _auction_house_name);
+
+    if (!_initialized)
```
mbrandenburger: This is tricky! So this code only works if init is called. This means that this chaincode only works with lifecycle v2 with init enabled. 

Also, ... invocation of init should be enforced by the peer. If the peer is correct, this code will never be reached?!

tlcc_enclave/enclave/ecc_json.cpp
```diff
@@ -8,6 +8,25 @@
...
+    {
+        json_array_append_string(list, argss[i].c_str());
+    }
+
+    char* ss = json_serialize_to_string(root);
+    encoded_string.append(ss);
+
+    json_free_serialized_string(ss);
+    json_value_free(root);
+    return 1;
```
mbrandenburger: we always return 1 here. is there also a chance that parson returns an error we might want to return here? otherwise marshal_ecc_args may also return void?!

examples/auction/auction_cc.cpp
```diff
@@ -21,25 +22,96 @@
 #define AUCTION_ALREADY_CLOSED "AUCTION_ALREADY_CLOSED"
 #define AUCTION_STILL_OPEN "AUCTION_STILL_OPEN"
 
-static std::string SEP = ".";
-static std::string PREFIX = SEP + "somePrefix" + SEP;
+#define INITALIZED_KEY "__initalized"
```
mbrandenburger: Isn't it ```#define INITIALIZED_KEY "__initialized"```

ecc_enclave/enclave/enclave.cpp
```diff
@@ -67,14 +69,123 @@ int ecall_bind_tlcc(const sgx_report_t* report, const uint8_t* pubkey)
...
+
+    // create Hash <- H(txType in {"init", "invoke"} || encoded_args || result || read set || write
+    // set)
+    // TODO: we should encode the hash below in an unambiguous fashion (which is not true with
+    //    simple concatenation as done below!)
+    //    Probably easiest by prefixing each field by length in fixed-size format?
+    sgx_sha256_hash_t hash;
+    sgx_sha_state_handle_t sha_handle;
+    sgx_sha256_init(&sha_handle);
+    LOG_DEBUG("txType: %s\n", txType);
```
mbrandenburger: `\n` leads to extra new lines which are not needed here. For all LOG_DEBUG statements no need for \b

ecc/crypto/ecdsa.go
```diff
@@ -144,20 +149,31 @@ func (v *ECDSAVerifier) Verify(args, responseData []byte, readset, writeset [][]
...
 	}
 
-	// H(args || response || readset || writeset)
+	// Note: below signature was created in ecc_enclave/enclave/enclave.cpp::gen_response
+	// see also replicated verification in tlcc_enclave/enclave/ledger.cpp::int parse_endorser_transaction (for TLCC)
+
+	// H(txType in {"init", "invoke"} || encoded_args || response || read set || write set)
 	h := sha256.New()
-	h.Write(args)
+	logger.Debugf("txType: %s\n", string(txType))
```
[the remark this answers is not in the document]
g2flyer: alas, there is none in fabric:: the levels specified (but not yet implemented) are all there is (https://godoc.org/github.com/op/go-logging#Level

ecc_enclave/enclave/enclave.cpp
```diff
@@ -33,8 +33,10 @@ sgx_aes_gcm_128bit_key_t state_encryption_key = {
...
 int ecall_bind_tlcc(const sgx_report_t* report, const uint8_t* pubkey)
 {
+    LOG_DEBUG("ecall_bind_tlcc: \tArgs: &report=%p, &pk=%p", report, pubkey);
+
     // IMPORTANT!!!
-    // here is out testing backdoor for starting ecc without a tlcc instance
+    // here is our testing backdoor for starting ecc without a tlcc instance
     if (report == NULL && pubkey == NULL)
     {
         LOG_DEBUG("Start without TLCC!!!!");
```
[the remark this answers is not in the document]
g2flyer: makes sense (this was though your DEBUG, i think; but i've changed it now anyway  ;-)

ecc_enclave/enclave/shim.h
g2flyer: thanks for pointing that out, will consider this when going last stretch with shim revamp and implement logging.  We definitely still should have logging-level edl functions as eventually with our own untrusted stub/shim we still want to be able to route (trusted) shim log-messages to peer.  Given that we do not really have a full libc available, i think we also still would have to provide some logging functionality to user, but then on untrusted side might just throw it directly to stdout (rather than as right now to logger which then passes it to stdout ..)

examples/auction/auction_cc.cpp
g2flyer: not necessarily, init could also have failed?  In any case, if with lifecycle v2 then you definitely must enforce such initialized aspects yourself in the code if your state really depends on it ?

tlcc_enclave/enclave/ecc_json.cpp
g2flyer: I followed the precedent of the other already existing function which also doesn't do error checking.  If i would have had more time and the PR wouldn't have been already that big i might have changed it already here but now it has to wait for another occasion  ...

examples/auction/auction_cc.cpp
g2flyer: definitely a typo, but being a constant at least not leading to bugs :-) (i've fixed it now, though)

ecc_enclave/enclave/enclave.cpp
g2flyer: i think it got there from copy/paste from existing log statement. but eradicated now all \n throughout code.   Interestingly, i also never really noticed any empty lines in the logs, though.

PR #140, opened 2019-08-14
Fix to makefiles identified in review of issue 128

g2flyer:
addresses 1) and 3) of issues identified by Ann on 2019-08-14

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ include $(TOP)/build.mk
 
 SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
 
+build : godeps
+
 build test clean :
 	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
 
@@ -27,3 +29,11 @@ linter: gotools build
 gotools:
 	# install goimports if not present
 	$(GO) get golang.org/x/tools/cmd/goimports
+
+godeps:
+	$(GO) get github.com/spf13/viper
+	$(GO) get golang.org/x/sync/semaphore
+	$(GO) get github.com/pkg/errors
+	$(GO) get github.com/golang/protobuf/proto
+	$(GO) get github.com/onsi/ginkgo
+	$(GO) get github.com/onsi/gomega
--- a/ercc/Makefile
+++ b/ercc/Makefile
@@ -11,7 +11,7 @@ VSCC_OUT = ercc-vscc.so
 DECO_SRC = attestation/ias_credentials/decoration.go
 DECO_OUT = ercc-decorator.so
 
-build: provider ercc vscc-plugin decorator-plugin
+build: ias-keys provider ercc vscc-plugin decorator-plugin
 
 .PHONY: .sgx_mode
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -27,3 +29,11 @@ linter: gotools build
 gotools:
 	# install goimports if not present
 	$(GO) get golang.org/x/tools/cmd/goimports
+
+godeps:
```
bvavala: much cleaner!

ercc/Makefile
```diff
@@ -32,6 +32,16 @@ else
 SGX_MODE_DEPENDENCY=
 endif
 
+ias-keys:
```
bvavala: what if I tested HW, then SIM, then HW?
This overwrites my keys and I need to start over with the configuration but... only I after I find out that it does not work!!!

g2flyer: no, it does not override, it creates files only if they do not exist (at least that's the intention)

bvavala: point to you! :P

New round: if I test in SIM then HW, files already exist but...they are wrong.
So, should we separate these credentials?

Makefile
```diff
@@ -8,6 +8,8 @@ include $(TOP)/build.mk
 
 SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
 
+build : godeps
```
mbrandenburger: mr. picky always complains about spaces :D

ercc/Makefile
bvavala: another solution is in ercc/makefile: create .sgx_mode file and put a dependency
or maybe we could graduate that file to the top folder

g2flyer: different files requires different fabric configs which i don't think is an appealing solution, at least not better than this one. Your case is valid but only for somebody who hasn't followed the instruction. 

iff we get access to SGX_MODE in ercc/attestation/ias_credentials/decoration.go we also could just do it there but not easy via go build parameters (and i don't really like the templating stuff to extend it here as well). I was thinking first on whether could do it based on the attestation type (which for sim always should be mock?) but somehow didn't see exactly how i access it in decoration.go.

So for now i still will leave it as-is ..

bvavala: > Your case is valid but only for somebody who hasn't followed the instruction.
I disagree. 
As a user, _I_ should follow the instruction, and _the framework_ should warn me properly when things are not set up properly.
Here there are four problems:
1) the connection to IAS will fail - and the user has to figure out that this is due to the credentials
2) the mock SIM credentials, created by the framework, are accepted by the framework in HW mode.
3) this has already happened in PDO!

> iff we get access to SGX_MODE in ercc/attestation/ias_credentials/decoration.go we also could just do it there but not easy via go build parameters (and i don't really like the templating stuff to extend it here as well). I was thinking first on whether could do it based on the attestation type (which for sim always should be mock?) but somehow didn't see exactly how i access it in decoration.go.

This is complicated!

g2flyer: well, i'm not arguing this PR is a perfect solution but it is (still) the best simple fix i know of to address the issue identified by Ann.  It is certainly better than the status-quo where the documentation was misleading  :-)

The related code almost certainly also will change quite a bit with new ercc, proper chaincode where the decorators should go away and the values are directly accessed from system cc and alike, so we are talking anyway only of an intermediary fix.

g2flyer:
as discussed on call today, i did add now a section to the trouble shooting section.  Unfortunately, the content of the dummy files does not influence the error message and so cannot lead to a more explicit error message. But from output the error is very obvious to spot and then should be easy to match to the troubleshooting section.

Issue #143, opened 2019-08-19
fix test in tlcc_enclave


g2flyer:
make tlcc test case reproducible (or at least documented) and adapt to cover case which caused issue #139

g2flyer:
Partially addressed now in PR #144:  Test is re-enabled and new test-cases are generated using  target `gen_testcases` in `tlcc_enclave`.  This in turn relies on the deployment_test scenario and the script `tlcc_enclave/test/getLedgerBlocks`.

However, the test does  not seem to really testing much as in logs i see error but nothing causes the test to fail (not even an out-of-order block!)

Issue #139, opened 2019-08-14
TLCC channel meta-data parsing is broken


g2flyer:
We currently get tons of  errors in tlccs of the form
```
... tlcc_enclave/enclave/crypto.cpp:102) [Enclave] [31mCrypto: Invalid certificate: unable to get local issuer certificate[0m
... tlcc_enclave/enclave/ledger.cpp:215) [Enclave] [31mLedger: Envelope signature validation failed[0m
```
The root cause seems to be that we are not parsing properly anymore the channel config infos anymore ...
```
tlcc_enclave/enclave/ledger.cpp:74) [Enclave] [33mLedger: Block number: 0[0m
tlcc_enclave/enclave/ledger.cpp:165) [Enclave] [33mLedger: Process Envelope[1/1][0m
tlcc_enclave/enclave/ledger.cpp:187) [Enclave] [33mLedger: 	\-> Verify envelope signature[0m
tlcc_enclave/enclave/ledger.cpp:198) [Enclave] [33mLedger: 		\-> MSPID: SampleOrg[0m
tlcc_enclave/enclave/ledger.cpp:224) [Enclave] [33mLedger: Skip signature validation; genesis block[0m
tlcc_enclave/enclave/ledger.cpp:310) [Enclave] [33mLedger: ### Parse Config ###[0m
tlcc_enclave/enclave/ledger.cpp:315) [Enclave] [33mLedger: ConfigEnv.config.ChannelGroup.Groups:[0m
tlcc_enclave/enclave/ledger.cpp:320) [Enclave] [33mLedger: 	Group: Orderer[0m
tlcc_enclave/enclave/ledger.cpp:341) [Enclave] [33mLedger: 		Org: SampleOrg[0m
tlcc_enclave/enclave/ledger.cpp:353) [Enclave] [33mLedger: 			MSP Config: SampleOrg[0m
tlcc_enclave/enclave/ledger.cpp:355) [Enclave] [33mLedger: 			\-> Root certs: 1[0m
tlcc_enclave/enclave/ledger.cpp:365) [Enclave] [33mLedger: 			\-> Admin certs: 1[0m
tlcc_enclave/enclave/ledger.cpp:347) [Enclave] [31mLedger: Can not decode protobuffer[0m
```
... maybe due to outdated protobuf files?

[...]

PR #123, opened 2019-08-01
upgrade to fabric v1.4.2 & config upgrade from old release-1.3 version

g2flyer:
looking at some configtx options i realized that we still were using older 1.3 configs. Given that the comments in these yaml files are really the main documentation for available parameters in these files (at least i couldn't find anything detailed in the normal docu ...) i thought i upgraded them to current version.  Belatedly i also noticed that i was using the latest released version of config which happens to be 1.4.2 and the configs for that do not run with 1.4.1.  But keeping track with latest released version seems to make anyway sense, so i've tested with 1.4.2 -- works fine .. -- and updated also docu to refer to 1.4.2 ... 

Signed-off-by: Michael Steiner <michael.steiner@intel.com>

Issue #141, opened 2019-08-15
Hide setup call in peer wrapper


g2flyer:
Setup should not be a user-visible function and neither does the retrieval of the enclavePK. So our example flows shouldn't have to do ...
```
    try ${PEER_CMD} chaincode invoke -o ${ORDERER_ADDR} -C ${CHAN_ID} -n ${CC_ID} -c '{"Args":["setup", "ercc"]}' --waitForEvent

    try ${PEER_CMD} chaincode query -o ${ORDERER_ADDR} -C ${CHAN_ID} -n ${CC_ID} -c '{"Args":["getEnclavePk"]}'
```
While the second call is anyway not functionally necessary before we do e2e encryption, the former is necessary.   However, we can do that in the peer wrapper as part of the chaincode instantiate interception:  Just call `setup` automatically after a successfull instantiaton. 

To keep namespaces somewhat more seperate between such internal messages and user-space invokcations (and inspired by C compilers), we might call function like this setup `__setup`

mbrandenburger:
I support making `setup` a FPC operation and removing it's user visibility.

Regarding `init` we should omitting it for the moment and focus on other parts until we really have to support it.

mbrandenburger:
Ideally, setup is provided by the Enclave Manager (see new ercc design diagrams) and calls ecc/ or the enclave without leveraging the chaincode interface. This is easy once we have a "custom" chaincode type - or an extended chaincode shim.

g2flyer:
> Regarding `init` we should omitting it for the moment and focus on other parts until we really have to support it.

"outsourced" in separate issue #145 to keep discussions separate

g2flyer:
closed in PR #144

g2flyer:
closed in PR #144

PR #144, opened 2019-08-19
Hide enclave setup functions in peer cli wrapper

g2flyer:
This addresses the first half of issue #123 

Additionally, this also separate more cleanly the namespace of internal functions  from the namespaces of functions passed to FPC chaincode.  This right now causes tlcc_enclave to fail with now quick fix. Given that tlcc_enclave anyway has to be fixed (issue #139), it can be best addressed in that context (there is a corresponding issue #143)


ecc/enclave_chaincode.go
```diff
@@ -61,13 +61,21 @@ func (t *EnclaveChaincode) Invoke(stub shim.ChaincodeStubInterface) pb.Response
...
+		// Note: above string is also used in ecc_valdation_logic.go,
+		// so if you change here you also will have to change there ...
+		// If/when we refactor we should define such stuff somewhere as constants..
 		return t.setup(stub)
-	} else if function == "getEnclavePk" { //get Enclave PK
+	} else if function == "__getEnclavePk" { //get Enclave PK
 		return t.getEnclavePk(stub)
-	} else {
-		return t.invoke(stub)
+	} else if function == "__init" { // call FPC's init in enclave
```
mbrandenburger: why is fpc "__init" invoked per "fabric invoke" instead of "fabric init"?

g2flyer: not sure i understand: `init` is implicitly called during instantiate but the go-chaincode `init` is of course too early to do anything in the enclave? So we have no choice other than call it through `invoke`?  (ecc of course would then forward the `__init` payload to the enclave's `init`

g2flyer:
> Regarding init, let's use #141 to elaborate on this topic first.

FYI: Moved from #141 to #145

g2flyer:
tlcc_enclave test is now re-enabled. See commit log for what and what is not fixed related to issue #143

ecc/enclave_chaincode.go
bvavala: do we need `__init` at this point, or should it rather be our `setup`?
In the former case, I would drop it. In the latter case, I would rename `setup`

tlcc_enclave/test/test.c
```diff
@@ -8,6 +8,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <glob.h>
```
bvavala: I don't see this `glob.h`

g2flyer: this is in `/usr/include/glob.h`.  and should be standard glibc.  On four different systems (including on some wheree i've never installed pdo and/or fpc) it is part of libc6-dev 
```
msteiner@ada2 ~$ dpkg-query -S /usr/include/glob.h
libc6-dev:amd64: /usr/include/glob.h
```

ecc/enclave_chaincode.go
g2flyer: @Bruno: Ok, init is now eradicated from the PR (overall took me almost same time as implementing init e2e).   (see https://github.com/hyperledger-labs/fabric-private-chaincode/compare/49e97fe1791720a48915236f3dde544fcd765156..519c4db2d4b6d589d00aa3c791baa12f74c26c2e).

Following Marcus' dictum this morning , i've changed the assignee to you.

As Ann is waiting for this PR, would be good to get reviews in soon

fabric/bin/peer.sh
```diff
@@ -136,7 +136,23 @@ incode/go/example02/cmd' "${OTHER_ARGS[@]}" )
...
+    # Note: we rely on above to analyze args and assign (& translate) CC_NAME for -n and -c init-args CC_MSG and IS_FPC_CC
+
+    # - instantiate chaincode
+    try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
+    # Note: the -c init-args which ultimately have to go to the enclave are also passed here: ecc just ignores them, so no point to filter them. Besides we need it if CC is _not_ FPC ..
+    sleep 3 # unfortunately, no --waitForEvent equivalent, so do heuristic sleep ...
+
+    if [ ${IS_FPC_CC} -eq 1 ]; then
+        # - setup internal ecc state, e.g., register chaincode
+        try $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke -o ${ORDERER_ADDR} -C ${CHAN_ID} -n ${CC_NAME} -c '{"Args":["__setup", "'${ERCC_ID}'"]}' --waitForEvent
```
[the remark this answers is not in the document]
g2flyer: You have a point. However, right now we don't have multi-peer and i also strongly feel for the sdk milestone we should hide setup as it will (should) not be there in the future.  Even in multi-peer setting we still can handle that in the wrapper by persisting info on whether setup has happened and check before invoke and query whether it has happened and if not, call it there.    I think this should be done together with multi-peer, though. Have updated issue #98 accordingly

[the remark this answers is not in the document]
mbrandenburger: Ok, let's go forward with this strategy. Could you please add a TODO in the cli wrapper for invoke and query setup must be called if needed and link to #98 so we don't miss that?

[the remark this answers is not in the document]
g2flyer: oh, well, if i anyway have to update the patch, i've just fixed it (and as a side-effect also fixed a few bugs in the scripts and made them also more versatile, in particular if you use the interactive version with ledger_init.sh ....

PR #146, opened 2019-08-21
Integration test fixes

annumberhocker:
Addressing Issue 104: 
I have improved the auction test to check all error codes and scenarios.  I added the code to extract out the actual response payload from the info returned from the peer, and a function to be able to test on the response code.  

I track how many failures occur and report at the end.

Ultimately, we will have a better integration test, but this will work for now.  I am pushing the code for auction first and will upgrade echo next.

I also added a green text function to the fabric utilities.

### DIFF
--- a/integration/auction_test.sh
+++ b/integration/auction_test.sh
@@ -4,7 +4,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-#set -x
 SCRIPTDIR="$(dirname $(readlink --canonicalize ${BASH_SOURCE}))"
 FPC_TOP_DIR="${SCRIPTDIR}/.."
 FABRIC_SCRIPTDIR="${FPC_TOP_DIR}/fabric/bin/"
@@ -22,9 +21,9 @@ ENCLAVE_SO_PATH=examples/auction/_build/lib/
 CC_VERS=0
 num_rounds=3
 num_clients=10
+FAILURES=0
 
 auction_test() {
-
     # install, init, and register (auction) chaincode
     try ${PEER_CMD} chaincode install -l fpc-c -n ${CC_ID} -v ${CC_VERS} -p ${ENCLAVE_SO_PATH}
     sleep 3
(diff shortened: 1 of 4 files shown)


integration/auction_test.sh
```diff
@@ -21,37 +24,103 @@ CC_VERS=0
 num_rounds=3
 num_clients=10
 
+# Try function which returns the response string
+try_r() {
```
g2flyer: if you return the result without filtering, this could potentially also be a useful function for common_utils.sh?

integration/auction_test.sh
```diff
@@ -21,37 +24,103 @@ CC_VERS=0
...
+    echo "$@" 
+    export RESPONSE=""
```
g2flyer: do you need RESPONSE in sub-shells?  Otherwise i wouldn't export it, just set ..

integration/auction_test.sh
```diff
@@ -21,37 +24,103 @@ CC_VERS=0
...
+    "$@" 2>&1 | tee -a /tmp/response.txt || die "test failed: $*"
```
g2flyer: hmm, this fails only if tee has an error, not if "$@" fails!

In bash you can access the exit codes via ${PIPESTATUS[0]}, ${PIPESTATUS[1]}, ${PIPESTATUS[2]}, ...

In fact, do you even need a pipe and a temp-file here? why not directly do `result=$("$@" 2>&1) and then test exit-code via $?

integration/auction_test.sh
annumberhocker: Yes, will move try_r()  and check_result() to common_utils.sh

integration/auction_test.sh
annumberhocker: I tried it without exporting it and it wasn't visible outside the function

g2flyer: hmm, unless you do it somewhere in a sub-shell, you do not have to export variables.

E.g.,
````
FOO=foo
foo() {
   FOO=bar
}
foo
echo $FOO
```
returns `bar`, not `foo` in bash

integration/auction_test.sh
annumberhocker: I figured there was a better way to do this, just couldn't get the syntax correct.  This seems to work, however
```
try_r() {
    echo "$@" 
    result=$("$@" 2>&1) || die "test failed: $*"
    echo $result
    export RESPONSE=$(echo $result | awk -F "\"" '{print $5}' | awk -F "\\" '{print $1}' | base64 -d)
    say $RESPONSE
}
``` It goes to the die if the invoke fails.

annumberhocker:
Turns out I don't see any changes that need to be made to Echo or Deployment tests except integrating the changes from PR 144. Please comment if there are other changes that should be made.

g2flyer:
> Turns out I don't see any changes that need to be made to Echo or Deployment tests except integrating the changes from PR 144. Please comment if there are other changes that should be made.

I would have assumed you also want to check the response for the other tests? E.g., that echo returns the expected value?  But then maybe i misunderstood your question above?

annumberhocker:
I was looking at the fact that there aren't any error conditions to test on.  I will confirm that it returns what it is called with.

mbrandenburger:
> @mbrandenburger Don't merge yet. I am going to restructure the try_r and check_result functions to move the string processing into check_result(). We can probably get rid of try_r () function and just change the try() function.

Ouch! Almost merged as it has many approvals already! Please change to WIP if still not ready to merge.

annumberhocker:
@mbrandenburger It is ready to merge now.  If everyone has reviewed.  I didn't change the description.  Not sure who took off the WIP.

PR #147, opened 2019-08-26
Fix broken tlcc config tx parser

mbrandenburger:
Fabric uses different types of config updates. TLCC currently
supports basic FABRIC MSP updates; other updates, such as
endpoints are ignored.


mbrandenburger:
This PR addresses #139

Issue #87, opened 2019-06-14
bad return values in sgxcc


bvavala:

mbrandenburger:
Can you please clarify?

bvavala:
at least one function return an uninitialized variable.
it's worth doing a review of that code.

bvavala:
this was solved in #158

PR #157, opened 2019-09-13
improved error handling in sgxcclib

bvavala:
This addresses issue #87 . Additionally
* improved (but not complete) handling of sig_rl
* removed ecall for get_target, use the sgx function instead
* improved log functions in sgxcclib (next: move these in a separate package/source file)

Note: linter/indentation checks might fail. This is addressed in separate PR


mbrandenburger:
This PR is redundant with #158. Let's

Issue #42, opened 2019-04-22
auction demo example with no submitting transactions fails


piachristel:
If one creates an auction, closes it and evaluates it (so without having submitted any bits), an error is thrown and the peer stops.
[error.txt](https://github.com/hyperledger-labs/fabric-secure-chaincode/files/3105833/error.txt)

mbrandenburger:
I cannot reproduce this error with the current code base. I suggest we close it and if needed reopen again.

annumberhocker:
This is still a problem.  Scenario is to create and close an auction with no bids, the create a a new auction.  The peer crashes
```
 >>>> Evaluate auction. Response should be NO_BIDS 
/home/bcuser/work/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/../fabric/bin//peer.sh chaincode invoke -o localhost:7050 -C mychannel -n auction_test -c {"Args":["[\"eval\",\"MyAuction3\"]", ""]}
2019-08-23 17:12:11.147 UTC [chaincodeCmd] chaincodeInvokeOrQuery -> INFO 001 Chaincode invoke successful. result: status:200 payload:"{\"ResponseData\":\"Tk9fQklEUw==\",\"Signature\":\"MEUCIAGXJSt+VUirxZ+6Ex+qFGDN4vY+hla3qjV+GmwcSJjYAiEAjcgOU4S5UlHbDZsLFTiTHmHZchNFRIR0hv5LCY0PDtw=\",\"PublicKey\":\"MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXRDkJgANkhO2fiPA488fELfP00gqyTNVJb3ZX50pHQTFdR2pj1HNLxz8U+gF5/UmkFfpxavi4uPBjQKni1W/mA==\"}"
auction_test.sh: NO_BIDS
 PASSED 
 >>>> Create a new auction. Response should be OK 
/home/bcuser/work/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/../fabric/bin//peer.sh chaincode invoke -o localhost:7050 -C mychannel -n auction_test -c {"Args": ["[\"create\",\"MyAuction4\"]", ""]} --waitForEvent
[...]

PR #159, opened 2019-09-17
Fixing Issue #42

mbrandenburger:
This commit fixes the problem of ecc-vscc crashes when processing
a transaction containing an empty range queries result. This may
happen, for instance, when an auction evaluation is invoked but
no bids are submitted. In detail, a range query is performed but
does not contain any raw reads.


mbrandenburger:
This issue fixes #42

mbrandenburger:
> Does fix the crash on both simulator and hw mode. Does not address the somewhat related security/functionality issue identified in issue #42: doesn't have to be fixed here but then we should create an issue separate of #42 for that? Leave it up to you which way to go, so approve here.

ok, will create a new issue for that and this here can be merged

PR #160, opened 2019-09-17
Fixing issue 152

mbrandenburger:
This commit fixes the issue that some read-write set validations at tlcc
failed. In particular, this was the case when multiple writes are in one
transaction.


mbrandenburger:
Created issue #162 for that

Issue #77, opened 2019-05-17
Make generate_enclave.sh more robust


mbrandenburger:
Currently, `ecc_enclave` and `tlcc_enclave` use a script to generate `mrenclave.go`. However, this script is a) duplicated and b) not robust.

As suggested by @g2flyer ...

... bigger problem seems that we do no catch any errors and so it was happily running so far (just producing only an empty mrenclave). Easiest fix for that seems to be o just add a 'set -e' in the beginning?

Also, given that the generate_mrenclave.sh is the same in ecc_enclave and tlcc_enclave, maybe we should move that to a common bin directory?

Thinking of it, the method here also seems rather brittle, the way we do it in pdo is more robust:

add to sgx_sign sign the '-dumpfile "${SIGNED_ENCLAVE_METADATA}"' option and then
extract it with a more understandable like the ' VAR_MRENCLAVE=$$(perl -0777 -ne 'if (/metadata->enclave_css.body.enclave_hash.m:([a-fx0-9 \n]+)/) { $$eh = $$1; $$eh=~s/0x| |\n//g; $$eh=~tr/a-z/A-Z/; print "$${eh}\n"; }' $(CONTRACT_ENCLAVE_MRENCLAVE_META_FILE))' in target '$(CONTRACT_ENCLAVE_MRENCLAVE_C_FILE)' in pservice/Makefile

PR #168, opened 2019-09-19
New Docker and Travis Support

mbrandenburger:
This PR adds new FPC docker runtime and dev images with SGX SDK 2.6 support. Moreover, it adds support for Travis-CI builds.


common/enclave/generate_mrenclave.sh
```diff
@@ -19,7 +19,6 @@ go_name=mrenclave.go
 cd $build_dir
 sgx_sign gendata -enclave enclave.so -config $enclave_dir/enclave.config.xml -out $tmp_name
 dd if=$tmp_name bs=1 skip=188 of=$hex_out_name count=32
-hexdump -C $hex_out_name
```
mbrandenburger: Removed as it only outputs the hexdata here but nothing more. Also hexdump was not available in new dockerbase image and decided not to add it.

Makefile
```diff
@@ -30,7 +30,7 @@ gotools:
 	# install goimports if not present
 	$(GO) get golang.org/x/tools/cmd/goimports
 
-godeps:
+godeps: gotools
```
mbrandenburger: goimports is needed to build `utils` but only added later.

common/enclave/generate_mrenclave.sh
g2flyer: we might want at some point remove it completely and just rely on the -dumpfile option of sgx_sign (see issue #77)

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+
+go_import_path: github.com/hyperledger-labs/fabric-private-chaincode
+
+env:
+  global:
+    # SGX config
+    - SGX_MODE=SIM
+    - SGX_SDK=/opt/intel/sgxsdk
+    # SGX SSL
+    - OPENSSL=1.1.0j
```
g2flyer: We might upgrade to 1.1.0l which is the latest version?  (btw: apropos version, bruno noticed that you forgot to update README.md from 1.4.2 to 1.4.3, so if you change versions, you might fix also README.md. While version changes could deserve a separate pr, you already changed the sdk version, so we might also update openssl version and fix other version-related omissions ...)

Makefile
```diff
@@ -6,7 +6,7 @@
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
+SUB_DIRS = docker utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
```
g2flyer: I must admit i liked having the docker stuff in utils (given that we already have it and kept docker stuff previously there) rather than putting it in the top-level ....

docker/base/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    libssl-dev \
+    libcurl4-openssl-dev \
+    libtool \
+    cmake \
+    vim \
+ && apt-get -y -q upgrade \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+ARG GO_VERSION=go1.13
```
g2flyer: For clarity, it would be nicer next to the GO_TAR definition.  Also, i guess right now we need go really only for dev? So why not just put it there?

On the other hand, right now we have sgxssl only in dev, once we would add deterministic build we would need it also in ``ccenv-build'' and so might need it in two different docker images?

docker/dev/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM fabric-private-chaincode/base:latest
+
+ARG OPENSSL=1.1.0j
```
g2flyer: See comments in travis.yaml regarding version ...

docker/Makefile
```diff
@@ -0,0 +1,24 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+TOP = ..
+include $(TOP)/build.mk
+
+FPC_DOCKER_NAMESPACE := fabric-private-chaincode
```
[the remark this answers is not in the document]
mbrandenburger: agree!

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+
+addons:
+  apt:
+    packages:
+      - protobuf-compiler
+      - python-protobuf
+
+branches:
+  only:
+    - travis
```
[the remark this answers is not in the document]
mbrandenburger: this restriction can be removed.

.travis.yml
mbrandenburger: Regarding fabric version, yes it seems that I didn't catch all 1.4.2 when updating README.md. Can one use variables in markdown?

Makefile
mbrandenburger: sure, we can keep it in utils

docker/base/Dockerfile
mbrandenburger: ok, let's move go to dev image for the moment.

.travis.yml
mbrandenburger: Upgraded openssl and added this to commit msg

mbrandenburger: with openssl 1.1.0j CI-build goes through

PR #173, opened 2019-10-02
generate enclave signer's key on the fly, and remove hardcoded ones

bvavala:


ecc_enclave/enclave/CMakeLists-common-app-enclave.txt
```diff
@@ -52,15 +52,16 @@ target_link_libraries(enclave "${SGX_COMMON_CFLAGS} -Wl,--no-undefined -nostdlib
 
 add_custom_command(TARGET enclave
     POST_BUILD
+    COMMENT "Create enclave signer's key"
+    COMMAND openssl genrsa -3 -out ${CMAKE_BINARY_DIR}/enclave_signer_private.pem 3072
```
mbrandenburger: why do we create a new signer key for every build?

g2flyer: > why do we create a new signer key for every build?

You forgot we are Intel, we want users to use as much cycles as possible  ... ;-)

bvavala:
* There is no need to ship those keys, particularly private ones, with the repo - so, yes, it is a good idea to remove them. 
* We have no requirement that a user/signer has to have such key - so it's easier not to impose users to make their own.
  * Similarly, it is not the case to keep a key lying around when it's not necessary - creating it in the build folder forces an appropriate cleanup when artifacts are removed.
* We have no requirement that such key has to be the same for all enclaves created locally.
* Also, we have no (strong) performance requirements for enclave build times.

mbrandenburger:
ok, please rebase and I will merge.

PR #188, opened 2019-11-14
Fpc network

### LINKED ISSUE #120: Schedule & present Executive Readout 1

### LINKED ISSUE #185: Dockerize the peer
For the demo setup it would be handy to start a peer as docker container. In particular, 
this may simplify to setup a multi-peer network and start demo front-end, fabric-client, etc as docker.

Fabric already allows to create this container and we just need to pass the configuration + plugins to the container;

More details TBD.

swetharepakula:
This pull request adds scripts and configuration to be able to stand up a FPC Network using docker-compose. The PR also includes a way to build an FPC Peer image that can be used in the network.

this commit resolves #185

### DIFF
--- a/utils/docker-compose/README.md
+++ b/utils/docker-compose/README.md
@@ -1,0 +1,230 @@
+# FPC Docker Compose Network
+This docker-compose example has been adapted from a [Fabric 101 Workshop](https://github.com/swetharepakula/Fabric101Workshop) which was adapted from the basic
+network and fabcar example in the [Fabric Samples](https://github.com/hyperledger/fabric-samples).
+This example does not use TLS, which means the Fabric Go SDK cannot be
+used to interact with the network. Currently, there are two orgs, one peer,
+one orderer, and one fabric-ca in the network.
+
+## Configuration
+- [core-fpc.yaml](network-config/core-fpc.yaml) : Peer configuration that has
+the SGX plugins and locations relative to location within docker image
+- [core.yaml](network-config/core.yaml) : Regular Peer configuration
+without FPC. Used if `$USE_FPC` is set `false`.
+- [orderer.yaml](network-config/orderer.yaml) : Orderer configuration
+- [crypto-config.yaml](network-config/crypto-config.yaml) : File used with cryptogen to generate
+certs for specified number of orgs, peers, users, and orderer. The CA credentials
+can be used to start instances of fabric-ca
+- [configtx.yaml](network-config/configtx.yaml)  : File used with configtxgen to generate the
+genesis block which is used as the basis of the specified channel
+- [docker-compose.yml](network-config/docker-compose.yml) : Configuration of the
+fabric network to be used with `docker-compose`. This file depends on two
+environment variables to properly bring up a network. `$FPC_CFG` can be set to
+`-fpc` or shall be empty. If set to `-fpc` the `core-fpc.yaml` & FPC peer image
+is used. Otherwise it will use `core.yaml` and the regular peer image.
+`$PEER_CMD` must also be set to the location of binary or script that will start
+ the peer.  **Docker version 17.06.2-ce or higher is needed**
+
+## Steps
+1. Build the peer image in `utils/docker/peer` directory which is defined by the
+peer [Dockerfile](../docker/peer/Dockerfile). This step
+assumes you have already built the [fabric-private-chaincode base image](../docker/base/Dockerfile).
+Take a look at building the docker dev environment in the main [README](../../README.md#docker).
+After you have created the base image, run the following to create a modified
+peer image and the plugins necessary to start the peer.  `$FPC_PATH` is the
+location fabric-private-chaincode repository on your host machine.
+```
+cd $FPC_PATH/utils/docker/peer
+docker build -t hyperledger/fabric-peer-fpc .
+```
+By default the image will clone the master branch on
+https://github.com/hyperledger-labs/fabric-private-chaincode. If you want to use
+a different fork of the repo or a different branch you provide
+`FPC_REPO_URL` and `FPC_REPO_BRANCH` as build args.
+```
+cd $FPC_PATH/utils/docker/peer
+docker build -t hyperledger/fabric-peer-fpc --build-arg FPC_REPO_URL=<repo-url> --build-arg FPC_REPO_BRANCH=<repo-branch> .
+```
+If you want to build the peer image using your local copy of your repo you can
+use the same build args, but specify `file:///tmp/build-src/.git` as the
+`FPC_REPO_URL`. You will also need to create the image at the root of this repo
+so that the local repo will be in the build context for the docker daemon.
+```
+cd $FPC_PATH
+docker build -t hyperledger/fabric-peer-fpc -f utils/docker/peer/Dockerfile --build-arg FPC_REPO_URL=file:///tmp/build-src/.git --build-arg FPC_REPO_BRANCH=$(git rev-parse --abbrev-ref HEAD) .
+```
+2. Download the necessary fabric binaries. Run the
+[bootstrap script](scripts/bootstrap.sh) which will download the Fabric 1.4.3
+into a local bin directory. The bootstrap script will download all the binaries
+to the location from where the scripts are run from. The rest of the tutorial
+expects the binaries to be in. If you already have the binaries downloaded in
+your path, this step can be skipped.**Fabric 1.4.3 versions of configtxgen and
+cryptogen are required to use the configurations above.**
+```
+cd $FPC_PATH/utils/docker-compose
+scripts/bootstrap.sh
+```
+
+3. Generate the cryptographic material needed for the network by running the
+[generate](scripts/generate.sh) script. Cryptogen will be used to generate all the
+credentials needed based on the configuration filesabove and place them in the
+`network-config/crypto-config` directory.  Configtxgen will be used to create
+the genesis block which is used to start up the orderer as well as the peer
+create channel configuration transaction. These will be placed in the
+`network-config/config` directory. The `crypto-config` & `config` directory will
+be mounted into every container of the FPC network as specified in the
+docker-compose file. **This script is not
+idempotent and will delete the contents of `crypto-config` & `config` when run
+to ensure a clean start.**
+```
+scripts/generate.sh
+```
+
+4. Start the network. Run the [start](scripts/start.sh) script. This will use
+docker-compose to start the network as well as starting the channel `mychannel`.
+By default, this script will use FPC peers. If non FPC peers are desired, set
+`$USE_FPC` to `false`.
+```
+scripts/start.sh
+```
+
+## Deploying your FPC Chaincode
+The [examples](../../examples) directory has been [mounted](base/base.yaml) into
+ the peer container for convenience, under
+ `/project/src/github.com/hyperledger-labs/fabric-private-chaincode/examples`.
+ **NOTE** If you are running a normal fabric network, the rest of the tutorial
+ will not work.
+
+1. Follow the [steps](../../examples/README.md) in the tutorial to build your
+chaincode outside of the peer container. Do not continue to the testing step.
+Though this tutorial references the hello world example, users can also deploy
+other FPC examples using similar steps.
+
+The rest of these steps should be done within the peer container.
+
+2. Exec into the peer container.
+```
+docker exec -it peer0.org1.example.com bash
+```
+
+3. Set environment variable to use the admin credentials and set the
+orderer address.
+```
+export CORE_PEER_MSPCONFIGPATH=/etc/hyperledger/msp/users/Admin@org1.example.com/msp
+export ORDERER_ADDR=orderer.example.com:7050
+```
+
+4. Install your chaincode. `PEER_CMD` is defined in the container already.
+```
+${PEER_CMD} chaincode install -l fpc-c -n helloworld_test -v 0 -p examples/helloworld/_build/lib
+```
+
+5. Instantiate your chaincode
+```
+${PEER_CMD} chaincode instantiate -o orderer.example.com:7050 -C mychannel -n helloworld_test -v 0 -c '{"Args":["init"]}' -V ecc-vscc
+```
+
+## Interact with the FPC Chaincode
+1. Store asset1 with a value of a 100
+```
+${PEER_CMD} chaincode invoke -C mychannel -n helloworld_test -c '{"Args":["storeAsset","asset1","100"]}'
+```
+
+2. Retrieve the current value of asset1.
+```
+${PEER_CMD} chaincode query -C mychannel -n helloworld_test -c '{"Args":["retrieveAsset","asset1"]}'
+```
+The response should look like the following:
+```
+{
+    "ResponseData":"YXNzZXQxOjEwMA==",
+    "Signature":<signature>,
+    "PublicKey": <public-key>
+}
+```
+
+3. Verify the encrypted response data shows that asset1 is equal to a hundred.
+```
+> echo "YXNzZXQxOjEwMA==" | base64 -d
+asset1:100
+```
+
+## Create a User with Fabric-CA
+5. Enter into the [`node-sdk`](node-sdk) directory, to use the node sdk scripts
+to create new users.
+```
+cd node-sdk
+```
+
+4. Ensure you have all the node modules
+```
+npm install
+```
+
+6. Enroll as the admin download the admin credentials
+```
+node enrollAdmin.js
+```
+After running this, the directory `wallet/admin` should have been created and
+have public and private key pair. **NOTE** These credentials are not an admin in
+the network, but just the admin for Fabric-CA and have the ability to register
+more users.
+
+7. Register another user and download the credentials.
+```
+node registerUser.js <username>
+```
+After running this with your desired username, the directory `wallet/<username>`
+should have been created and have the public and private key pair.
+
+## Interact with the chaincode using the Node SDK
+**NOTE: You must run peer invoke for this chaincode once using the peer cli
+commands in the peer container before you can use these node sdk scripts**
+
+1. Ensure you have all the node modules
+```
+npm install
+```
+
+2. Query the asset you stored previously
+```
+node query.js <username> mychannel helloworld_test retrieveAsset asset1
+```
+The response should look similar to what you saw above when you queried using
+the peer cli.
+```
+Transaction has been submitted, result is:
+{
+      "ResponseData":"YXNzZXQxOjEwMA==",
+      "Signature":<signature>,
+      "PublicKey":<public-key>
+}
+```
+In general the query script works as:
+```
+node query.js <identity-to-use> <channel-name> <chaincode-id> <args>...
+```
+3. To invoke a transaction:
+```
+node invoke.js <username> mychannel helloworld_test storeAsset asset2 200
+```
+The response should look like the following:
+```
+Transaction has been submitted, result is:
+{
+      "ResponseData":"T0s=",
+      "Signature":<signature>,
+      "PublicKey":<public-key>
+}
+```
+In general the invoke script works as:
+```
+node invoke.js <identity-to-use> <channel-name> <chaincode-id> <args>...
+```
+
+## Teardown the network
+
+1. Run the [teardown script](./scripts/teardown.sh) to clean up your environment. Run
+this in the root of this repo. **NOTE** This will try to remove all your
+containers and prune all excess volumes.
+```
+scripts/teardown.sh
--- a/utils/docker-compose/network-config/docker-compose.yml
+++ b/utils/docker-compose/network-config/docker-compose.yml
@@ -1,0 +1,83 @@
+#
+# Copyright IBM Corp All Rights Reserved
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Environment Variables that need to be set to use this file
+#
+#   COMPOSE_PROJECT_NAME - This is used to ensure all chaincode containers are running on the same network
+#                          as the peers, orderer, and ca. The only requirement is that this is set to
+#                          some string and cannot be left blank. To match with the scripts use `fabric-fpc`.
+#   FABRIC_VERSION       - This is the version of fabric being used that should match with an image tag
+#                          for hyperledger/fabric-orderer. Recommended value is `1.4.3`.
+#   FPC_CONFIG           - Set this to `-fpc` to bring up a network using FPC Peers. Leaving it unset
+#                          will cause the network to use the hyperledger/fabric-peer:latest images
+#   PEER_CMD             - This variable indicates which peer binary should be used to start the peer.
+#                          If regular peers are desired set this value to `peer`, if FPC peers are
+#                          desired set this value to
+#                          `/project/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/bin/peer.sh`
+#
+#
+#
+version: '2'
+
+networks:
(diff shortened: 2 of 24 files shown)


utils/docker-compose/README.md
```diff
@@ -0,0 +1,225 @@
...
+cd $FPC_PATH/utils/docker/peer
+docker build -t hyperledger/fabric-peer-fpc --build-arg FPC_REPO_URL=<repo-url> --build-arg FPC_REPO_BRANCH=<repo-branch> .
+```
+If you want to build the peer image using your local copy of your repo you can
+use the same build args, but specify `file:///tmp/build-src/.git` as the
+`FPC_REPO_URL`. You will also need to create the image at the root of this repo
+so that the local repo will be in the build context for the docker daemon.
+```
+cd $FPC_PATH
+docker build -t hyperledger/fabric-peer-fpc -f utils/docker/peer/Dockerfile --build-arg FPC_REPO_URL=file:///<your-gopath>/src/github.com/hyperledger-labs/fabric-private-chaincode.git .
```
g2flyer: this should be "....../fabric-private-chaincode.git/.git ."?  And given that you already assume $FPC_PATH, maybe even simpler write it as "...-build-arg FPC_REPO_URL=file:///${FPC_PATH}/.git"  (or to be more robust "...-build-arg FPC_REPO_URL=file:///$(readlink -f {FPC_PATH})/.git"?

g2flyer: This should be ``... -build-arg FPC_REPO_URL=file:///tmp/build-src/.git'' ..

utils/docker/peer/Dockerfile
```diff
@@ -0,0 +1,100 @@
+# Copyright IBM Corp. All Rights Reserved.
```
g2flyer: This is more or less a true super-set of dev/Dockerfile
-> 
Couldn't you just build on top of that image?  E.g., "backport" all changes you have done up except make plugin and last three ENV/CMD statement to the dev image and then build the peer image from the dev image, just with the plugin build and ENV/CMDs in the peer image as commands?

utils/docker-compose/scripts/generate.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
g2flyer: This scripts uses bash'ism => s:/bin/sh:/bin/bash:

utils/docker-compose/scripts/generate.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+export SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+export PATH=${SCRIPT_DIR}/../bin:${PWD}:$PATH
+export FABRIC_CFG_PATH=${SCRIPT_DIR}/../network-config
+CHANNEL_NAME=mychannel
+
+# remove previous crypto material and config transactions
+rm -fr ${FABRIC_CFG_PATH}/config/*
```
g2flyer: you also should add a ``mkdir -p ${FABRIC_FG_PATH/config}'', i had to manually create this config directory or the script failed

utils/docker/peer/Dockerfile
swetharepakula: I have made this change locally. The only concern I have before I update the PR is that this Peer image always copies in the entire docker context, which essentially means they every time the dev image is built it will not be able to use the cache for FPC repo images layers. I am thinking that is all right because chances are if you are rebuilding there are changes in the FPC repo. What do you think?

utils/docker-compose/scripts/generate.sh
swetharepakula: fixed

utils/docker-compose/scripts/generate.sh
swetharepakula: fixed

utils/docker-compose/README.md
swetharepakula: fixed

utils/docker/peer/Dockerfile
g2flyer: As long as you run in the "default" mode from within the dev/peer directory, it shouldn't really be a problem as it would only be an issue of the Dockerfile has changed? Or is docker not smart enough to realize that? There are also interesting workaround to make copy conditional (e.g., ONBUILD variant from https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile) but they are still rather kludgy and not worth adopting.

I guess there is the other potential issue that you might want peer and dev image with different repos (e.g., peer from local repo whereas you might have a peer with the remote repo) but that could be solved simply by naming conventions of the user.  

So maybe you are right and having duplicates is better. My main concern was inconsistent updates. But hopefully we don't have to update much.  What might help though is adding some comments in the Dockerfiles themselves pointing out the duplication and the potential need of replicating changes also to the "sibling"?

utils/docker-compose/network-config/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
+#
```
g2flyer: Might be worth mentioning in the header that it depends on the FPC_CONFIG, PEER_CMD and COMPOSE_PROJECT_NAME env variables?  
Along these lines, it would also be nice if the defaults of these variables from start.sh are also set as defaults in this yaml file, so i can use this yaml file directly to  run docker-compose -f <this-file> with commands such as ps, log, etc ..  
(Not really a big deal, i figured it out quickly as docker-compose gave meaningful warnings and then figuring out necessary values by greping -r for these VARs, but still would be handy if i wouldn't have to define these vars myself :-)

utils/docker/peer/Dockerfile
swetharepakula: Unfortunately, docker doesn't know it has the same stuff until it builds the layer again. I did look into the conditional building but didn't like the approaches there.

One approach is having one more intermediary Dockerfile which takes all the common pieces and then dev will only have the FPC clone piece and the peer will have its varied FPC clone. Would that be more desirable? We could also make some changes to the images so we have a ccenv file, a base that builds from ccenv, and a dev that builds from base, and a peer that builds from base.

utils/docker-compose/README.md
```diff
@@ -0,0 +1,228 @@
...
+docker-compose to start the network as well as starting the channel `mychannel`.
+By default, this script will use FPC peers. If non FPC peers are desired, set
+`$USE_FPC` to `false`.
+```
+.scripts/start.sh
+```
+
+## Deploying your FPC Chaincode
+The examples directory has been mounted into the peer container for convenience,
+under `/opt/examples`. All of these steps should be done within the peer container.
```
g2flyer: there is no /opt/examples and i couldn't find volumes which would match this in the docker-compose file?

utils/docker/peer/Dockerfile
g2flyer: I guess let's just leave it for now and revisit if ever we run into issues with keeping the similar files in sync. I thought there might be an easy clean way to reuse but evidently this is not the case. So doesn't seem worth spending more time on this now.

utils/docker-compose/README.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+5. Install your chaincode
+```
+${PEER_CMD} chaincode install -l fpc-c -n helloworld_test -v 0 -p examples/helloworld/_build/lib
+```
+
+6. Instantiate your chaincode
+```
+${PEER_CMD} chaincode instantiate -o orderer.example.com:7050 -C mychannel -n helloworld_test -v 0 -c '{"Args":["init"]}' -V ecc-vscc
```
g2flyer: you might want to use ${ORDERER_ADDR} here?

utils/docker-compose/README.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+3. Verify the encrypted response data shows that asset1 is equal to a hundred.
+```
+> echo "YXNzZXQxOjEwMA==" | base64 -d
+asset1:100
+```
+
+## Create a User with Fabric-CA
+4. Ensure you have all the node modules
```
g2flyer: sequence seems wrong here, i have to do 5 before 4 for npm install to work ....

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+## Steps
+1. Download the necessary fabric binaries. Run the
+[bootstrap script](scripts/bootstrap.sh) which will download the Fabric 1.4.3
+into a local bin directory. If you already have the binaries downloaded in your
+path, this step can be skipped. **Fabric 1.4.3 versions of configtxgen and
+cryptogen are required to use the configurations above.**
+```
+./scripts/bootstrap.sh
+```
+1. Build the peer image in `utils/docker/peer` directory which is defined by the
```
mbrandenburger: Is this step a substep of 1. Download .... ?

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+peer image and the plugins necessary to start the peer.  `$FPC_PATH` is the
+location fabric-private-chaincode repository on your host machine.
+```
+cd $FPC_PATH/utils/docker/peer
+docker build -t hyperledger/fabric-peer-fpc .
+```
+By default the image will clone the master branch on
+https://github.com/hyperledger-labs/fabric-private-chaincode. If you want to use
+a different fork of the repo or a different branch you provide
+FPC_REPO_URL and FPC_REPO_BRANCH as build args.
```
mbrandenburger:
```suggestion
`FPC_REPO_URL` and `FPC_REPO_BRANCH` as build args.
```

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+the genesis block which is used to start up the orderer as well as the peer
+create channel configuration transaction. These will be placed in the
+`network-config/config` directory. The `crypto-config` & `config` directory will
+be mounted into every container of the FPC network as specified in the
+docker-compose file. **This script is not
+idempotent and will delete the contents of `crypto-config` & `config` when run
+to ensure a clean start.**
+```
+cd $FPC_PATH/utils/docker-compose
+scripts/generate.sh
```
mbrandenburger: when I run this step, I get the following error:
```
  docker-compose git:(pr/188) ✗ ./scripts/generate.sh
./scripts/generate.sh: line 18: cryptogen: command not found
Failed to generate crypto material...
```
I indeed did step 1 downloading the binaries and therefore did not expect this error here. Double checking that download step fetched cryptogen shows it is there but not recognised by the `generate.sh` script.

```
bin git:(pr/188) ✗ pwd
/Users/marcusbrandenburger/Developer/gocode/src/github.com/hyperledger-labs/fabric-private-chaincode/utils/docker-compose/scripts/bin
➜  bin git:(pr/188) ✗ ll
total 288408
-rwxr-xr-x  1 marcusbrandenburger  staff    16M Aug 26 23:34 configtxgen
-rwxr-xr-x  1 marcusbrandenburger  staff    17M Aug 26 23:34 configtxlator
-rwxr-xr-x  1 marcusbrandenburger  staff    11M Aug 26 23:34 cryptogen
-rwxr-xr-x  1 marcusbrandenburger  staff    16M Aug 26 23:34 discover
-rwxr-xr-x  1 marcusbrandenburger  staff    14M Aug 26 23:31 fabric-ca-client
-rwxr-xr-x  1 marcusbrandenburger  staff   9.4M Aug 26 23:34 idemixgen
-rwxr-xr-x  1 marcusbrandenburger  staff    24M Aug 26 23:34 orderer
-rwxr-xr-x  1 marcusbrandenburger  staff    29M Aug 26 23:34 peer
```

mbrandenburger: when I change `generate.sh` it works
```
-export PATH=${SCRIPT_DIR}/../bin:${PWD}:$PATH
+export PATH=${SCRIPT_DIR}/bin:${PWD}:$PATH
```

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+```
+3. Start the network. Run the [start](scripts/start.sh) script. This will use
+docker-compose to start the network as well as starting the channel `mychannel`.
+By default, this script will use FPC peers. If non FPC peers are desired, set
+`$USE_FPC` to `false`.
+```
+.scripts/start.sh
```
mbrandenburger: a little inconsistency here ... we have three variants how to start a script.
```
./scripts/bootstrap.sh
scripts/generate.sh
.scripts/start.sh
```

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+## Deploying your FPC Chaincode
+The examples directory has been mounted into the peer container for convenience,
```
mbrandenburger: a pointer to the fpc examples would be handy here

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+under `/opt/examples`.
```
mbrandenburger: mh I can't see the examples:
```
➜  docker-compose git:(pr/188) ✗ docker exec -it peer0.org1.example.com bash

root@2c999fa8118d:/opt/gopath/src/github.com/hyperledger/fabric# cd /opt/examples
bash: cd: /opt/examples: No such file or directory
```

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+2. Exec into the peer container.
+```
+docker exec -it peer0.org1.example.com bash
+```
+
+3. Set up the peer command environment variable for convenience. `$FPC_CMDS` is
+already defined in the container.
+```
+export PEER_CMD=$FPC_CMDS/peer.sh
```
mbrandenburger: why is this step needed? I guess if non-fpc is used, PEER_CMD needs to point to the normal peer binary and not the fpc wrapper.

As $FPC_CMDS` is already defined, can we also simply define PEER_CMD for convenience? :D

utils/docker-compose/network-config/docker-compose.yml
swetharepakula: I don't know how to provide default values for the docker-compose.yml itself if one wants to use `docker-compose -f` directly. For now I will put some comments at the top of the file outlining the environment variables needed.

I added `set +x` to the `start.sh` so we should see the export statements printed out now

utils/docker-compose/README.md
swetharepakula: Updated with the right location and added a link to the file that has the mount.

utils/docker-compose/README.md
swetharepakula: Thanks for catching that! I will remove the flag since it is already defined as an env var, we don't need to specify it will the flag.

utils/docker-compose/README.md
swetharepakula: fixed

utils/docker-compose/README.md
swetharepakula: Oh that is a typo, should have been `2`, just like the structure you outlined.

utils/docker-compose/README.md
swetharepakula: I haven't seen someone use that github feature yet. That's pretty cool!

swetharepakula: fixed

utils/docker-compose/README.md
swetharepakula: yes, and @sjanakir2015 also pointed out the last one even has a typo. 

I am going to stick with `scripts/<script-file>`

utils/docker-compose/README.md
swetharepakula: that's a good idea. added the link.

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+[bootstrap script](scripts/bootstrap.sh) which will download the Fabric 1.4.3
+into a local bin directory. If you already have the binaries downloaded in your
+path, this step can be skipped. **Fabric 1.4.3 versions of configtxgen and
+cryptogen are required to use the configurations above.**
+```
+./scripts/bootstrap.sh
+```
+1. Build the peer image in `utils/docker/peer` directory which is defined by the
+peer [Dockerfile](../docker/peer/Dockerfile). This step
+assumes you have already built the [fabric-private-chaincode base image](../docker/base/Dockerfile).
```
g2flyer: Given that we have makefiles to build the other images, would it make sense to add a make target also for the peer? If so, of course, the question would be which config to use for which i would argue the local checkoued repo version (file:///tmp/build-src/.git  variant) might be the natural one?

utils/docker-compose/README.md
g2flyer: interestingly, this did work for me (although i've invoked it as `bash ./scripts/generate.sh` as the script identified itself wrongly as bourne shell at that time). Looking at the script now it looks correct and what you suggest wrong.  Do you have somehow managed to get the bin directory inside the scripts dir?

utils/docker-compose/README.md
g2flyer: the main issue seems that the last one (which is referenced here) should not work :-)

utils/docker-compose/README.md
g2flyer: it's mounted under /projects (i actually thought i've added a comment on this but i guess i must have lost it)

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+`$USE_FPC` to `false`.
+```
+.scripts/start.sh
+```
+
+## Deploying your FPC Chaincode
+The examples directory has been mounted into the peer container for convenience,
+under `/opt/examples`.
+
+1. Follow the [steps](../../examples/README.md) in the tutorial to build your
```
g2flyer: The hello world example requires quite some manual steps, so to some extent if you are only interested in the "delta" of the docker-compose, this is a bit painful.
When i tested, i just used the echo cc which worked also fine.  Probably not worth effort to change it now, but maybe point that out in text?

utils/docker-compose/network-config/docker-compose.yml
g2flyer: while i had the "luxury version" of a single line 'echo export A="${A}" B="${B}" ...' in mind, set -x also does the job :-)

utils/docker-compose/node-sdk/invoke.js
```diff
@@ -0,0 +1,57 @@
+/*
```
g2flyer: Couldn't we make these scripts directly executable via a #!/usr/bin/node?
(The same comment of course applies to the other js scripts ...)

utils/docker-compose/node-sdk/invoke.js
```diff
@@ -0,0 +1,57 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const ccpPath = path.resolve(__dirname, 'connection.json');
+const ccpJSON = fs.readFileSync(ccpPath, 'utf8');
+const ccp = JSON.parse(ccpJSON);
+
+async function main() {
+    try {
+        var user = process.argv[2]
```
g2flyer: given that this is a very handy general purpose tool, it would be nice to do error-checking and/or have a -h/-? option with usage?

utils/docker-compose/node-sdk/invoke.js
```diff
@@ -0,0 +1,57 @@
...
+        var channel = process.argv[3]
+        var chaincode = "ecc_" + process.argv[4]
```
g2flyer: As we still have the USE_FPC option, above wouldn't really work if we are not using FPC?

utils/docker-compose/README.md
swetharepakula: I can define it within the image itself for convenience. Though I think you bring another good point, should I indicate that this README/tutorial will not work for non FPC peers, because after the network setup the rest of the commands aren't really doable, except maybe creating new users using the node-sdk scripts.

utils/docker-compose/README.md
swetharepakula: That is a vestige of my initial attempt of mounting the examples. Michael also pointed it out. Should be fixed in the next commit.

utils/docker-compose/README.md
swetharepakula: Oh, interesting, where did you run the bootstrap script from? The expected location is supposed to be at the same level as the `SCRIPT_DIR`. Let me try to fix the bootstrap script so it doesn't download relative to where the script is run but relative to the script location

utils/docker-compose/network-config/docker-compose.yml
swetharepakula: ahhh, that's not hard to add :)

utils/docker-compose/README.md
swetharepakula: Actually after going through the steps I am going do:
1. Build
2. Download
3. Crypto
4. Start Network

Because the first step requires the user to be outside the `docker-compose` dir and then the rest can be done inside the `docker-compose` directory.

utils/docker-compose/README.md
swetharepakula: Okay, so the script is rather large and would require a bit of refactoring to be able to download relative to the script location for this problem. I will try to address that in a followup pr. For now I put more info on the download step so users are aware they have to run the script in a specific location.

utils/docker-compose/README.md
```diff
@@ -0,0 +1,227 @@
...
+can be used to start instances of fabric-ca
+- [configtx.yaml](network-config/configtx.yaml)  : File used with configtxgen to generate the
+genesis block which is used as the basis of the specified channel
+- [docker-compose.yml](network-config/docker-compose.yml) : Configuration of the
+fabric network to be used with `docker-compose`. This file depends on two
+environment variables to properly bring up a network. `$FPC_CFG` can be set to
+`-fpc` or shall be empty. If set to `-fpc` the `core-fpc.yaml` & FPC peer image
+is used. Otherwise it will use `core.yaml` and the regular peer image.
+`$PEER_CMD` must also be set to the location of binary or script that will start
+ the peer.  **Docker version 17.06.2-ce or higher is needed**
```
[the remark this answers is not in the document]
swetharepakula: oh, missed that. I will update it to be consistent.

utils/docker-compose/README.md
swetharepakula: I do like this. I am envisioning `peer` and a `peer-local` target. For now I am going to add the `peer` target only and in a followup I will add the local one.

utils/docker-compose/node-sdk/invoke.js
swetharepakula: Yes, that is correct. For now I have made a note in the README that if users choose to use a non-fpc network most of the tutorial will not work. But I can make changes in a followup commit/pr.

utils/docker-compose/node-sdk/invoke.js
swetharepakula: I agree! I will address that in a followup commit

utils/docker-compose/node-sdk/invoke.js
swetharepakula: For now I think I am going to leave them as node scripts and not executables. Locally on my machine node is installed in a different location so my guess is that is not a universally accepted location of node? But I do like the idea of running them like an executable eventually.

swetharepakula:
I addressed most of the comments and have made corresponding edits to the commit. Some of the comments I will address in followup commits/prs.

utils/docker-compose/README.md
g2flyer: actually, it is very easy to fix (2 lines): will come in my upcoming PR :-)

PR #191, opened 2019-12-10
Add clock auction backend client and folder structure

sjanakir2015:
Create clock_auction folder structure
    - Add backend node SDK client which enables interaction with FPC Network

This clock_auction folder is _not_ complete by itself.  The objective is to use this along with other components which are yet to be integrated.


g2flyer:
As discussed on slack, i think we should consolidate everything under demo and not have demo and clock_auction.  To some extent i would have thought putting everything under demo/clock_auction might be the cleanest but moving just everything from clock_auction to demo is probably easier as marcus and bruno also seem to assume everything is straight in demo and it also means that none of the '..' of TOP_DIR in makefiles have to be fixed as it is lateral move of the same depth ..

sjanakir2015:
> As discussed on slack, i think we should consolidate everything under demo and not have demo and clock_auction. To some extent i would have thought putting everything under demo/clock_auction might be the cleanest but moving just everything from clock_auction to demo is probably easier as marcus and bruno also seem to assume everything is straight in demo and it also means that none of the '..' of TOP_DIR in makefiles have to be fixed as it is lateral move of the same depth ..
> 
> Correction: the stuff in demo was some leftover network configuration stuff from a prior version. moving to demo (or demo/clock_auction), though probably still seems a better place than a top-level clock-auction

Moved top level from clock_auction to demo, maintaining the same level.

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+              discovery:           { enabled: false, localhost: bLocalHost },
+              eventHandlerOptions: { strategy: null }
+      		  });
+    console.log ("Connected to gateway");
+    network =  await gateway.getNetwork (channelName);
+    console.log ("Connected to channel, " + channelName);
+
+    contract =  await network.getContract (chaincodeID);
+    console.log ("Connected to chaincode, ", chaincodeID);
+    console.log ('...........................');
```
mbrandenburger: `fabricClient.js` is very verbose. In general logging is a good thing but here I wish to reduce it or introduce the concept of log levels? Seems that most log statements were primarily used for developing this code but don't bring necessarily any value when running the demo?!

Please revisit the logging!

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+    } catch (error) {
+       console.log(`Error in call to fabric_ca_client.newIdentityService: ${error}`);
+       console.log(error.stack);
+       throw (`Error in call to fabric_ca_client.newIdentityService: ${error}`);
+    }
+
+    var adminIdentity = await gateway.getCurrentIdentity();  //  which is org1admin
+
+    //  user must be a hf.registrar
+    try {
```
mbrandenburger: is the try really needed round so much code or is it only needed for the `idService.getAll` call?

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+        if (identities != undefined) {
+          for (var i =0; i < identities.length; i++) {
+              tmp = {};
+              tmp.id = identities[i].id;
+              tmp.approle = getRole (identities[i]);
+              result.push(tmp);
+          }
+        }
+        console.log ("<<<  Function getRegisteredUsers... ")
+        return result;
```
mbrandenburger: what about using array.map here and make this much more elegant?

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+         "ecert": true
+     });
+
+     var newUserDetails = {
+         enrollmentID: userid
+         , enrollmentSecret: pwd
+         , role: "user"
+         , affiliation: org1Affiliation
+         // , profile: 'tls'
+         //, attrs: certAttributes
```
mbrandenburger: do we need this later for the demo, otherwise `certAttributes` can be removed?

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+        const userExists = await wallet.exists(userName);
+        if (userExists) {
+           var result =  {
+                'status': {
+                  'rc': USER_EXISTS,
+                  'msg': 'An identity for the user ' + userName + ' already exists in the wallet' }
+              }
+
+        console.log('An identity for the user ' + userName + ' already exists in the wallet');
+        return result;
```
mbrandenburger: some code formatting here needed

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+        const identity = await X509WalletMixin.createIdentity(orgName, enrollment.certificate, enrollment.key.toBytes());
+        await wallet.import(userName, identity);
+
+        //  return success
+        var result =
+        {
+          "status": {  'rc': SUCCESS,
+          'msg': 'Successfully enrolled user ' + userName + 'and imported it into the wallet' }
+        };
+        console.log('Successfully enrolled user ' + userName + ' and imported it into the wallet');
```
mbrandenburger: why not just `console.log(result.msg)`?

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+        return result;
+
+    } catch (error) {
+      var result =
+      {
+        "status": {  'rc': USER_UNAUTHORIZED,
+        'msg': 'Failed to enroll user ' + userName + ': ${error}'  }
+      };
+
+      console.error( 'Failed to enroll user ' + userName + ': ${error}');
```
mbrandenburger: same comment here

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+        //  contract is global variable, used by invoke and query functions
+        contract =  await network.getContract (configdata["smart_contract_name"]);
+        console.log ("Connected to chaincode, ", configdata["smart_contract_name"]);
+        console.log ('...........................');
+        //  return success
+        var result =
+        {
+          "status": {  'rc': SUCCESS,
+          'msg': 'Successfully set user context to  user:  ' + userName }
+        };
```
mbrandenburger: this code would also benefit from a little helper function

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+    console.log (">>> In function, fabricClient:invoke... ");
+    console.log ("userName =***" + userName + "***");
+    console.log ("txName =  ", txName);
+    console.log ("args =  ", args);
+
+    try {
+    await setUserContext (userName,secret);
+    }  catch (error) {
+        throw (error);
+    }
```
mbrandenburger: is this the correct way to return the error? I am not an javascript expert but this just does look good/right?!

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+
+    //  convert buffer to string
+    var bufferData = {
+                           "type" : "Buffer",
+                           "data" : []
+                    };
+       bufferData["data"] = response;
+       strResponse = String.fromCharCode.apply (null, bufferData["data"]);
+
+    console.log ("Response (string): ", strResponse);
```
mbrandenburger: what happens here? can we use something like buf.to_String or buf.tojson? https://nodejs.org/api/buffer.html#buffer_buf_tojson

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+    return Promise.resolve(strResponse);
+    },(error) =>
+    {
+       //  handle error if transaction failed
+       console.log('Error thrown from tx promise',error);
+       return Promise.reject(error);
+    });
+}
+
+async function query (userName, txName, ...args) {
```
mbrandenburger: this function looks very similar to invoke except of `evaluateTransaction`. Please try to reduce code duplication

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+
+    return Promise.resolve(strResponse);
+    },(error) =>
+    {
+        //  handle error if transaction failed
+        console.log('Error thrown from tx promise',error);
+        return Promise.reject(error);
+    });
+
+    console.log ("<<< function, fabricClient.query: ");
```
mbrandenburger: unreachable

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+  console.log ("<<< function, fabricClient.beginNetworkConnect: ");
+
+}
+
+async function endNetworkConnect () {
+  console.log (">>> In function, fabricClient.endNetworkConnect: ");
+
+  console.log ("<<< function, fabricClient.endNetworkConnect ");
+
+}
```
mbrandenburger: Is there some cleanup needed?

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,477 @@
...
+//          this userName
+async function setUserContext (userName, pwd) {
+    console.log ('>>> Entry: function setUserContext ....');
+
+    try {
+        //  verify if user is registered
+        await enrollUser(userName, pwd);
+    } catch (error) {
+      throw (error);
+    }
```
mbrandenburger: when reading about async functions and await (https://javascript.info/async-await) seems that you can just call await encrollUser(...) and if this throws an error ... you can catch it later with setUserContext.catch( err => ....

demo/client/backend/src/server.js
```diff
@@ -0,0 +1,48 @@
...
+
+const fabricClient = require('./fabricClient');
+const ccRoute = require('./ccRoute');
+const apiRoute = require('./apiRoute');
+const clockauctionRoute = require('./clockauctionRoute');
+
+////////////////////////////  Global constants  /////////////////
+//  errorcodes
+const SUCCESS = 0;
+const FAILURE  = 1;
```
mbrandenburger: not used

demo/client/backend/src/server.js
```diff
@@ -0,0 +1,48 @@
...
+
+//  SJ: install express
+var express = require('express');
+var app = express();
+app.use(express.json());
+app.use(function (req, res, next) {
+    res.header("Access-Control-Allow-Origin", "*");
+    res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept");
+    res.header("Access-Control-Allow-Methods", "PUT, POST, GET, DELETE, OPTIONS");
```
mbrandenburger: we dont use PUT and DELETE, OPTIONS, right? Even though I know that the I sends indeed some OPTIONS, but I dont think this is needed. We could be more restrictive here

demo/client/backend/src/server.js
```diff
@@ -0,0 +1,48 @@
...
+}));
+
+//  different routes defined
+app.use('/api/cc', ccRoute);
+app.use('/api', apiRoute);
+app.use('/api/clock_auction', clockauctionRoute);
+
+async function main() {
+    console.log ("Listening on localhost:3000...");
+    var server=app.listen(3000,function() {});
```
mbrandenburger: would be good to read the port from config or args so we can configure it when creating the docker network.

demo/client/backend/src/server.js
```diff
@@ -0,0 +1,48 @@
...
+
+    await fabricClient.connectToNetwork ("mychannel", "mockcc");
```
mbrandenburger: channel and chaincode name should be read from config.json

demo/scripts/installCC.sh
```diff
@@ -0,0 +1,26 @@
...
+#  When (install + instantiate) commands are executed by hand, it works.  With scripting, the delay 
+#  due to manual action is absent.  Is that the reason?? 
+#  This version works.  TODO:  To come back and identify the reason for this error
+#  above comment is not valid anymore;  But this version works !  
+set -ev
+export WAIT_TIME=15
+export version=1.1
+
+docker image prune -f
+docker exec cli peer chaincode install -n mockcc -v ${version} --path /opt/chaincode/javascript  -l NODE
```
mbrandenburger: ideally, chaincode name is also read from config.json so we have a single place for configuration.

.gitignore
```diff
@@ -9,3 +9,5 @@ utils/docker-compose/network-config/crypto-config/
 utils/docker-compose/network-config/config/
 utils/docker-compose/node-sdk/node_modules
 utils/docker-compose/node-sdk/wallet
+clock_auction/client/backend/node_modules 
+clock_auction/client/backend/gateway/localhost/wallet
```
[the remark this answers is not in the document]
sjanakir2015: > I suggest creating your own `.gitignore` in `demo/client/backend` and add the node modules and wallet there. This makes it easier to maintain the .gitignore when changes in the repo happens, for instance, when renaming folder from `clock_auction` -> `demo`.

Will fix this

demo/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Build images once
+```
+cd <FPC-INSTALL-DIR>/demo/client/backend
+docker build -t auction_backend  .
+```
+
+## Run client
+```
+export NETWORK_NAME=fabric-fpc_basic
```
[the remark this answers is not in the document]
sjanakir2015: > I suggest that we also rename this network to `fpc-auction-demo` or similar

network name is fixed in the docker-compose file when FPC network is brought up.    I would assume this is to be done in a different PR.

demo/scripts/installCC.sh
sjanakir2015: These scripts are not intended to be used in a production setup.   It is presented to be used during dev time.  

This is work in progress.  To read from a json file, currently jq can be used.  However, since jq will be an additional dependency, it is not preferred.  Until a different tool is identified, it is hardcoded here.

sjanakir2015: > These scripts (installCC.sh and upgradeCC.sh) are not intended to be used in a production setup. It is presented in this PR to be used during dev time for integration purposes.
> 
> Also, this is work in progress. To read from a json file, currently jq can be used. However, since jq will be an additional dependency, it is not preferred. Until a different tool is identified, it is hardcoded here.

EDIT: Now that jq is added to the base image, will change this script to pick up chaincode name from config.json

demo/client/backend/src/fabricClient.js
```diff
@@ -0,0 +1,399 @@
...
+var contract;
+
+//  exported functions
+async function connectToNetwork() {
+    console.log (">>> In function, connectToNetwork");
+
+    await beginNetworkConnect();
+
+    try {
+      gateway = await new Gateway();
```
mbrandenburger: I am wondering why this code continues when `new Gateway()` throws an error. I assume that `gateway.connect` will also fail, thus no need to continue here?!

demo/client/backend/config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "connection_profile_filename":"../gateway/localhost/connection_profile.json",
+    "channel_name":"mychannel",
```
g2flyer: btw: my PR for docker-compose makes now channel name configurable if you define the CHANNEL_NAME env variables (of similar interest might be NODE_WALLETS for your wallet below; would make sure that you don't have left-over inconsistent wallets if you remove CA state with teardown --clean-slate  ...

demo/client/backend/src/server.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+app.get('/',(function(req,res){
+    res.send('Welcome to Spectrum Auction by Elbonia Communication Commission - Enabled by FPC');
+}));
+
+//  different routes defined
+app.use('/api/cc', ccRoute);
+app.use('/api', apiRoute);
+app.use('/api/clock_auction', clockauctionRoute);
```
g2flyer: this is different to what marcus has for his mock auction.  I like, though, your path better as it keeps auction agnostic stuff separate for auction-specific stuff which will make later generalization -- the client framework has only few auction-specific stuff and might be very convenient also for other applications .. -- easier ..

demo/scripts/installCC.sh
g2flyer: (additional reply for marcus) for FPC net you anyway will have to run them on the peer itself and with ${PEER_CMD} (the same of course also would apply to upgradeCC.sh)

utils/docker/dev/Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN apt-get update \
  && apt-get install -y \
     docker.io \
     clang-format \
+    jq \
```
g2flyer: you might want to add this dependency to the requirement section of README.md

demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+##  Register auction application users (bidders and auctioneers) with Certificate Authority
+```
+cd fabric-private-chaincode/demo/client/backend
+./registerUsers.sh
+```
+
+## Run client
+```
+export NETWORK_NAME=fabric-fpc_basic
```
g2flyer: after my PR this will be fabricfpc-fpc_basic as docker-compose in some cases squashes anything which is not alphanumeric

demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Some dependencies are currently hardcoded. This is work in progress.  As the scripts to stand up the demo application evolve, these will be modified. Example: NETWORK_NAME used in following instructions.
+
+Usage:
+```
+## Build image once
+cd fabric-private-chaincode/demo/client/backend
+docker build -t auction_backend  .
+```
+
```
g2flyer: you might add a short section that you assume a network setup via $FPC_PATH/utils/docker-compose with `USE_FPC=false`.

demo/client/backend/auction.json
```diff
@@ -0,0 +1,203 @@
...
+		"id": 2,
+		"displayName": "B-Net",
+		"principle": {
+			"mspid": "org2",
+			"dn": "bidder@org2"
+		}
+	}, {
+		"id": 3,
+		"displayName": "C-Mobil",
+		"principle": {
```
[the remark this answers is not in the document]
sjanakir2015: > is this data replaced based on the crypto material created for the network? or at least does it match?

1. The list of 3 "displayName"s  match with the bidders' usernames specified in config.json.   
2. MSPID is defined in the docker-compose file. 
 This is work in progress and is to be manually verified that these match.

demo/client/backend/auction.json
```diff
@@ -0,0 +1,203 @@
...
+		"bidderId": 1,
+		"number": 10000
+	}, {
+		"bidderId": 2,
+		"number": 10000
+	}, {
+		"bidderId": 3,
+		"number": 10000
+	}],
+	"owner": {}
```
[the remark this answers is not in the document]
sjanakir2015: > I don't think you need any owner as this is the template for createAuction (which doesn't have an owner, only getAuctionDetails does ...)

Will check with Marcus and fix accordingly since this is json content was  defined and used by UI component.

demo/README.md
sjanakir2015: > you might add a short section that you assume a network setup via $FPC_PATH/utils/docker-compose with `USE_FPC=false`.

why specifically 'USE_FPC=false' ?  For the client, it does not matter if it is FPC or Fabric network

sjanakir2015: > > you might add a short section that you assume a network setup via $FPC_PATH/utils/docker-compose with `USE_FPC=false`.
> 
> why specifically 'USE_FPC=false' ? For the client, it does not matter if it is FPC or Fabric network.  

Added a sentence which says "Before the following steps can be run, an FPC network should be setup using..."

demo/client/backend/auction.json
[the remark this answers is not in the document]
mbrandenburger: @sjanakir2015  "owner" can be removed here

demo/client/backend/config.json
```diff
@@ -0,0 +1,13 @@
...
+    "smart_contract_name":"mockcc",
+    "wallet": "../gateway/localhost/wallet",
+    "admin_user_name":"org1admin",
+
+    "users" : [
+    {"userName":  "A-Telecom",   "userRole" : "auction.bidder"} ,
```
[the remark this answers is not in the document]
sjanakir2015: > probably makes sense to add the MSPid and just set it right now to Org1 everywhere: That way it would mark it somewhat explicitly already now as a place we might have to eventually adapt when moving to to multi-org. That said, registeredUsers.sh as consumers will anyway have to change for multi-peer and maybe to move node-client from utils/docker-compose/node-sdk, so not sure whether it might better to hold off to the time we move to multi-org overall and we have to the "bigger picture" on what multi-org really means? So my feel is that for now might just leave it at most by adding mspid here but not (yet!) try to adapt the scripts correspondingly ...

I will set MSPID to "Org1MSP".  This is the value set in fabric-private-chaincode/utils/docker-compose/network-config/docker-compose.yml.

demo/client/backend/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENV PATH /usr/src/app/node_modules/.bin:$PATH
+WORKDIR /usr/src/app
+
+# RUN npm rebuild
+
+COPY gateway /usr/src/app/gateway
+COPY src /usr/src/app/src
+COPY config.json /usr/src/app/config.json
+
+CMD npm install; npm start
```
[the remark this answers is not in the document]
sjanakir2015: Good catch !  I was facing some errors (due to the dependencies) when I was creating this Dockerfile.  It is a remnant of a hacky attempt.
  
I have moved it to image build time.

demo/scripts/installCC.sh
sjanakir2015: > ideally, chaincode name is also read from config.json so we have a single place for configuration.

chaincode name is not only input needed for the scripts, installCC.sh and upgradeCC.sh.  (the version number is the one that changes frequently during dev time.   Hence, used as input to upgradeCC.sh)

Reading chaincode name from config.json will be an _incomplete_ work and reading all the inputs (chaincode name, chaincode folder, version, orderer address, channel name, args to chaincode) is not worth the effort now for 2 reasons. 
- it will be an overkill for a convenience script intended for repeated use in dev time. 
-  Integrated scripts which bring up end-to-end demo network will have to do this.  As Michael pointed out, these are to be done using {PEER_CMD} FPC peer wrapper.

So, I propose that I remove these scripts from this PR since these use hardcoded values which is not preferred.

mbrandenburger: yes, I think the scripts folder can be removed from this PR. We will add them later.

g2flyer: why remove? Still useful if you would want to use your go mock chaincode where you do not need the fpc network?

demo/README.md
g2flyer: your subsequent instruction would work only for USE_FPC=false as (a) the network-name is different for USE_FPC=true and (b) you have to invoke installs directly in peer using ${PEER_CMD} rather than doing it in cli docker ...

demo/client/backend/config.json
sjanakir2015: > btw: my PR for docker-compose makes now channel name configurable if you define the CHANNEL_NAME env variables (of similar interest might be NODE_WALLETS for your wallet below; would make sure that you don't have left-over inconsistent wallets if you remove CA state with teardown --clean-slate ...

currently, the code uses config.json for all configuration.    I prefer to go with either env variables or config.json.   
I propose to postpone this change for a future PR so that we have all the configuration parameters (and their usage) and we can make _one_ change to env variables, if needed.

g2flyer: this was mostly an FYI that the docker-compose network has now a configurable CHANNEL_NAME, so once you get there to configure it from config.json, you just would have to set the CHANNEL_NAME env-variable to whatever there is in config.json before calling the scripts from utils/docker-compose and they should use the correct channel name. Definitely did _not_ mean to do all of this in this PR.

demo/scripts/installCC.sh
sjanakir2015: Given that the mock chaincode is not part of this PR. having the scripts to use with mockcc does not help.  These scripts are available in the branch _which_ has mockcc (and the work is not lost). 

Moreover, the main purpose of this PR is to bring in backend client for the demo with FPC peer. The scripts can be part of another PR which _can_ address the dependencies.

demo/README.md
sjanakir2015: 1.  removing the scripts installCC,.sh and upgradeCC.sh from this PR;   reference to these scripts will be removed from README. 

2.  can we have the same network name for both cases of USE_FPC=true or false?
  
For this PR, I will change the note that (this is work-in-progress;) the NETWORK_NAME is dependent on USE_FPC as mentioned in README.  Will instruct the user to verify the network name and set it accordingly.

demo/README.md
sjanakir2015: Added note to README and a reference to docker-compose/README - to explain the dependency on the network setup - FPC or non-FPC.

demo/scripts/installCC.sh
sjanakir2015: Since the purpose of this PR is to merge the backend client code, will remove the scripts which work _only_ with non-FPC / fabric network.  
If the scripts are needed, I would add it another PR with all the hardcoded input and dependencies addressed.

demo/client/backend/src/apiRoute.js
```diff
@@ -0,0 +1,74 @@
...
+
+
+ApiRouter.route('/getRegisteredUsers').get (function (request, response) {
+    console.log('>>>  Entry:  apiRoute.getRegisteredUsers');
+    fabricClient.getRegisteredUsers().then((result) => {
+        // process response
+        console.log('Process getAllUsers response');
+        response.send(result);
+        },(error) => {
+            error.errorCode = 0;
```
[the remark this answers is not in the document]
sjanakir2015: Changed:  In case of error, getRegisteredUsers returns a status object.  

However, based on our slack conversation, in case of success, getRegisteredUsers() returns an json array of users.  This is the only function that is NOT returning status object.  I prefer to change it such that it returns a json status object in which ."users" will give the json list of users.

demo/client/backend/src/fabricClient.js
sjanakir2015: Added a console.log to error handling code.  That was missing here.

demo/client/backend/src/fabricClient.js
sjanakir2015: Again, this was dev code without _any_  scrutiny done, trying to bring in a basic dev code for end-to-end integration.  

Specifically, this try-catch block was intended (as a reminder) to catch errors from enrollUser so that they can be handled differently.    In later commit, the error is converted to a status object.

demo/README.md
sjanakir2015: BTW, when I was integrating FPC peer,  Bruno's chaincode and the backend client, I had modified installCC.sh to install / upgrade CC on FPC network - by running {PEER_CMD} instead of  fabric "peer" inside peer container - _from cli / script_.  I have shared the modified scripts / notes  with Swetha.

mbrandenburger:
@g2flyer @bvavala on more review please and we are ready to merge this!

sjanakir2015:
> I see the nodejs folder but... I don't see the minor changes to README and I see you've duplicated looots of files. Was that intended?

That was not intended.  Sorry.  Fixing it now.  Once done, will re-request review by you. 

BTW, the folder name is now "fabric-gateway" instead of "nodejs".

mbrandenburger:
Thanks for this PR!

PR #197, opened 2020-01-07
Mock with Creator 

g2flyer:
This PR adds a Creator field to the mock stub (i.e., a new fabric patch!) and then adds logic to create a creator object from the X-User field in the mock backend.   The format should match the identities we would see in "production" based on `demo/client/backend/fabric-gateway/registerUsers.sh`; the auction creation templates are also accordingly updated.

Note: to test you will need the additional fabric test. Assuming your environment already has the previous patches, then the following should be sufficient
```
cd $FPC_PATH
pushd ../../hyperledger/fabric
git am $FPC_PATH/fabric/0001-mock-also-creator-identity-like-signedproposal-or-Ch.patch
GO_TAGS=pluginsenabled make peer
popd
make clean
make
```

Also, license test will fail due to the *.json and patch-file not being able to have license headers ....

### DIFF
--- a/demo/chaincode/golang/auction.go
+++ b/demo/chaincode/golang/auction.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 
 	"github.com/hyperledger/fabric/core/chaincode/shim"
+	cid "github.com/hyperledger/fabric/core/chaincode/shim/ext/cid"
 	pb "github.com/hyperledger/fabric/protos/peer"
 )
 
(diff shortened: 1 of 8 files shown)


demo/chaincode/golang/auction.go
```diff
@@ -11,6 +11,7 @@ import (
 	"fmt"
 
 	"github.com/hyperledger/fabric/core/chaincode/shim"
+	cid "github.com/hyperledger/fabric/core/chaincode/shim/ext/cid"
```
mbrandenburger: I think the "cid" alias here is not needed here as it is the same as the package name. That is, below you call `cid.New(...` anyway.

PR #201, opened 2020-01-11
Mockcc integration

swetharepakula:
This has a few fixes that were needed for the components to all talk to each other. I have separated those fixes into separate commits from the end to end demo scripts. I can break this into two PRs if that is easier to review.

The scripts all work to install and use the mockcc. Once the auction cc is merged in I can make a followup commit to install and use the auctioncc instead.


demo/README.md
```diff
@@ -35,15 +100,13 @@ Note: This is work in progress.  Some environment variables are hardcoded in the
...
 
-```                 
-cd $FPC_PATH/demo/client/backend/fabric-gateway
-export NETWORK_NAME=fabricfpc-fpc_basic
+```
+export NETWORK_NAME=fabricfpc_basic
 export BACKEND_PORT=3000
-docker run --network $NETWORK_NAME -d  -v ${PWD}:/usr/src/app \
-		 -p $BACKEND_PORT:$BACKEND_PORT --name client_backend auction_client_backend
+docker run --network $NETWORK_NAME -d -p $BACKEND_PORT:$BACKEND_PORT --name auction-client auction_client_backend
```
g2flyer: it would seem more natural to start it with docker-compose, in which case the network name and backend-port issue should be moot?

demo/README.md
```diff
@@ -72,3 +135,46 @@ Use following functions to invoke a transaction or query.  Please note that any
...
+1. Build the Frontend UI docker image
+```
+cd $FPC_PATH/demo/client/frontend
+make build
+```
+
+2. Run the docker image
+```
+export NETWORK_NAME=fabricfpc_basic
+docker run --network $NETWORK_NAME -p 5000:5000 -d --name auction_frontend auction_frontend
```
g2flyer: See above comment for backend regarding use of docker-compose instead of docker ...

demo/scripts/startFPCAuctionNetwork.sh
```diff
@@ -0,0 +1,45 @@
...
+"${SCRIPT_DIR}/start.sh"
+
+# Install and Instantiate Auction Chaincode
+"${DEMO_SCRIPTS_DIR}/installCC.sh"
+
+# Register Users
+"${DEMO_ROOT}/client/backend/fabric-gateway/registerUsers.sh"
+
+# Build Auction Client Backend
+pushd  "${DEMO_ROOT}/client/backend/fabric-gateway"
```
g2flyer: Couldn't we capture that inside docker-compose file? If so, would be nicer/easier?

demo/scripts/teardown.sh
```diff
@@ -0,0 +1,24 @@
...
+DEMO_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+DEMO_DOCKER_COMPOSE=${DEMO_SCRIPT_DIR}/../docker-compose.yml
+
+# SCRIPT_DIR is the docker compose script dir that needs to be defined to source environment variables from the FPC Network
+SCRIPT_DIR=${DEMO_SCRIPT_DIR}/../../utils/docker-compose/scripts
+. ${SCRIPT_DIR}/lib/common.sh
+
+COMPOSE_IGNORE_ORPHANS=true docker-compose -f ${DEMO_DOCKER_COMPOSE} kill && COMPOSE_IGNORE_ORPHANS=true docker-compose -f ${DEMO_DOCKER_COMPOSE} down
+
+"${SCRIPT_DIR}/teardown.sh" --clean-slate
```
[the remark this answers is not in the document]
swetharepakula: The only reason I do clean-slate is because I run generate right below, and we need to get rid of the CA volumes if we recreate the certs. I could separate out the generate and then add the flag here too if you prefer.

demo/scripts/installCC.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+
+set -ev
+
+export WAIT_TIME=15
+export version=1.0
+export FABRIC_BIN_DIR=/project/src/github.com/hyperledger/fabric/.build/bin
+export PEER_CMD=/project/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/bin/peer.sh
+
+docker exec -e "CORE_PEER_LOCALMSPID=Org1MSP" -e "CORE_PEER_MSPCONFIGPATH=/etc/hyperledger/msp/users/Admin@org1.example.com/msp" peer0.org1.example.com ${PEER_CMD} chaincode install -n mockcc -v ${version} --path github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/golang/cmd -l golang
```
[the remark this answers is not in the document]
swetharepakula: I actually think I could remove some of the environment variables. This script I initially wrote before your fixes to the docker-compose, so let me test it out and try removing it completely.

demo/scripts/teardown.sh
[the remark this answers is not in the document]
g2flyer: i don't think this is true (wipe ca volumes when generating new (client) certs: i think we had the same discussion on the prior PR and whatever we discussed then i did test and CA wipe was not necessary ...

demo/README.md
swetharepakula: I did think about just giving directions for docker-compose directly and not giving directions to bring up each container separately, but I thought since this section is about bringing each individual component this would be better. I also wanted to build on the directions that Sowmya started. 

Saying that I am open to removing these directions and only using docker-compose. The difficulty is I don't think you can use docker-compose to bring about a single service at a time. How do you think we should proceed with these directions? We don't necessarily need to have individualized directions for each component.

demo/scripts/startFPCAuctionNetwork.sh
swetharepakula: I have to look into that. I have never built images using docker-compose

demo/README.md
g2flyer: i'm pretty sure you can manipulate a single service with docker-compose, including starting it up.

demo/client/backend/mock/api/serverapi.json
```diff
@@ -36,7 +36,7 @@
           {
             "id": 2,
             "name": "b",
-            "impairment": 0
```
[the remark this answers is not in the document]
swetharepakula: There was a problem with the auction cc not being able to convert json values of 0 in the c++ auction chaincode. So this is more leftover from what I did locally to get the json to work.

demo/scripts/startFPCAuctionNetwork.sh
```diff
@@ -0,0 +1,45 @@
...
+. ${SCRIPT_DIR}/lib/common.sh
+
+# Cleanup any previous iterations of the demo
+"${SCRIPT_DIR}/teardown.sh" --clean-slate
+
+# Generate the necessary credentials and start the FPC network
+"${SCRIPT_DIR}/generate.sh"
+"${SCRIPT_DIR}/start.sh"
+
+# Install and Instantiate Auction Chaincode
```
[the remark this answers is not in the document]
swetharepakula: One way we can achieve not rebuilding this is by not cleaning up the docker images in the teardown script. So once the image is built it shouldn't need to rebuild since there will be a local image. I can make that one of those `--clean-slate` flag options?

demo/scripts/startFPCAuctionNetwork.sh
g2flyer: it's pretty straight forward, .e.,g below an example from PDO:
```
..... 
 pdo-build:
    image: pdo-sim-build
    build:
      context: .
      dockerfile: Dockerfile.pdo-build
      args:
        PDO_REPO_URL: ${PDO_REPO_URL:-https://github.com/hyperledger-labs/private-data-objects.git}
        PDO_REPO_BRANCH: ${PDO_REPO_BRANCH:-master}
        SGX_MODE: SIM
        PDO_INTERPRETER: ${PDO_INTERPRETER:-gipsy}
....
```

utils/docker-compose/network-config/base/base.yaml
```diff
@@ -37,6 +38,7 @@ services:
         - /var/run/:/host/var/run/
         - ../config:/etc/hyperledger/configtx
         - ../../../../examples:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/examples
```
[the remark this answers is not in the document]
swetharepakula: I like that idea, but I have to play around a little before I know what would be the best way to approach it. The different paths was one of the reasons I mounted the entire demo directory instead of just the golang chaincode. 

The other issue is with different paths, if they have to be mounted to different locations, then the rest of the scripts/instructions might not work.

demo/scripts/teardown.sh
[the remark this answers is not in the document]
swetharepakula: Oh I must have misunderstood that conversation. I understood it as the CA volumes only need to be wiped if the certificates have changed. Then in that case I can remove the flag. That was my only reason for adding the `clean-slate`

utils/docker-compose/scripts/bootstrap.sh
```diff
@@ -161,34 +161,17 @@ DOCKER=true
...
-  export FABRIC_TAG=${MARCH}-${VERSION}
-  export CA_TAG=${MARCH}-${CA_VERSION}
-  export THIRDPARTY_TAG=${MARCH}-${THIRDPARTY_IMAGE_VERSION}
-else
-  # starting with 1.2.0, multi-arch images will be default
-  : ${CA_TAG:="$CA_VERSION"}
-  : ${FABRIC_TAG:="$VERSION"}
-  : ${THIRDPARTY_TAG:="$THIRDPARTY_IMAGE_VERSION"}
-fi
+export FABRIC_TAG=${CA_TAG:="$CA_VERSION"}
```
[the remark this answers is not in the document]
swetharepakula: Yes I did, thanks for catching that!

demo/README.md
swetharepakula: Oh, that's easy enough. I can update the instructions for that.

demo/scripts/startFPCAuctionNetwork.sh
swetharepakula: perfect! i was struggling finding non trivial build docker-compose files in the docker docs.

demo/README.md
```diff
@@ -1,32 +1,97 @@
-## Goal: Bring up Clock Auction Demo Application
+# Clock Auction Demo Application
 
-###  Prerequisites
+##  Prerequisites
 - It is assumed that Fabric Private Chaincode repository on your machine is installed in $FPC_PATH.
 - JSON processor, jq is installed.
+- FPC Peer Docker Image has been created as per the FPC Network
+[Instructions](../utils/docker-compose/README.md#Steps) in Step 1.
+- Relevant Fabric binaires and docker images have been downloaded as per FPC
```
sjanakir2015: Typo in "binaries"

demo/client/backend/mock/api/serverapi.json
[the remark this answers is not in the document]
sjanakir2015: what if the value needs to be 0?

demo/README.md
```diff
@@ -1,32 +1,97 @@
...
+If you prefer to install it manually use the following steps.
+
+1. Exec into the peer container
+```
+docker exec -it peer0.org1.example.com bash
+```
+
+2. Install the mockcc
+```
+${PEER_CMD} chaincode install -n mockcc -v 1.0 --path github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/golang/cmd -l golang
```
sjanakir2015: Not sure where the variable, PEER_CMD is set.  It may be helpful to call it out here or set it here.

demo/client/backend/mock/api/serverapi.json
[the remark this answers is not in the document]
swetharepakula: Verified with Bruno, the impairment can be 0 so removed this change in a fixup commit. So there should be no change once all squashed.

utils/docker-compose/scripts/bootstrap.sh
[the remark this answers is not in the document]
swetharepakula: fixed

demo/scripts/teardown.sh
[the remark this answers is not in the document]
swetharepakula: Added a `clean-slate` flag to the demo teardown script. When the flag is added it will run the FPC network teardown with the clean-slate and get rid of the mockcc chaincode image as well.

demo/scripts/installCC.sh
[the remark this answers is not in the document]
swetharepakula: The environment variables are unnecessary, so these should be removed now. In the eventual multi-org setup the only thing we need to making into a variable will be the peer container name. I am going to leave the peer container hard-coded for now, but this should be easy to adapt for the multi-org setup.

demo/README.md
swetharepakula: Updated instructions so that the auction client image is now built and run in one step while using the docker-compose backend. The user will need to provide the COMPOSE_PROJECT_NAME when running the command though.

demo/README.md
swetharepakula: Similar to above, docker-compose can be used to build and run the auction frontend in one step. However the COMPOSE_PROJECT_NAME must be provided now.

demo/scripts/startFPCAuctionNetwork.sh
swetharepakula: docker images have been updated so that the auction client and auction frontend are built if not already there when using docker-compose to start the services.

demo/README.md
swetharepakula: thanks! should be fixed in one of the fixup commits.

demo/README.md
swetharepakula: I can add a little to the documentation saying that the environment variables, if not explicitly set here are set in the FPC network setup. It is explained in the FPC network docs, so I don't want to replicate all the information because I think it will be harder to maintain in the future.

swetharepakula: I added a link to the FPC documentation and mention there are environment variables already set in the container to make the steps more convenient in a fixup commit

demo/scripts/startFPCAuctionNetwork.sh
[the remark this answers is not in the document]
swetharepakula: I believe this is resolved in the sense that the teardown will not bring down the chaincode image unless `--clean-state` is given so the chaincode image does not need to be recreated every time. But as you pointed out, it is slightly unavoidable to restart from scratch if you want to re-install the CC

swetharepakula:
The only comment that hasn't been addressed yet as been the examples and demo mount in the peer. I do agree with @g2flyer that it is a little weird to see when the network is meant to be general. I am still thinking about how to go about that.

Other than that I think the PR is ready for another round of reviews :)

demo/scripts/startFPCAuctionNetwork.sh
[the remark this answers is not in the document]
g2flyer: you do though still a hardcoded --clean-state in the start, so the option to not requiring it in teardown is not thaaat useful?

swetharepakula:
I will autosquash and push it up so someone can merge it in. Thanks!

swetharepakula:
I have autosquashed the fixup commits and rebased over the license PR since that one looks ready to be merged in

Issue #183, opened 2019-10-12
Deterministic (de-randomized) state encryption & randomness


g2flyer:
To enable multi-enclave endorsement, state encryption should be deterministic yet should still provide semantic security.  Additionally, we should be careful with bounds on numbers of encryption per key where in our case an enclave might be easily used as a fast "encryption-oracle" which also would suggest some de-randomized approach to increase these bounds.

Additionally, such an approach also could potentially enable a secure way to give access to chaincode to (pseudo-)random values, something which is generally very difficult to achieve in "normal" smart contracts but unlocks additional smart contract functionalities.

g2flyer:
On a high-level, the approach is to derive a key from the contract master key and then derive a transaction-encryption-randomness seed from that key using a PRF with the _complete_ transaction context (state, inputs) as input. This seed can the be taken to seed a PRNG which generates the randomness (e.g., IV for AES-GCM) during the state encryption.  This assumes that the actual chaincode (and the shim) is deterministic (which is anyway a requirement). This approach should ensure that randomness should be identical for a repeated execution of the same transaction/state yet is (cryptographically) independent for different transactions. Additionally, the randomness should be completely unpredictable, even for insiders.  

There are though some challenges with above to address/think through: it is paramount that not only encryption state but also inputs are considered when computing the PRF.  One challenge related to this are sub-transactions as they might(?) potentially run on different states and hence returned results might make the computation deviate with different plaintext being encrypted with the same randomness. In the case of AES-GCM this might completely break the integrity and might also break confidentiality (as it would leak leak relations of plaintexts between the different code-paths). A similar effect would happen if the chaincode itself (or the shim) is not deterministic.  Using AES-GCM-SIV should largely eliminate this. AES-GCM-SIV should also improve to some extent the "fast-encryption-oracle" issue, although maybe additional levels of key-diversification might be useful to go beyond the (not-so-hard-to-reach) 2^32 encryptions bound with AES-GCM.

Above approach should also work directly to expose randomness to the application (important, though, that the user-level transaction-encryption-randomness seed is independent from the state-encryption version thereof, implying an additional level of key-derivation or tweaking the PRF.

g2flyer:
Relevant prior work:
- Duan and Haibin, Practical State Machine Replication with Confidentiality, SRDS2016, https://doi.org/10.1109/SRDS.2016.031
- AES-GCM-SIV:
[...]

PR #202, opened 2020-01-16
Spectrum auction chaincode

bvavala:
This PR contains:
* the spectrum auction logic based on documents of the Federal Communications Commission (FCC)
* (very) basic tests
* an extension of `ecc_enclave` crypto APIs to expose (and use in the auction) the random number generator
The chaincode implements a simple randomized version of the Assignment phase of the auction, and so it does not implement the `submitAssignmentBid`.

Note: this chaincode uses the random number generator provided by Intel SGX. Therefore, such chaincode is suitable only for single-endorser Fabric applications. This is related to #183 .

### DIFF
--- a/ecc_enclave/enclave/crypto.h
+++ b/ecc_enclave/enclave/crypto.h
@@ -23,3 +23,4 @@ int decrypt_state(sgx_aes_gcm_128bit_key_t* key,
     uint32_t cipher_len,
     uint8_t* plain,
     uint32_t plain_len);
+int get_random_bytes(uint8_t* buffer, size_t length);
--- a/demo/chaincode/fpc/CMakeLists.txt
+++ b/demo/chaincode/fpc/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Copyright 2019 Intel Corporation
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.5.1)
+
+set(SOURCE_FILES
+    dispatcher.cpp
+    entry.cpp
+    error-codes.cpp
+    spectrum-auction.cpp
+    spectrum-auction-message.cpp
+    storage.cpp
+    principal.cpp
+    territory.cpp
+    eligibility.cpp
+    bidder.cpp
+    auction-state.cpp
+    utils.cpp
+    evaluate.cpp
+    bid.cpp
+    )
+
+include($ENV{FPC_PATH}/ecc_enclave/enclave/CMakeLists-common-app-enclave.txt)
+
--- a/demo/chaincode/fpc/Makefile
+++ b/demo/chaincode/fpc/Makefile
@@ -1,0 +1,22 @@
+# Copyright 2019 Intel Corporation
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+TOP = $(FPC_PATH)
+include $(TOP)/build.mk
+
+BUILD_DIR := _build
+
+$(BUILD_DIR):
+	@if [ ! -d $(BUILD_DIR) ]; then \
+		mkdir -p $(BUILD_DIR) && \
+		cd $(BUILD_DIR) && \
+		cmake ./..; \
+	fi
+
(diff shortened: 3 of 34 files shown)


g2flyer:
A few quick thoughts:
- this is a fairly big PR. Would it make sense for Bruno on Tuesday to go over the code and give the lay-of-the-land (e.g., general structure, what's where, conventions used ...)?
- the chaincode should probably move one level done from demo/chaincode to demo/chaincode/fpc or it obscures the golang mock chaincode ...
- would be also useful to rebase to master so we have (part of ) the network automation

mbrandenburger:
> * this is a fairly big PR. Would it make sense for Bruno on Tuesday to go over the code and give the lay-of-the-land (e.g., general structure, what's where, conventions used ...)?

I would really appreciate this @bvavala! This one is big.

bvavala:
> A few quick thoughts:
> 
> * this is a fairly big PR. Would it make sense for Bruno on Tuesday to go over the code and give the lay-of-the-land (e.g., general structure, what's where, conventions used ...)?
> * the chaincode should probably move one level done from demo/chaincode to demo/chaincode/fpc or it obscures the golang mock chaincode ...
> * would be also useful to rebase to master so we have (part of ) the network automation
> 
> PS: while many commits seem from message somewhat self-contained, is the idea to maintain all of them or consolidate? If the latter, might be nice to do already right now? (might also facilitate fixing the DCO :-)

* yes I can give an overview
* rebased to master
* chaincode folder is `demo/chaincode/fpc`

ecc_enclave/enclave/crypto.h
```diff
@@ -23,3 +23,4 @@ int decrypt_state(sgx_aes_gcm_128bit_key_t* key,
     uint32_t cipher_len,
     uint8_t* plain,
     uint32_t plain_len);
```
g2flyer: hmm, couldn't you just have defined that inside your own chaincode? I don't think crypto.h should be exported as part of the shim.  Eventually we want to have some extension to shim.h which provides randomness, but i think we might need a bit more discussion on API and design thereof ...

bvavala: since you did, I certainly could too!
Obviously this dirties the abstraction either above in the chaincode (your case) or below in the shim (my case).

Main reasons to keep this:
* we need it in the infrastructure (and this is the second application already that needs it). Yes, we will need discussions on the randomness thing. 
* crypto already uses it, and it is already interfaced with the SGX sdk. So, why exposing that also to the chaincode!

g2flyer: > since you did, I certainly could too!
> Obviously this dirties the abstraction either above in the chaincode (your case) or below in the shim (my case).
> 
> Main reasons to keep this:
> 
> * we need it in the infrastructure (and this is the second application already that needs it). Yes, we will need discussions on the randomness thing.
> * crypto already uses it, and it is already interfaced with the SGX sdk. So, why exposing that also to the chaincode!
> 
> Actually, now I see that: (1) the crypto stuff could use it; (2) crypto does not check the return code of the rng (I'll put an issue once merged).

My main objection is that you do it in crypto.h.  The main (only) interface fpc chaincode is supposed to have is shim.h.  Obviously, in absense of a module system in C (or some build trickery to remove symbols from libraries) nothing stops CC to use symbols from other header-files/libraries right now but might not work in future when we might have a binary SDK which would (should) not contain internal header files.

bvavala: I have removed the crypto header dependency.

g2flyer: Well, you haven't removed the dependency on the shim internals/crypto ;-) If you want to leave that, then maybe leaving the include as before and add a comment on why this is kind of unkosher (and what the current constraint is) would seem cleaner.  Or we just take the current get_random_bytes api and add it as-is to shim.h as first stab on such a shim API function (more or less with description that this functions is supposed to give you randomness which is identical if run on different peers iff is on exactly same CC with same state and input or otherwise the randomness is independent.

All that said probably not worth spending too much time on it now for the demo (although i think a comment would still be useful)

bvavala: alright, dependency on `shim.h`
added comment

PR #203, opened 2020-01-16
Demo Scripting ...

g2flyer:
Initial proposal for a scripting utility.  See demo/client/scripting/scenario-run.sh as the main orchestration, demo/client/scripting/lib/dsl.sh for the simple "DSL" used to script it.   Also see the TODO in dsl.sh and cli.go for potential next steps .....


demo/client/scripting/lib/dsl.sh
```diff
@@ -0,0 +1,107 @@
...
+#---------------------------------------------
+#
+# TODO (maybe)
+# - get auction status after submit and output when rounds/auction is finished?
+#
+# terms of language are
+# - scenario [<scenario-path>] if path is missing, take directory where script file is located
+# - "submit" <user> <action>
+# - "wait" <message> [<variable-for-user-input>]
+
```
mbrandenburger: another useful thing might be a optional `sleep` after each action to slow things down if needed for "visualisation". This can be added later as a feature

mbrandenburger: a `comment` action would be also very useful to link the lines in the console with the script while debugging os just showing line numbers?

demo/scenario/script
```diff
@@ -0,0 +1,29 @@
+# Copyright Intel Corp All Rights Reserved
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+#  Scripted scenario
+#  (to be excuted by demo/client/scripting/scenario-run.sh)
+#------------------------------------------------------------
+
+scenario
```
mbrandenburger: Maybe rename `/demo/scenario` -> `/demo/scenarios` and thus give some space for different scenarios. Jeb would work a `live demo` scenario, whereas bruno might need a `testing` scenario. `short_video`, `long_video`, can we get the FCC traces and put it in `the_real_fcc_auction`?

demo/client/scripting/lib/dsl.sh
g2flyer: > another useful thing might be a optional `sleep` after each action to slow things down if needed for "visualisation". This can be added later as a feature

for non-mock you probably have to worry less about speed :-) But i guess one could add a global option/variable `delayTime` or alike and use this to add a delay for each submit (note this is a simple script file, so if you want selective delays, you can just sprinkle the script with `sleep 2` as-is)

g2flyer: > a `comment` action would be also very useful to link the lines in the console with the script while debugging os just showing line numbers?

Not 100% sure i understand: do you just mean a `debug` flag or alike which triggers all actions to print more info, e.g., line numbers?  I guess this could be done using `caller` although i wouldn't have expected these scripts becoming so long that it becomes and issue.  Note you can also easily get debug info by running script as `bash -x ....` and/or `bash -vx ...` ..

demo/scenario/script
g2flyer: the script is certainly designed to have different scenarios (sequence of transactions) and different scripting of a particular scenario (more or less of the transactions are through UI).  if you just have different scripts for same transactions, i would have kept it in the same directory. so long and short might be according to my terminology same scenario. FCC trace certainly would be a different one and require a different (sub-)directory.  But i think we can do that once we get there .... :-)

PR #205, opened 2020-01-22
Ui updates

mbrandenburger:
This PR adds some nice visual improvements and bug fixes.


demo/client/frontend/src/components/Bidding/ClockBidding.vue
```diff
@@ -5,168 +5,205 @@ SPDX-License-Identifier: Apache-2.0
...
+                append-icon="fa-search"
+                label="Search"
+                single-line
+                hide-details
+              />
+            </v-card-title>
+
+            <v-card-text class="pa-0">
+              <v-data-table
+                v-if="auction.currentRound === 1"
```
bvavala: this should be leftover!?

mbrandenburger: fixed

g2flyer:
some UI changes seem to rely on additional new features in the mock backend.  Does UI, though, still work with fabric-client backend (with probably some reduced functionality)? Maybe make somewhere explicit which features work only with mock?

mbrandenburger:
> some UI changes seem to rely on additional new features in the mock backend. Does UI, though, still work with fabric-client backend (with probably some reduced functionality)? Maybe make somewhere explicit which features work only with mock?

I added some comments saying that notification stream and the debug view is currently only supported by the mock server. I also tested the UI using the gateway and it seems that the missing notification stream does not impact the functionality of the normal auction UI. Even though, an error is thrown in the javascript console, this one can be ignored.

sjanakir2015:
> some UI changes seem to rely on additional new features in the mock backend. Does UI, though, still work with fabric-client backend (with probably some reduced functionality)? Maybe make somewhere explicit which features work only with mock?

@mbrandenburger 
I tested this with fabric-gateway backend.  UI seems to be functional although I did not do a thorough test to exercise all chaincode calls.  
1.  I noticed that call to  `localhost:3000/api/notification`  results in an error - only noted in Developer Tools / Console.  (you have already responded to this)

2. Noted the following error in Developer Tools / Console.  Hyperledger Fabric logo in the left bottom (present in the video posted on slack) is missing.  Is it picked up from a local mount when you run the application?
```
webpack-internal:///./node_modules/vuetify/lib/util/console.js:37 [Vuetify] Image load failed
src: ./img/fabric-logo.png
```
3.  CORS issue found.  To resolve this, in file: demo/client/backend/fabric-gateway/src/server.js, please replace the line 
```
res.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept, x-user');
```
with the following line: 
```
res.header('Access-Control-Allow-Headers', 'Access-Control-Request-Method, Access-Control-Request-Headers, Origin, X-Requested-With, Content-Type, Accept, x-user');

```

mbrandenburger:
> > ```
> webpack-internal:///./node_modules/vuetify/lib/util/console.js:37 [Vuetify] Image load failed
> src: ./img/fabric-logo.png
> ```
> 

Did you try to rebuild the frontend container? the image is part of the commit. Cannot reproduce. :/


> 1. CORS issue found.  To resolve this, in file: demo/client/backend/fabric-gateway/src/server.js, please replace the line
> 
> ```
> res.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept, x-user');
> ```
> 
> with the following line:
> 
> ```
> res.header('Access-Control-Allow-Headers', 'Access-Control-Request-Method, Access-Control-Request-Headers, Origin, X-Requested-With, Content-Type, Accept, x-user');
> ```

Fixed now!

mbrandenburger:
Rebased and autosquashed!

mbrandenburger:
Did another rebase after merging #209

g2flyer:
A few observations after running the UI against mock server and demo/chaincode/fpc (all rebuilt from the version in this PR):

- The following buttons didn't do anything (and no visible error was shown in dev tools or alike). I assume they are all not yet implemented?  (Maybe might be useful to have somewhere a list on what is supposed to work and should be tested?): 'Auction History', 'Bid Summary', 'My results', 'Settings', 'Help'.
- in auctioneer view, the order of buttons as 'create', 'endRound', 'startNextRound' seems a bit confusing; i would have expected it to be 'create', 'start', 'end'
- i would assumed that either startNextRound or endRound would have increased the round number but it was always one (i have to admit, i only tested where there was a single round with a single bid which resulted in the first endround to close clocking and the following startnextround performed assignment phase and closed auction
- the auction status as "closed" seems a bit confusing; gives the impression of the auction being closed rather than just bidding is not open yet.  Maybe rephrasing it as 'waiting for round to be opened' or alike would be clearer.   There also didn't seem a difference between the round being inactive and the auction actually being finished? Related to that i was not sure, though, also how i could refresh the view given that stuff happens asynchronously in the background (auto-refresh could of course also work but i don't think the app currently does it, at least i couldn't see corresponding network events)
- is there currently any way to view the final results? Couldn't find it.
- might be useful to have somewhere visible (and editable) the auction id?  For mock server you can easily reset the state so you can again create auction with auctionid 1 but with real fabric network this gets much more complex/lengthy?
- in the transaction viewer you currently also show transaction which failed (not that this is per-se a problem for the demo, just noticed ...)

Anyway, above my notes playing with the UI a bit.  Doesn't have to be all addressed in this PR, i.e., i'm perfectly fine merging it as-is, but at least for the first question (i.e., what is supposed to work and what not) would be useful to know 
-

PR #207, opened 2020-01-28
Add Docker Chaincode Builder Script

swetharepakula:
- add minimum build to make file to only build components needed to
 build FPC chaincodes
 - add cc-builder image
 - add make target to utils/docker to build the cc-builder

Signed-off-by: Swetha Repakula <srepaku@us.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ include $(TOP)/build.mk
 
 SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
 PLUGINS = ercc ecc_enclave ecc tlcc_enclave tlcc
+FPC_SDK = utils/fabric ecc_enclave ecc
 
 .PHONY: license
 
@@ -45,3 +46,6 @@ godeps: gotools
 
 plugins:
 	$(foreach DIR, $(PLUGINS), $(MAKE) -C $(DIR) build || exit;)
+
+fpc-sdk: godeps
+	$(foreach DIR, $(FPC_SDK), $(MAKE) -C $(DIR) build || exit;)
--- a/utils/docker/cc-builder/Dockerfile
+++ b/utils/docker/cc-builder/Dockerfile
@@ -1,0 +1,10 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ include $(TOP)/build.mk
 
 SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
 PLUGINS = ercc ecc_enclave ecc tlcc_enclave tlcc
+MIN = utils/fabric ercc ecc_enclave ecc
```
g2flyer: MIN is a bit MIN as describing purpose ;-)   If i see that correctly, then this is essentially the target to build the "FPC chaincode SDK"? If so, maybe a corresponding name might be better in describing the purpose of the target?

utils/docker/cc-builder/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM hyperledger/fabric-private-chaincode-dev:latest
+
+RUN cd $FPC_PATH
+RUN make min
+
+WORKDIR $FPC_PATH/demo/chaincode
```
g2flyer: I would have imagined this is a generic cc builder where you "inject" the code via volume and then call make (or whatever you used to build the chaincode) in that injected volume. In that case, there wouldn't be any need for WORKDIR?  Seems a bit strange to have this WORKDIR here pointing to demo ...

Makefile
mbrandenburger: ercc should not be part of the FPC SDK - right?! 

I really like the description of FPC SDK and maybe FPC runtime?!

utils/docker/cc-builder/Dockerfile
mbrandenburger: I agree with Michael

mbrandenburger: I suggest, for now we can keep it but add a comment here (maybe also the docker command that it needs to mount some chainode inside this container)

swetharepakula:
verified that `ercc` is not needed, and updated the names of the target and the variable. Should be ready to merge now.

PR #208, opened 2020-01-28
Script & json-files for demo auction

jrlinton:
Adding demo script for short video of Spectrum Auction

### DIFF
--- a/demo/scenario/Auctioneer1.createAuction.json
+++ b/demo/scenario/Auctioneer1.createAuction.json
@@ -1,7 +1,7 @@
 {
-	"name": "Spectrum Auction 2020",
+	"name": "Elbonian Communications Commission Spectrum Auction 2020",
 	"activityRequirementPercentage": 95,
-	"clockPriceIncrementPercentage": 10,
+	"clockPriceIncrementPercentage": 20,
 	"territories": [{
 		"id": 1,
 		"name": "Elbograd",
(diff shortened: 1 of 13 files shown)


demo/scenario/B-Net.submitClockBid.4.json
```diff
@@ -0,0 +1,15 @@
...
+    "round": 4,
+    "bids": [
+	{
+	    "terId": 1,
+	    "qty": 3,
+	    "price":14400
+	},{
+	    "terId": 2,
+	    "qty": 1,
+	    "price":5500
```
bvavala: bid in ter 2 is below posted price. so it fails. is this intended?

demo/scenario/script
```diff
@@ -17,10 +17,37 @@ submit Auctioneer1 createAuction
...
+submit Auctioneer1 startNextRound
+submit A-Telecom submitClockBid
+submit B-Net submitClockBid
+submit C-Mobile submitClockBid
+submit Auctioneer1 endRound
+
+wait "End Round 2  (Hit any key to continue)"
+
+submit Auctioneer1 startNextRound
+wait "enter C-Mobile bid manually ...  (Hit any key to continue)"
```
bvavala: is this "enter manually" intended?
If I don't enter anything, or I enter the actual C-mobile bid, there are different outcomes.

demo/scenario/script
```diff
@@ -17,10 +17,37 @@ submit Auctioneer1 createAuction
...
+
+wait "End Round 3  (Hit any key to continue)"
+
+submit Auctioneer1 startNextRound
+submit A-Telecom submitClockBid
+submit B-Net submitClockBid
+submit C-Mobile submitClockBid
+submit Auctioneer1 endRound
+
+wait "End Round 4 - Clock Phase Complete  (Hit any key to continue)"
```
bvavala: this is true only if you submit the bid for C-mobile in round 3

demo/scenario/script
```diff
@@ -17,10 +17,37 @@ submit Auctioneer1 createAuction
...
+
 # assignment rounds ...
 submit Auctioneer1 startNextRound
 # bidding not yet implemented, add here once done ....
```
bvavala: This is not a problem but note that: the end round here fails because the assignment phase is started **and terminated**  already in `startNextRound`.

demo/scenario/Auctioneer1.createAuction.json
```diff
@@ -50,94 +50,6 @@
 		}]
 	}, {
 		"id": 2,
-		"name": "Mudberg",
```
mbrandenburger: Removing two territories is a good simplification ... but on the other hand it does not look so "good" anymore in the UI. Is there a simple way that we keep all four territories but don't really "use" them in the auction. We dont even have to mention them. I guess the problem is that in order to finish the auction there needs to be some telco to bid for these licences. But using the scenario script this bid can just happen in the background without mentioning it at all?

demo/scenario/script
mbrandenburger: I think we need a comment here what C-mobile has to enter now in order to continue the scenario as intended.

g2flyer:
Note: issue identified by Bruno will be solved in a follow-up PR by Michael which also will add some run-scenario script improvements ...

Issue #199, opened 2020-01-09
Generalize demo-related mock & fabric-gateway backends as well as return/error handling.


g2flyer:
The code in demo/client/backend is currently auction-specific but has the potential of being much more widely usable to either efficiently test chaincode (mock) or running UI apps (fabric-gateway). Note this is not even necessarily FPC specific but would apply also to other fabric chaincode!

Most of the code is already generic, the exception being the 
- /api/clockauction/getDefaultAuction action 
- a few other auction-specific parts in mock 
It probably also would be  worth moving the calls to __init & __setup as well as the unpacking of the payload objects, which are specific to FPC to chaincode/fpc_chaincode.go (or handle these calls for non-fpc in chaincode/go_chaincode.go such that actual go chaincode doesn't have to know about it?)

A related issue is also whether the return value processing (see PR #196) could be generalized and abstracted into the FPC shim.  Ultimately, the underlying problem here, though, is that fabric itself doesn't offer a good way to return program-actionable (e.g., numeric error codes) but only human actionable (i.e., a string) error messages.  So doing it only in FPC would make the FPC programming model deviate from fabric and cleaner would be to extend/change fabric itself (by e.g., adding an enum or alike to the shim response).

g2flyer:
[keyhole-fabric-api-gateway](https://github.com/hyperledger-labs/keyhole-fabric-api-gateway) which just got [accepted into hyperledger labs](https://lists.hyperledger.org/g/labs/topic/74030894) seems also to get in that direction (and beyond ...)

PR #211, opened 2020-01-31
Demo Scripting Improvements

g2flyer:
* enhanced demo script DSL which
  - provides a new verb submit_manual which provides the template
    of what should be manually entered
  - has dry-run and non-interactive modes for script/json-file debugging and verification
  - does extract rc from status code and verifies against expected return code
* addressed issues identified in review of PR #210, e.g., added the two previously
  deleted territories back and fixed some typos
* added missing Makefiles


demo/scenario/script
```diff
@@ -9,36 +9,35 @@
...
+say "End Round 1"
 
 submit Auctioneer1 startNextRound
 submit A-Telecom submitClockBid
 submit B-Net submitClockBid
 submit C-Mobile submitClockBid
 submit Auctioneer1 endRound
 
-wait "End Round 2  (Hit any key to continue)"
+yell "End Round 2"
```
mbrandenburger: I guess a `say` would also work here ....

Makefile
```diff
@@ -6,7 +6,7 @@
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration # docs
+SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration demo # docs
```
mbrandenburger: I would not include demo to the toplevel make at the moment - we maybe promote this in the future to a dedicated repo; or move it to examples or it just stays here. But I still think that a `demo/Makefile` makes sense.

mbrandenburger: or if we add it to the top level make; let's build the demo bu just calling `make demo` but not including it into `make`.

demo/scenario/Auctioneer1.createAuction.json
```diff
@@ -78,7 +78,96 @@
 			"name": "f",
 			"impairment": 30
 		}]
-	}], 
+
+	}, {
+                "id": 3,
+                "name": "Mudberg",
```
bvavala: fine adding the territories for better visual look, yet... it looks complicated again with 10 channels each! How about 1 channel each?

demo/scenario/script
g2flyer: I put a yell as round number seemed relevant to know for the subsequent manual step. But had no strong feelings. Just thought we probably do not want a wait here but still somewhat highlight the new round ...

Makefile
g2flyer: Hmm, not sure why eventually outsourcing demo (with which i agree, although as discussed in issue #199 we might also keep some stuff in generalized form in the main repo ...) should have any implication on building or not?   My motivation was two-folds:
- to make sure that demo artifacts are build when you build the overall repo and
- make sure travis also touches the demo on PRs, by at very least building it.
What is the problem of building demo in main build? It's not that it extends build time drastically or alike?  Also 'make demo' might actually be a better target to potentially (eventually) automatically set up the whole demo via a top-level make?

demo/scenario/Auctioneer1.createAuction.json
g2flyer: Which view are you referring by "looks complicated"?  I thought currently in demo we don't show the createAuction and only the final result would show channels, a view i haven't seen yet.   That said, the number of channels looks in general high but i think we might want to reduce it overall more uniformally down to something like 3-5 rather than having some 10 channel and some 1 channel territories?  In any case, I think we will have to check it with the actual views (and based on some expected screen real estate (e.g., what resolution do we expect with what font-size)?

Makefile
bvavala: I have mixed feelings about this: I agree with Marcus (yes, we discussed that); but also, as we update FPC, I would not like the demo to simply be "forgotten" and never work anymore because it's not maintained.
My conclusion is: (i) let's discuss if we should move it elsewhere, (ii) let's maintain it while it is in the repo (which is the good practice to follow).

mbrandenburger: as discussed today we keep the demo in the repo and therefore also integrate it in the main make to ensure that travis has the chance to detect issues there.

demo/scenario/script
mbrandenburger: ok

Issue #81, opened 2019-06-13
return value of ocalls (e.g., ocall_get_state) are not checked


bvavala:

KimRasz:
Generic debuggin

PR #214, opened 2020-02-12
fix bug: binary array used as string

bvavala:

### DIFF
--- a/ecc_enclave/enclave/shim.cpp
+++ b/ecc_enclave/enclave/shim.cpp
@@ -41,7 +41,7 @@ void get_state(
     const char* key, uint8_t* val, uint32_t max_val_len, uint32_t* val_len, shim_ctx_ptr_t ctx)
 {
     uint8_t encoded_cipher[(max_val_len + SGX_AESGCM_IV_SIZE + SGX_AESGCM_MAC_SIZE + 2) / 3 * 4];
-    uint32_t encoded_cipher_len;
+    uint32_t encoded_cipher_len = 0;
 
     get_public_state(key, encoded_cipher, sizeof(encoded_cipher), &encoded_cipher_len, ctx);
 
@@ -53,7 +53,7 @@ void get_state(
     }
     if (encoded_cipher_len > sizeof(encoded_cipher))
     {
-        char s[] = "Enclave: encoded_cipher_len greater than (or equal to) buffer length";
+        char s[] = "Enclave: encoded_cipher_len greater than buffer length";
         LOG_ERROR("%s", s);
         throw std::runtime_error(s);
     }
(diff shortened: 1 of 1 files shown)


ecc_enclave/enclave/shim.cpp
```diff
@@ -40,8 +40,8 @@ void get_creator_name(
 void get_state(
     const char* key, uint8_t* val, uint32_t max_val_len, uint32_t* val_len, shim_ctx_ptr_t ctx)
 {
-    uint8_t encoded_cipher[(max_val_len + SGX_AESGCM_IV_SIZE + SGX_AESGCM_MAC_SIZE + 2) / 3 * 4];
     uint32_t encoded_cipher_len;
+    uint8_t encoded_cipher[(max_val_len + SGX_AESGCM_IV_SIZE + SGX_AESGCM_MAC_SIZE + 2) / 3 * 4];
```
g2flyer: any deeper reason why you swapped the order of the two vars?

ecc_enclave/enclave/shim.cpp
```diff
@@ -51,8 +51,26 @@ void get_state(
         *val_len = 0;
         return;
     }
+    if (encoded_cipher_len > sizeof(encoded_cipher))
```
g2flyer: Hmm, this shouldn't really happen as we pass to get_public_state sizeof(encoded_cipher) as maximum buffer length. So if that happens, there is a bug elsewhere and we should fix it there? Looking at the code, though, not clear how this could happen ...
Maybe the issue is that we never initialized encoded_cipher_len and due to some errors it never gets assigned? Maybe we should set it to zero in get_*_state before doing the ocall?

ecc_enclave/enclave/shim.cpp
bvavala: no, left over!

ecc_enclave/enclave/shim.cpp
bvavala: Right, it should not happen.
However, as `encoded_cipher_len` is input data, it seems appropriate to perform input validation before building the string.

ecc_enclave/enclave/shim.cpp
g2flyer: well, spurious (non-)changes are a bit of a distraction ...

ecc_enclave/enclave/shim.cpp
g2flyer: input validation is done down in the SDK. Don't see any reason to do it again here and clutter the code. And if so, we should do it everywhere else as well?

bvavala: the input **to this function** is not validated. I don't see why you would not want to do it. Actually, we shall check the buffers as well.

g2flyer: i'm missing something here: neither encoded_cipher_len nor encoded_cipher are inputs but local variables, so you are only checking return values from our own get_public_state function?

bvavala: right -- though I don't care that it is our own `get_public_state`.
Also, the `get_public_state` grabs this value from the ocall and, again, does not validate the range of the value.

g2flyer: but then we should move it to get_public_state, not get_state?

bvavala: I think that we should rather implement the check there too. Yes, it is redundant, but I don't know how these functions will evolve over time. So I prefer to follow "per function" due diligence.

g2flyer: after thinking again what the SDK does and not does, i think adding the check immediately after the ocalls, e.g., in get_public_state (and also after the ocall to ocall_get_state_by_partial_composite_key) is actually a good idea as the SDK almost certainly does not check this (i was wrong on this in my earlier comment). However, i would object to do it get_state where we call our own function as this is redundant and blurs the really important place for the checks, i.e., immediately after the ocall returns,

bvavala: (this is related to #81 )

PR #215, opened 2020-02-12
PublishAssignmentResults, test chaincode in build; short circuit assignment phase

bvavala:
This PR implements:
* the `PublishAssignmentResults` chaincode api
* the test of the auction chaincode in the build process
* an improvement of the Assignment phase short circuit: when a round ends and this also results in the termination of the clock phase, then the Assignment phase is immediately started and terminated with a random assignment; finally, the auction is closed. All these operations happen during the `end round` call.


bvavala:
current build fails because of bug identified in #214 , it should work after the fix is merged

demo/scenario/script
```diff
@@ -48,11 +48,13 @@ submit Auctioneer1 endRound
...
 
 # assignment rounds ...
-submit Auctioneer1 startNextRound
 # bidding not yet implemented, add here once done and
-# uncomment below endRound (Note: currently startNextRound in assignment
+# uncomment below startNextRound and endRound (Note: last endRound in clock
 # phase immediately assigns (randomly) the slots and terminates the phase)
+
+# submit Auctioneer1 startNextRound
 # submit Auctioneer1 endRound
```
g2flyer: Forgot to add to review, but would be nice to create a json file for a publishAssignemtResult tx here -- essentially a `cp Auctioneer1.endRound Auctioneer1.publishAssignmentResults` .. -- and then add here a `submit Auctioneer1 publishAssignementResults` ...

bvavala: done

PR #224, opened 2020-02-16
WIP: Docker-compose extensions

g2flyer:
This PR adds following additions to the fabric network scripts in `utils/docker-compose/scripts` (and transitively to the scripts in `demo/scripts`):
- running the peer using CouchDB and its graphical (state) viewer on http://localhost:5984/  when the environment variable `USE_COUCHDB` is defined.  Note: all state has mime-type octet-string, so you will have to download first the state (via "download attachment" menu) and then open that file in an editor to show the cleartext  or base64-encoded encrypted data.
- run explorer on http://localhost:8090/ when the environment variable `USE_EXPLORER` is defined.
- small bug-fix in cc-builder and a bit more robust `scenario-run.sh --bootstrap` ..


swetharepakula:
Tested it and it looks good to me. Merging.

Issue #227, opened 2020-02-18
Race condition in mock server


g2flyer:
The fabric mock ledger/stub is not thread-safe, which can lead to errors in mock server when multiple concurrent requests happen, e.g., with the demo script concurrent with UI events such as notification-triggered getX requests.  The delay in the script make these unlikely, but as randomized, not impossible.  Multiple roles with concurrent UIs/tabs also could trigger such errors, in particular as notification-triggered queries should be highly synchronized. (Also note: not only the ledger state is not concurrency controlled, but also the mock stub object which we modify with creator, txid and alike.  Hence, we need exclusive locks and cannot do read/write locks (with query=read and invoke=write) .....)

Mid-term solution would be to gate the mock stub with a lock (sync.Mutex)

g2flyer:
Fixed in #228

PR #228, opened 2020-02-19
make mock-server safe for concurrent requests

g2flyer:
Should address issue #227 

BTW: part of the tests i also tried to narrow down steps/clicks necessary for demo script.  If for C-mobile you just set qty of second territory from 5 to 4 but leave original prices and the qty of first one (contrary to what script currently suggests), the auction still will properly finish in round 4!  I.e., you need only one field modification and the two submits as actions to complete the demo in the ui ...

### DIFF
--- a/demo/client/backend/mock/server.go
+++ b/demo/client/backend/mock/server.go
@@ -129,8 +129,11 @@ func notifications(c *gin.Context) {
 }
 
 func startDemo(c *gin.Context) {
+	stub.Lock()
+	defer stub.Unlock()
+
 	// destroy enclave
-	Destroy(stub)
+	DestroyChaincode(stub)
 	notifier.Submit("restart")
 
 	// let's create a new chaincode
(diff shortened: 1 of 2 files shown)


g2flyer:
> 1. looks good
> 2. I think the `notifier` could still raise concerns, so
>    a) would you double check that?
>    b) if so, would you add the mutex for that as well?

Nope, notifier does not access any state. In fact, locking it would lead to a deadlock as this is a long-running connection ...

demo/client/backend/mock/server.go
```diff
@@ -327,20 +351,24 @@ func query(c *gin.Context) {
...
 	if err != nil {
 		logger.Error(fmt.Sprintf("Request Error: %s\n", err))
 		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
 	}
+	user := c.GetHeader("x-user")
 
 	logger.Debugf("%s queries %s", user, args)
+
+	// invoke transactions
```
mbrandenburger: better also add the RLock here?

g2flyer: no, you also need a write lock here: you are locking not only the state inside the mock object but also the mock-wrapper object which has to be protected ..

g2flyer: Oops, now i see what you mean: fixed now ...

Issue #219, opened 2020-02-14
Demo UI limitations & improvement items


g2flyer:
- [X] from walk-thru
  - [x] remove warning in ledger viewer
  - [x] simplified result view proposed by marcus, rename "Auction Results" to "Auction Result: License won"

- [ ] remaining from PR #218:
  - [ ] some way to trigger a refresh (see PR #218 for some suggestions)
  - [x] auction result view also for bidders

- [ ] remaining from PR #205 (and not yet addressed by PR #217 and #218)
   - [ ] show/allow modification/change of auction id) [important when running with fabric-gateway & real peer as otherwise you have to reset the whole environment for a new auction]
  - [ ] failed txs are not distinguished in tx viewer from successful ones (certainly not a critical issue)

- [ ] collected while preparing demo
  - [x] refresh of auction results doesn't work properly when already in the view (probably a concurrency problem)
  - [ ] submitBid info might be stale as not refreshed by notification (can be fixed, though, by quickly going to auction info page and back to place bid
  - [ ] Before final result view, auction info is "stuck" in info from before initial round.  So right now the bidder knows only implicitly what the status of his bids are (i.e., what he sees before editing the bid in bid summary) and the auctioneer knows nothing about progress other than the round number.
  - [ ]  A bit confusing for debugging in browser dev tools is also that we do a getAuctionStatus failures (Uncaught in promise) with invalid params  (status.rc=5)  when no auction exists yet on the server. Only a problem on server-resets, seems there is some race condition on when the auction.clear commit happens vs the test in App.vue for `if (this.auction.id === "")`...  
- [ ] issues mentioned elsewhere ..
  - [ ] issue #229 
  - [ ] issue #234

mbrandenburger:
> * from walk-thru
>   
>   * remove warning in ledger viewer
>   * simplified result view proposed by marcus, rename "Auction Results" to "Auction Result: License won"
> * remaining from PR #218:
>   
>   * some way to trigger a refresh (see PR #218 for some suggestions)
>   * auction result view also for bidders

These issues are covered with PR #222

mbrandenburger:
>   * some way to trigger a refresh (see PR #218 for some suggestions)

Fixed with PR #221

g2flyer:
> > * some way to trigger a refresh (see PR #218 for some suggestions)
> 
> Fixed with PR #221

[...]

PR #230, opened 2020-02-20
Minor UI updates

mbrandenburger:
- add zeros in clock bidding view
- fix fetchAssignmentResults bug in auction.state watcher
- load auction details on event notification if not loaded yet
- show auction results view only if results available


mbrandenburger:
This PR addresses issues from #219

PR #232, opened 2020-02-21
chaincode bug fix, and small script update

bvavala:
Two contributions:
* bug fix: in the chaincode, the price-point was incorrectly computed as a `double` between 0-1 and rounded to an `integer`, so always `0`. This resulted in a non-deterministic order of the queue. In the case of multiple bids (say two) that decrease demand, the last bid that made demand match supply would have determined the posted price. Being in random order, the posted price could be set either to match the first or the second bid. In the current code base, this can later induce an error in the demo script when, once the lower posted price is randomly chosen (say the first one, and determines the lower clock price), a bidder makes a bid with a price that is higher than the clock price -- this was not a problem when the second (higher) posted price was the outcome of the random choice.
* script update: first, the manual bid of C-Mobile now only requires to set the quantity of the first territory to 1; second, the final `endRound` (which closes the auction) must now be performed manually.

### DIFF
--- a/demo/chaincode/fpc/evaluate.cpp
+++ b/demo/chaincode/fpc/evaluate.cpp
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <cmath>
 #include "auction-state.h"
 #include "common.h"
 #include "shim.h"  // get_random_bytes
@@ -215,8 +216,8 @@ void ClockAuction::DynamicAuctionState::processRegularRoundBids(
             // compute price point
             double minPrice = postedPrice_[clockRound_ - 1][eBid.territoryIndex];
             double clockPrice = clockPrice_[clockRound_][eBid.territoryIndex];
-            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice);
-            eBid.pricePoint = (uint32_t)pp;
+            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice) * (100.0);
+            eBid.pricePoint = std::lround(pp);
 
             // compute random value for breaking ties
             if (get_random_bytes((uint8_t*)&eBid.randomValue, sizeof(eBid.randomValue)) != 0)
--- a/demo/scenario/C-Mobile.submitClockBid.3.json
+++ b/demo/scenario/C-Mobile.submitClockBid.3.json
@@ -4,12 +4,12 @@
     "bids": [
 	{
 	    "terId": 1,
-	    "qty": 2,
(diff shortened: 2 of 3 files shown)


demo/chaincode/fpc/evaluate.cpp
```diff
@@ -215,7 +215,7 @@ void ClockAuction::DynamicAuctionState::processRegularRoundBids(
             // compute price point
             double minPrice = postedPrice_[clockRound_ - 1][eBid.territoryIndex];
             double clockPrice = clockPrice_[clockRound_][eBid.territoryIndex];
-            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice);
+            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice) * (100.0);
             eBid.pricePoint = (uint32_t)pp;
```
g2flyer: I have to admit, i don't like type casts as they can bury easily problems: std::lround() might be nicer?

demo/chaincode/fpc/evaluate.cpp
```diff
@@ -215,7 +215,7 @@ void ClockAuction::DynamicAuctionState::processRegularRoundBids(
             // compute price point
             double minPrice = postedPrice_[clockRound_ - 1][eBid.territoryIndex];
             double clockPrice = clockPrice_[clockRound_][eBid.territoryIndex];
-            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice);
+            double pp = (eBid.demand.price_ - minPrice) / (clockPrice - minPrice) * (100.0);
```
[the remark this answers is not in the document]
bvavala: It seemed to emphasize better the three blocks and separators (/*).

demo/chaincode/fpc/evaluate.cpp
bvavala: possibly, let me change this...

demo/chaincode/fpc/evaluate.cpp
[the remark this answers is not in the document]
g2flyer: Hmm, not completely following.  To emphasize the (precedence and associativity of) operators/blocks, it would seem to me  ` pp = ( (eBid.demand.price_ - minPrice) / (clockPrice - minPrice) ) * 100.0`` would be the cleanest way?  But as long as we have the same semantics of the operation, i'm fundamentally fine ... :-)

demo/chaincode/fpc/evaluate.cpp
bvavala: done

PR #233, opened 2020-02-22
UI improvements

g2flyer:
- does not logout anymore after a server reset (can be re-enabled if desired)
- slightly modifed version of marcus' offline commits
  - sligthly slower transaction progress
  - avatars are locally loaded (but downloaded during build, not stored in repo)

PS: in browser devtools console you will see (like before) some getAuctionStatus requests failing (Uncaught in promise) with invalid params (status.rc=5) when no auction exists yet on the server. Happens during server-resets, seems there is some race condition on when the auction.clear commit happens vs the test in App.vue for if (this.auction.id === "")...

### DIFF
--- a/demo/client/frontend/.env.development
+++ b/demo/client/frontend/.env.development
@@ -1,1 +1,6 @@
+# Copyright IBM Corp. All Rights Reserved.
+# Copyright Intel Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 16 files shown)


g2flyer:
> Thanks Michael! The reset switch works as described! The clock bidding view has still a bug, see fix suggestion below.

Thanks for catching that: i've incorporated it and auto-squashed it ...

PR #237, opened 2020-02-28
Updated README to be clearer on how to develop using the Docker Conta…

annumberhocker:
Update README to be clearer on how to develop using the Docker Container Environment vs Local host setup.

Signed-off-by: Ann Umberhocker <ann.umberhocker@ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,64 @@ your [development environment](https://hyperledger-fabric.readthedocs.io/en/rele
 
 Moreover, we assume that you are familiar with the Intel SGX SDK.
 
+### **Initial Setup**
+
+#### 1. Intel SGX SDK and SSL
+
+Fabric Private Chaincode requires the Intel [SGX SDK](https://github.com/intel/linux-sgx) and
+[SGX SSL](https://github.com/intel/intel-sgx-ssl) to build the main components of our framework and to develop and build
+your first private chaincode.     
+
+Install the Intel SGX software stack for Linux (including the SGX driver, the SGX SDK, and the SGX Platform Software
+(PSW)) by following the official [documentation](https://github.com/intel/linux-sgx). Please make sure that you use the
+SDK version as denoted above in the list of requirements. 
+
+Moreover, if you don't have SGX hardware available you can also install the SGX SDK only and use simulation mode by
+setting `SGX_MODE=SIM` in your environment. In this case, also make sure that simulation mode is set when building
+and installing [SGX SSL](https://github.com/intel/intel-sgx-ssl#available-make-flags). Note that the simulation mode is
+for developing purpose only and does not provide any security guarantees. 
+
+Once you have installed the SGX SDK and SSL for SGX SDK please double check that ``SGX_SDK`` and ``SGX_SSL`` variables
+are set correctly in your environment. 
+
+Notice: by default the project builds in hardware-mode SGX, ``SGX_MODE=HW`` as defined in `<absolute-project-path>/fabric-private-chaincode/config.mk` and you can
+explicitly opt for building in simulation-mode SGX, ``SGX_MODE=SIM``. In order to set non-default values for install
+location, or for building in simulation-mode SGX, you can create the file `<absolute-project-path>/fabric-private-chaincode/config.override.mk` and override the default
+values by defining the corresponding environment variable.
+
+##### Register with Intel Attestation Service (IAS) if using SGX_MODE=HW
+
+We use Intel's Attestation Service to perform attestation with chaincode enclaves. If you run SGX in __simulation mode__
+only, you can skip this section and come back when you want setup with SGX hardware-mode.
+ 
+What you need:
+
+* a Service Provider ID (SPID)
+* the (primary) api-key associated with your SPID
+
+In order to use Intel's Attestation Service (IAS), you need to register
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -234,8 +230,7 @@ explicitly opt for building in simulation-mode SGX, ``SGX_MODE=SIM``. In order t
 location, or for building in simulation-mode SGX, you can create the file `config.override.mk` and override the default
 values by defining the corresponding environment variable.
 
-
-##### Intel Attestation Service (IAS)
+#### Intel Attestation Service (IAS)
```
g2flyer: splitting the section in the two options docker vs native install is a good idea. However, as the IAS setup applies to both settings you have to move this to a different place, probably best just before the two options (and with one '#' less)

README.md
```diff
@@ -293,25 +286,22 @@ Make sure Fabric is in your ``$GOPATH`` and you enable the plugin feature using
...
-
-    $ GO_TAGS=pluginsenabled make peer
-
+```
+$ GO_TAGS=pluginsenabled make peer
+```
 Please make sure that the peer is _always_ built with GO_TAGS, otherwise our custom validation plugins will (silently!)
 ignored by the peer, despite the settings in ``core.yaml``.
 
+#### Build the entire project
```
g2flyer: this is also not on the level of option 2 but does apply to both options, so should be one level up? Or maybe add new third-level section like Developping in FPC or alike with this and below as sub-sections (on same level as before). In that case you could potentially also move the IAS section here as first section or so

The "not restricted to option 2" of course also applies to the following section up to the "Developping your first private chaincode"

README.md
```diff
@@ -263,16 +258,14 @@ Place your ias api key and your SPID in the ``ias`` folder as follows:
     echo 'YOUR_API_KEY' > ${GOPATH}/src/github.com/hyperledger-labs/fabric-private-chaincode/config/ias/api_key.txt
     echo 'YOUR_SPID' > ${GOPATH}/src/github.com/hyperledger-labs/fabric-private-chaincode/config/ias/spid.txt
 
-
-### Fabric Private Chaincode
+#### Fabric Private Chaincode
 
 Clone the code and make sure it is on your `$GOPATH`. (Important: we assume in this documentation and default
```
mbrandenburger: I think here we could add a hint that the docker container already has the code and this step is not needed in tthat case

README.md
g2flyer: nope, options 1 and two are on level ### and SDK & SSL are both at level #### under option 2. Also why should IAS be a subsetion of SDK & SSL?  Having SDK & SSL under option 2 makes sense as for option 1 it is pre-installed. But IAS credentials you need in both cases

README.md
```diff
@@ -111,48 +110,18 @@ your [development environment](https://hyperledger-fabric.readthedocs.io/en/rele
...
-
-* Hyperledger [Fabric](https://github.com/hyperledger/fabric) v1.4.3
-
-* Clang-format 6.x or higher
-
-* [PlantUML](http://plantuml.com/) including Graphviz (for building documentation only) 
-
-
-#### Docker
+### **Option 1: Using the Docker-based FPC Development Environment**
```
g2flyer: why are we making titles also bold? i think we should rely on the formatting of headers by markdown.  That said, the headers do not distinguish by much. But in that case, we might more consider whether we couldn't flatten the hiearchacy?  Or is there a way to give (automatic) section numbers in markdown? that also would make it more clear ..

README.md
annumberhocker: Moved SGX and IAS sections to top.

README.md
annumberhocker: If GO_TAGS is required, can't it be automatically set in the docker development environment?  All the other required ENV variables are set.

README.md
annumberhocker: We used bold to set these headers apart at this level.  The hierarchy is getting too deep otherwise and the headers start look smaller than the body text.

README.md
g2flyer: I have to admit i don't know the exact semantics of GO_TAGS. Is this a Fabric specific variable or something generic go? In either case, it could potentially negatively affect our own FPC go builds?  So i would be a bit reluctant to that to the docker image environment

README.md
g2flyer: hmm, but this would rather ask for a change of hierachy rather than change (some) headers of a certain level from their standard formatting?

g2flyer: an alternative could of course be to maybe move option 2 to a separate document with the assumption many folks would start right away with docker and wouldn't care "bare metal" (and others might just replicate the commands from the docker file rather than read the instructions :-)

README.md
```diff
@@ -110,9 +110,64 @@ your [development environment](https://hyperledger-fabric.readthedocs.io/en/rele
 
 Moreover, we assume that you are familiar with the Intel SGX SDK.
 
+### **Initial Setup**
+
+#### 1. Intel SGX SDK and SSL
```
g2flyer: With option 2 you wouldn't need SDK or SGXSSL installation and PSW only (on host) if you want to run in HW mode.    So ideally we would talk here only about SGX & PSW and mention SGX_MODE but would outsource SGX SDK and SGXSSL to subsection in option 2?

PR #241, opened 2020-03-10
fix README.md

lengyijun:
Signed-off-by: lengyijun <sjtu5140809011@gmail.com>


mbrandenburger:
Thanks @lengyijun for your PR!

@g2flyer Seems that the avatar service breaks our build .... Not good ... can we remove this from the critical path again?

g2flyer:
> Thanks @lengyijun for your PR!
> 
> @g2flyer Seems that the avatar service breaks our build .... Not good ... can we remove this from the critical path again?

In general it does work, though?  (I've pushed quite a few PR (updates) since then?  I guess if it happens too often, we could make the makefile simply ignore any related error (with a '-' prefix ..)

PR #257, opened 2020-03-14
Rebase to latest master ...

g2flyer:


mbrandenburger:
@g2flyer thanks for rebasing! please have a look and check why DCO check failed. 

We can ignore travis fail for 0e1637f I think but DCO we cannot :/

g2flyer:
> @g2flyer thanks for rebasing! please have a look and check why DCO check failed.
> 
> We can ignore travis fail for [0e1637f](https://github.com/hyperledger-labs/fabric-private-chaincode/commit/0e1637f6418205cc798c03e733e1dd4e14bb869f) I think but DCO we cannot :/

Well, my PR was only a rebase with no new commit, so we must have failed already.  Looking at the error messages there are a lot of them where github just complains some email addresses do not correspond 1-1 with github ids.  I think these we definitely should ignore.  There is also (one commit from Bruno from last september)[https://github.com/hyperledger-labs/fabric-private-chaincode/pull/257/commits/9583608e8e50fa438b251b831f94e35ae2b77ab5] which lacks a DCO completely but assuming @bvavala doesn't repudiate that commit, i also would leave it and just override DCO. Rewriting the complete history of master for this doesn't seem worth it.

mbrandenburger:
what are our options here to resolve?

1. force merge this

1. move the new architecture documents to master and keep it in a folder that clearly shows that this is the design we aim for and does not necessary describe the current code base 

1. cherry-pick our new design to a fresh branch based on current master and drop the old one

My feeling is option 2 is the way to go to avoid any additional need to keep that branch updated.

@bvavala @MHBauer any suggestions?

As we also discussed ("offline") the use of feature branches, I think they only work when updated and carefully maintained regularly.  In particular, a feature branch for a milestone must be merged into master once the milestone is done. In other words, feature branches should be "short living" whereas this branch is an example of a long living branch.

g2flyer:
> As we also discussed ("offline") the use of feature branches, I think they only work when updated and carefully maintained regularly. In particular, a feature branch for a milestone must be merged into master once the milestone is done. In other words, feature branches should be "short living" whereas this branch is an example of a long living branch.

The issue here seems more how with github do you maintain shared branches like new_design or v2_in-progress-or-what-ever we call that.  We need a good process for rebasing/merging these branches on github (in both directions). So while i'm personally fine with 2., unless we also do the v2-work-in-progress-not-everything-works-yet also in master, we have to sort out the feature branch question. Once we have that, there is also not necessarily any need anymore to move new_design to master (but we might put it just into the v2-wip-feature-branch or, even simpler, just declare the new_design branch to be the v2-wip-feature-branch?).

g2flyer:
closing/cancelling this PR as we decided to merge it to master and take it from there ....

PR #262, opened 2020-03-17
Disclaimer and explicit list of "official" releases

g2flyer:


README.md
```diff
@@ -97,6 +97,16 @@ The system consists of the following components:
    ledger enclave, which crosschecks the decision before it finally commits
    the transaction to the ledger.
 
+## Releases
```
bvavala: This should probably be the first thing to show at the top.

g2flyer: this was my initial thought but then (a) i thought it would look ugly to start right away with a disclaimer (this was, though, before i looked at the fabric readme, saw and borrowed the release section) and (b) thought it would be also close to where you start playing around.

mbrandenburger:
Thanks @g2flyer

PR #266, opened 2020-03-19
proxy-patch for fabric v2 peer

g2flyer:
This should enable proxy support for the v2 peer and also re-enable patching in docker and travis

Should be only temporary necessary until externalbuilder support is there.


mbrandenburger:
@g2flyer this needs a rebase patch must be applied "again"

g2flyer:
btw: what is also annoying is that each clean build seems to touch both go.mod and go.sum!  I think  the problems are the // indirect links in go.mod. We probably should do a go mod tidy before checking in go.mod when it contains indirect references?

g2flyer:
Note the way we build the dev docker image right now, it is build from master and not from the pr, so travis will fail right now, but should work once this pr is merged.  (The underlying problem is of course the same as the one in #269 ...)

mbrandenburger:
Travis build failed:
```
error: patch failed: core/chaincode/chaincode_support.go:8
error: core/chaincode/chaincode_support.go: patch does not apply
Applying: proxy-enable chaincode containers if necessary
Patch failed at 0001 proxy-enable chaincode containers if necessary
Use 'git am --show-current-patch' to see the failed patch
```

bvavala:
> LGTM! I verified that patch can be applied on Fabric v2.0.1 release tag - cannot perform functional test as not sitting behind a proxy. @bvavala please do you review and we are good to go. Thanks!

Travis failed. Is it expected?

g2flyer:
> Travis failed. Is it expected?

yes, see my comment two comments above yours ....

g2flyer:
One other note: you might have to manually pull the image `hyperledger/fabric-ccenv:2.0`.  the underlying go-dockerclient doesn't seem to support proxy when pulling images and if you do not have that image ledger_init might timeout ....

Issue #251, opened 2020-03-13
New Tags for better issue structuring


mbrandenburger:
As discussed today we should carefully use the tag system in github to organize our issues better.

We could have the following feature tags:

- Test
- Documentation (I guess we already have that one)
- 

We could have the following component tags:

- ERCC
- TLCC
- FPC Shim
- FPC Validation
- FPC Examples
- Client SDK

WDYT? Please discuss

g2flyer:
For features i was more initially thinking of component cross-cutting stuff like key-management or alike, but right now i'm actually also not getting more for that list and that item alone doesn't deserve a separate category.

We should though have a tag capturing priority, in particular i think we need a label for
- MVP (minimal viable product / an issue we must resolve before we can get into fabric)

Regarding components, there is probably also one for "infrastructure" or alike which captures the network scripts and alike?  There is also "Demo" (but that tag already exists)

Lastly, i think it would be good if we can distinguish the different catogories:  If we can color-code them, that's probably the easiest.  Otherwise we could prefix them with something like 'c:' and 'f:' ?

g2flyer:
After last discussion, we decided that priority as far as MVP goes is done via a milestone. I've created now a corresponding [MVP milestone](https://github.com/hyperledger-labs/fabric-private-chaincode/milestone/38), a `feature` label, created issues from the MVP list in https://docs.google.com/document/d/1ztdJykj4I-e9G-YGGnn2ExDVCQQHmbuhgdaA5HWkd9M and assigned the MVP version of it to the milestone ....

g2flyer:
@mbrandenburger will you create the labels? See also the PR #302 for potential templates we could use in this context. Depending on how you create labels, you might want update the templates?

PR #302, opened 2020-03-20
Templates for github issues

g2flyer:
Starting point for issue templates for issue #251 ...

### DIFF
--- a/.github/ISSUE_TEMPLATE/work-item-template.md
+++ b/.github/ISSUE_TEMPLATE/work-item-template.md
@@ -1,0 +1,21 @@
+---
+name: Work item template
+about: This issue describes a work-item ....
+title: ''
+labels: ''
+assignees: ''
+
+---
+*TODO:*
+- *label with affected components*
+- *replace below italics text with corresponding info for this new issue*
+
+**Description**
+
+*Add a description of the work-itemn. Motivate briefly why we need this work item, what it will change, and what it needs to finish it.*
+
+**Link to feature or bug**
+*Please add link to any applicable feature or bug.*
+
+**Link to dependent issues**
+*Please link to any other work-item this issue depends on.*
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Log-filesScreenshots**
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/work-item-template.md
```diff
@@ -0,0 +1,12 @@
...
+name: Work item template
+about: This issue describes a work-item ....
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+- describe purpose of work-items
+- label issue with affected components
```
mbrandenburger: maybe also add
- link to a feature or bug (if applicable)

mbrandenburger:
```suggestion
**List affected components**
Give an overview what components are affected by this work-item. For instance, this work-item around attestation may touch ERCC and ECC components.
**Link to feature or bug**
Please link this work-item to a feature or bug (if applicable).
```

.github/ISSUE_TEMPLATE/work-item-template.md
```diff
@@ -0,0 +1,12 @@
...
+- list (or, better, link) any dependencies ..
```
mbrandenburger:
```suggestion
**List any dependencies**
List all other work-items, features, or bugs that must be resolved before this work-item can be finished.
```

.github/ISSUE_TEMPLATE/work-item-template.md
```diff
@@ -0,0 +1,12 @@
+---
+name: Work item template
+about: This issue describes a work-item ....
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+- describe purpose of work-items
```
mbrandenburger:
```suggestion
**Describe purpose of the work-item**
Motivate briefly why we need this work item, what it will change, and what it needs to finish it.
```

.github/ISSUE_TEMPLATE/work-item-template.md
g2flyer: makes sense, will add ...

.github/ISSUE_TEMPLATE/work-item-template.md
g2flyer: it seems you want sections rather than lists?

.github/ISSUE_TEMPLATE/work-item-template.md
g2flyer: hmm, if we have component labels according labels (#251) do we really have to list them? Seems a bit duplicate and might lead folks _not_ to label (yet label is kind of nice as you can identify set of issues touching component which you cannot do from label).

mbrandenburger: you are right - what I suggested here is redundant with github tags.

So you want here a grouping label to ... group a set of work-items? maybe a feature or a bug can be this kind of grouping element?

g2flyer: Look at the revised template: i added a TODO header which is supposed to tell the issue creator what to do.   as you can add as many labels as you want you do not really need a grouping element here?  But having labels in a way that same-type-labels-look-similar-and-distinct-from-others would seem useful, hence my comment in #251 to either group by color or by a common prefix?

Issue #247, opened 2020-03-12
Update go chaincode imports


mbrandenburger:
With Fabric 2.0 the go chaincode shim moved to a separate repo in `github.com/hyperledger/fabric-chaincode-go/shim`. Also protos have been moved to `github.com/hyperledger/fabric-protos-go/peer`. 

Must be also added to godeps and probably also move to use of go modules as done in https://github.com/hyperledger/fabric-samples/tree/master/chaincode/marbles02/go ?

Needs to be updated in the following components:
- [ ] ERCC 
- [ ] ERCC-VSCC
- [ ] TLCC
- [ ] ECC
- [ ] ECC-VSCC
- [ ] demo golang CC ...

mbrandenburger:
Note that MockStub moved to shimtest package, see [here](https://github.com/hyperledger/fabric-chaincode-go/blob/master/shimtest/mockstub.go). This is relevant for `utils/test_helper.go` and in mock Server.

g2flyer:
Related to the go modules, see also #94

PR #304, opened 2020-03-21
init mod from 'go get' calls

### LINKED ISSUE #94: go dep management
the current code base has some go dependencies and require a user to fetch them manually. As a workaround we use our make system to install these deps. However, this approach may lead to inconsistent versioning and unexpected errors.

MHBauer:
to recreate:
`git clone` a fresh checkout,, not in gopath
`make godeps` to download all deps as they currently exist
`go mod init github.com/hyperledger-labs/fabric-private-chaincode` in base directory.

`go mod download` can be used to initialize the go build cache in a similar manner to the existing `go get` calls.

As it existed the `go get` commands in the makefile have the potential
to float and not get any specific version. The newer go toolchains
that are module enabled support giving version specifiers to `go get`,
for example `go get github.com/spf13/viper@v1.6.2`. Another behavior
of module enabled toolchain, is that when in a module, `go get` will
write a dependency to the `go.mod` file. Combining these two facts, if
we enable modules, and then run the existing `go get` commands, we
will end up with the dependencies currently necessary to build. After
which, we use `go mod download` to download all modules in the
`go.mod` file. This change replaces all `go get` invocations with a
single invocation of `go mod download`.

There are probably actual used dependencies (retrieved when compiling
the code) that are still missing.

resolves #94

@mbrandenburger  can you try the ecc failure with this patch?


mbrandenburger:
This seems to help with build ecc-vscc.

MHBauer:
should I re-enable it in another commit or do you want to build on top of this?

mbrandenburger:
> should I re-enable it in another commit or do you want to build on top of this?

I suggest to make a separate PR for plugins.

mbrandenburger:
@MHBauer could you please link this PR to the corresponding issue and briefly explain what is happening here.

MHBauer:
@g2flyer 
Maybe not necessary to not be in the gopath. Unsure, Wrote down and communicated what I did do. Let's discuss the fabric build later. Remind me.
1. Yes. Let us do it separately, as part of #247. Demo probably should also be a separate module.
2. I thought it is done? We can't do it for the tools as those are binary targets. (Another go thing, if `go get` resolves to a binary, do the download, but also build and install it into `go/bin`)
3. I didn't do any builds after this, it's difficult for me to know what exists in terms of output artifacts. `go build` will modify go.mod and go sum every time if there's any changes needed. Someone should do a build and add a commit here with the changes after successful builds.
4. No threads here.

mbrandenburger:
> Note also that tidy revealed that we forgot to fix one of our chaincodes: `demo/client/backend/mock/chaincode/fpc_chaincode.go` still references the old v1.4 shim ...

I suggest that we add a TODO to bring the Demo up and running again.

g2flyer:
> Maybe not necessary to not be in the gopath.  Unsure
See notes under go 1.13 on https://github.com/golang/go/wiki/Modules ....

> 1. Yes. Let us do it separately, as part of #247. Demo probably should also be a separate module.
Makes sense (BTW: i've ticked off now ERCC in 247 as this should be now covered?)

> 2. I thought it is done? We can't do it for the tools as those are binary targets. (Another go thing, if `go get` resolves to a binary, do the download, but also build and install it into `go/bin`)
I missed adding the crucial words "in the `dev` and `peer` `Dockerfiles`". Given that marcus already merged, i guess it needs a separate PR.

> 3. I didn't do any builds after this, it's difficult for me to know what exists in terms of output artifacts. `go build` will modify go.mod and go sum every time if there's any changes needed. Someone should do a build and add a commit here with the changes after successful builds.
Ok, will do that as part of above PR.  You haven't though answered part of my question: should we do each time a `go mod tidy` or does that cause some issues?

g2flyer:
> > Note also that tidy revealed that we forgot to fix one of our chaincodes: `demo/client/backend/mock/chaincode/fpc_chaincode.go` still references the old v1.4 shim ...
> 
> I suggest that we add a TODO to bring the Demo up and running again.

@mbrandenburger I guess you mean (Work-item) Issue when you say TODO?  Currently some go commands fail now due to that. I will fix that as part of the PR i mentioned in reply to Morgan.  Can you create the work-item? (I've created so many lately, so more diversity in the creators might not hurt :-)

g2flyer:
FYI: created PR #308 for the pr mentioned above.  And as issue-creator-in-chief-lately i've also created an issue #309 for the demo ...

MHBauer:
@g2flyer as long as everything builds. `go mod tidy` shouldn't hurt anything.

Issue #254, opened 2020-03-13
Upgrade integration-test ledger scripts and ledger configuration to v2.0


g2flyer:
- scripts to change:  `fabric/bin/lib/*`, `fabric/bin/*`, `integration/*sh`
- config to upgrade: `integration/config`

For scripts, it means moving from `peer chaincode` to new `peer lifecycle chaincode` commands in `integration/*sh`, making sure v2 binaries and config is used in `fabric/bin/lib/*` and adding wrappers for these new lifecycle cli calls to `fabric/bin/peer.sh`. The new wrappers have to essentially just to replicate the semantics of the old `install` and `instantiate` wrappers (i.e., rename of cc with prefix fpc_ if it is fpc cc, create docker images for fpc cc and call __setup, …)

For config, this essentially means a three-way diff  of vanialla 1.4.3 config from byfn in fabric-samples on which we based `integration/config' with our version (changes should hopefully be all marked as FPC changes in our version now ...) and test-network configs from 2.0.1 fabric-samples.

Depends on (at least) resolution of #255 (and optionally #246), #247.

Probably best initial step is to do #255 for ercc (which shouldn't have any FPC dependencies?) and test that on 2.0 fabric-samples test-network, then tackle this

g2flyer:
Some notes from investigation fabric v2 regarding the chaincode container "switcheroo" we do for FPC chaincode (i.e., we create the container ourself with the correct naming and rely on the peer not re-creating it when it already exists):
- the peer still creates the container only if it does not exist
- the peer also still uses the same construction of naming containers as codified in `utils/fabric/get-fabric-container-name.go`.
- however, with the new lifecycle, there is no version provided anymore with `peer lifecycle package` and `peer lifecycle chaincode install` (the version used in CC definition materializes only in `peer lifecycle approveformyorg').  Yet the container will be created during `install`, so there must the version used there is not the one from the chaincode definition. Turns out that (a) the version used is the packageid and (b) the package-id for package `foo.tar.gz` seems to be `sha256sum foo.tar.gz`

Conclusion: switcheroo in `peer.sh` should still work with new lifecycle but we have to derive the version now as the hash of package we create and install rather than based on the version provided.

g2flyer:
> Conclusion: switcheroo in `peer.sh` should still work with new lifecycle but we have to derive the version now as the hash of package we create and install rather than based on the version provided.

[...]

PR #307, opened 2020-03-22
Modified intergration/config files to fabric 2.0

Jordy24:
Config files have been updated to fabric 2.0. However FPC Additions from 1 to 5 have been disabled (commented out)

TODO:
enable decorator , validation plugins and tlcc

Purpose of this PR:
This PR contains changes that make fabric/bin/ledger_init.sh and fabric/bin/ledger_shutdown.sh work with the new lifecycle. 
In this PR ercc.go chaincode is installed using the new fabric 2.0 lifecycle.

To test use:
env FABRIC_LOGGING_SPEC=debug FABRIC_CFG_PATH=$(/bin/pwd)/integration/config/ ./fabric/bin/ledger_init.sh

env FABRIC_LOGGING_SPEC=debug FABRIC_CFG_PATH=$(/bin/pwd)/integration/config/ ./fabric/bin/ledger_shutdown.sh

Signed-off-by: Jordan Karaze <jordan.karaze@ibm.com>

 

This PR addresses some of the components of PR #254 by upgrading config files
```

### DIFF
--- a/fabric/bin/lib/common_ledger.sh
+++ b/fabric/bin/lib/common_ledger.sh
@@ -48,6 +48,7 @@ define_common_vars() {
     : ${CHAN_ID:="mychannel"}
     : ${ERCC_ID:="ercc"}
     : ${ERCC_VERSION:="0"}
+    : ${ERCC_SEQUENCE:="1"}
 
     ORDERER_PID_FILE="${FABRIC_STATE_DIR}/orderer.pid"
     ORDERER_LOG_OUT="${FABRIC_STATE_DIR}/orderer.out"
@@ -112,6 +113,9 @@ ledger_init() {
     try rm -rf ${FABRIC_STATE_DIR}/*
 
     # 2. start orderer
+    # - Creating a genesisblock for orderer-system-channel......."
+    try ${CONFIGTXGEN_CMD} -outputBlock ${FABRIC_STATE_DIR}/genesisblock -profile SampleDevModeSolo -channelID orderer-system-channel
+    sleep 1
     ORDERER_GENERAL_GENESISPROFILE=SampleDevModeSolo ${ORDERER_CMD} 1>${ORDERER_LOG_OUT} 2>${ORDERER_LOG_ERR} &
     export ORDERER_PID=$!
     echo "${ORDERER_PID}" > ${ORDERER_PID_FILE}
--- a/fabric/bin/peer.sh
+++ b/fabric/bin/peer.sh
@@ -331,25 +331,45 @@ handle_channel_join() {
     try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
 
     # - handle ercc
-    say "Installing & Instantiating ercc on channel '${CHAN_ID}' ..."
-    #   - install ercc
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode install -n ${ERCC_ID} -v ${ERCC_VERSION} -p github.com/hyperledger-labs/fabric-private-chaincode/ercc/cmd
-    sleep 1
-    #   - instantiate ercc iff "channel creation" peer
-    if [ -e "${FABRIC_STATE_DIR}/${CHAN_ID}.creator" ]; then
-	try $RUN ${FABRIC_BIN_DIR}/peer chaincode instantiate -n ${ERCC_ID} -v ${ERCC_VERSION} -c '{"args":["init"]}' -C ${CHAN_ID} -V ercc-vscc
-	sleep 4
-	try rm "${FABRIC_STATE_DIR}/${CHAN_ID}.creator"
-    fi
-    #   - get SPID (mostly as debug output)
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n ${ERCC_ID} -c '{"args":["getSPID"]}' -C ${CHAN_ID}
+    say "Installing ercc on channel '${CHAN_ID}' ..."
+    say "Packaging chaincode ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode package ${FABRIC_STATE_DIR}/ercc.tar.gz -p ../../ercc --lang golang --label erccv1 
+    para
+    sleep 3
+    say "In installing chaincode ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode install ${FABRIC_STATE_DIR}/ercc.tar.gz 
+    para
+    sleep 3
+    say "Query installed chaincodes.."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode queryinstalled >&log.txt
+    para
+    PACKAGE_ID=$(awk '/Package ID:/{print}' log.txt | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
+    say "Approve for my org"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode approveformyorg -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ${ERCC_ID} --version ${ERCC_VERSION} --package-id ${PACKAGE_ID} --sequence ${ERCC_SEQUENCE} # -V ercc-vscc   # TODO: re-add validation plugin once they are enabled in peer ... 
+    para
+    sleep 3
+    say "Check for commit readiness"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode checkcommitreadiness --channelID ${CHAN_ID} --name ${ERCC_ID} --version ${ERCC_VERSION} --sequence ${ERCC_SEQUENCE} --output json # -V ercc-vscc   # TODO: re-add validation plugin once they are enabled in peer ...
+    para
+    sleep 3
+    say "Commit chaincode definition...."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode commit -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ${ERCC_ID} --version ${ERCC_VERSION} --sequence ${ERCC_SEQUENCE} # -V ercc-vscc   # TODO: re-add validation plugin once they are enabled in peer ...
+    para
+    sleep 3
+    say "Query commited chaincodes on the channel"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} 
+    para
+    sleep 3
+    say "call chaincode invoke..."
+    try $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke -n ${ERCC_ID} -c '{"function":"getSPID","args":[]}' -C ${CHAN_ID} 
     sleep 3
 
-    # - ask tlcc to join channel
-    #   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
-    #   it will re-join old channels but tlcc will not!
-    say "Attaching TLCC to channel '${CHAN_ID}' ..."
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID}
+
+    ## - ask tlcc to join channel
+    ##   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
+    ##   it will re-join old channels but tlcc will not!
+    #say "Attaching TLCC to channel '${CHAN_ID}' ..." # TODO: Enable once tlcc is integrated.....
+    #try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID} # TODO: Enable once tlcc is integrated
 
     # - exit
     exit 0
--- a/integration/config/configtx.yaml
+++ b/integration/config/configtx.yaml
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-
+# Based on hyperledger/fabric/sampleconfig
 ---
 ################################################################################
 #
@@ -22,6 +22,14 @@ Organizations:
         # configuration transactions.
         # Name can include alphanumeric characters as well as dots and dashes.
         Name: SampleOrg
+
+        # SkipAsForeign can be set to true for org definitions which are to be
+        # inherited from the orderer system channel during channel creation.  This
+        # is especially useful when an admin of a single org without access to the
+        # MSP directories of the other orgs wishes to create a channel.  Note
+        # this property must always be set to false for orgs included in block
+        # creation.
+        SkipAsForeign: false
 
         # ID is the key by which this org's MSP definition will be referenced.
         # ID can include alphanumeric characters as well as dots and dashes.
@@ -49,6 +57,9 @@ Organizations:
             Admins:
                 Type: Signature
                 Rule: "OR('SampleOrg.admin')"
+            Endorsement:
+                Type: Signature
+                Rule: "OR('SampleOrg.member')"
 
         # OrdererEndpoints is a list of all orderers this org runs which clients
         # and peers may to connect to to push transactions and receive blocks respectively.
@@ -86,52 +97,34 @@ Capabilities:
     # supported by both.
     # Set the value of the capability to true to require it.
     Channel: &ChannelCapabilities
-        # V1.4.3 for Channel is a catchall flag for behavior which has been
-        # determined to be desired for all orderers and peers running at the v1.4.3
+        # V2.0 for Channel is a catchall flag for behavior which has been
+        # determined to be desired for all orderers and peers running at the v2.0.0
         # level, but which would be incompatible with orderers and peers from
         # prior releases.
-        # Prior to enabling V1.4.3 channel capabilities, ensure that all
-        # orderers and peers on a channel are at v1.4.3 or later.
-        V1_4_3: true
-        # V1.3 for Channel enables the new non-backwards compatible
-        # features and fixes of fabric v1.3
-        V1_3: false
-        # V1.1 for Channel enables the new non-backwards compatible
-        # features and fixes of fabric v1.1
-        V1_1: false
+        # Prior to enabling V2.0 channel capabilities, ensure that all
+        # orderers and peers on a channel are at v2.0.0 or later.
+        V2_0: true
 
     # Orderer capabilities apply only to the orderers, and may be safely
     # used with prior release peers.
     # Set the value of the capability to true to require it.
     Orderer: &OrdererCapabilities
-        # V1.4.2 for Orderer is a catchall flag for behavior which has been
-        # determined to be desired for all orderers running at the v1.4.2
+        # V1.1 for Orderer is a catchall flag for behavior which has been
+        # determined to be desired for all orderers running at the v1.1.x
         # level, but which would be incompatible with orderers from prior releases.
-        # Prior to enabling V1.4.2 orderer capabilities, ensure that all
-        # orderers on a channel are at v1.4.2 or later.
-        V1_4_2: true
-        # V1.1 for Orderer enables the new non-backwards compatible
-        # features and fixes of fabric v1.1
-        V1_1: false
+        # Prior to enabling V2.0 orderer capabilities, ensure that all
+        # orderers on a channel are at v2.0.0 or later.
+        V2_0: true
 
     # Application capabilities apply only to the peer network, and may be safely
     # used with prior release orderers.
     # Set the value of the capability to true to require it.
     Application: &ApplicationCapabilities
-        # V1.4.2 for Application enables the new non-backwards compatible
-        # features and fixes of fabric v1.4.2
-        V1_4_2: true
-        # V1.3 for Application enables the new non-backwards compatible
-        # features and fixes of fabric v1.3.
-        V1_3: false
-        # V1.2 for Application enables the new non-backwards compatible
-        # features and fixes of fabric v1.2 (note, this need not be set if
-        # later version capabilities are set)
-        V1_2: false
-        # V1.1 for Application enables the new non-backwards compatible
-        # features and fixes of fabric v1.1 (note, this need not be set if
-        # later version capabilities are set).
-        V1_1: false
+        # V2.0 for Application enables the new non-backwards compatible
+        # features and fixes of fabric v2.0.
+        # Prior to enabling V2.0 orderer capabilities, ensure that all
+        # orderers on a channel are at v2.0.0 or later.
+        V2_0: true
 
 ################################################################################
 #
@@ -149,8 +142,22 @@ Application: &ApplicationDefaults
         # (e.g.,who can receive Block events). This section does NOT specify the resource's
         # definition or API, but just the ACL policy for it.
         #
-        # User's can override these defaults with their own policy mapping by defining the
+        # Users can override these defaults with their own policy mapping by defining the
         # mapping under ACLs in their channel definition
+
+        #---New Lifecycle System Chaincode (_lifecycle) function to policy mapping for access control--#
+
+        # ACL policy for _lifecycle's "CheckCommitReadiness" function
+        _lifecycle/CheckCommitReadiness: /Channel/Application/Writers
+
+        # ACL policy for _lifecycle's "CommitChaincodeDefinition" function
+        _lifecycle/CommitChaincodeDefinition: /Channel/Application/Writers
+
+        # ACL policy for _lifecycle's "QueryChaincodeDefinition" function
+        _lifecycle/QueryChaincodeDefinition: /Channel/Application/Readers
+
+        # ACL policy for _lifecycle's "QueryChaincodeDefinitions" function
+        _lifecycle/QueryChaincodeDefinitions: /Channel/Application/Readers
 
         #---Lifecycle System Chaincode (lscc) function to policy mapping for access control---#
 
@@ -218,6 +225,12 @@ Application: &ApplicationDefaults
     # For Application policies, their canonical path is
     #   /Channel/Application/<PolicyName>
     Policies: &ApplicationDefaultPolicies
+        LifecycleEndorsement:
+            Type: ImplicitMeta
+            Rule: "MAJORITY Endorsement"
+        Endorsement:
+            Type: ImplicitMeta
+            Rule: "MAJORITY Endorsement"
         Readers:
             Type: ImplicitMeta
             Rule: "ANY Readers"
@@ -351,7 +364,7 @@ Orderer: &OrdererDefaults
             MaxInflightBlocks: 5
 
             # SnapshotIntervalSize defines number of bytes per which a snapshot is taken
-            SnapshotIntervalSize: 20 MB
+            SnapshotIntervalSize: 16 MB
 
     # Organizations lists the orgs participating on the orderer side of the
     # network.
@@ -553,11 +566,12 @@ Profiles:
     # profiles, only the 'Application' section and consortium # name are
     # considered.
     SampleSingleMSPChannel:
+        <<: *ChannelDefaults
         Consortium: SampleConsortium
         Application:
             <<: *ApplicationDefaults
             Organizations:
-                - *SampleOrg
+                - <<: *SampleOrg
 
     # SampleDevModeEtcdRaft defines a configuration that differs from the
     # SampleDevModeSolo one only in that it uses the etcd/raft-based orderer.
--- a/integration/config/core.yaml
+++ b/integration/config/core.yaml
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
+# Based on hyperledger/fabric/sampleconfig
 
 ###############################################################################
 #
@@ -11,15 +12,18 @@
 ###############################################################################
 peer:
 
-    # The Peer id is used for identifying this Peer instance.
+    # The peer id provides a name for this peer instance and is used when
+    # naming docker resources.
     id: jdoe_test
 
-    # The networkId allows for logical seperation of networks
+    # The networkId allows for logical separation of networks and is used when
+    # naming docker resources.
     networkId: dev_test
 
     # The Address at local network interface this Peer will listen on.
     # By default, it will listen on all network interfaces
-    listenAddress: 0.0.0.0:7051
+    # FPC change: the original 0.0.0.0 didn't always work in some settings and we need it only for localhost
+    listenAddress: 127.0.0.1:7051
 
     # The endpoint this peer uses to listen for inbound chaincode connections.
     # If this is commented-out, the listen address is selected to be
@@ -36,18 +40,21 @@ peer:
     # in the same organization. For peers in other organization, see
     # gossip.externalEndpoint for more info.
     # When used as CLI config, this means the peer's endpoint to interact with
-    address: 127.0.0.1:7051 # Note: setting to 127.0.0.1 restricts to localhost but does make it work behind proxies
+    # FPC change: see comment above listenAddress above ..
+    address: 127.0.0.1:7051
 
     # Whether the Peer should programmatically determine its address
     # This case is useful for docker containers.
     addressAutoDetect: false
 
-    # Setting for runtime.GOMAXPROCS(n). If n < 1, it does not change the
-    # current setting
-    gomaxprocs: -1
-
     # Keepalive settings for peer server and clients
     keepalive:
+        # Interval is the duration after which if the server does not see
+        # any activity from the client it pings the client to see if it's alive
+        interval: 7200s
+        # Timeout is the duration the server waits for a response
+        # from the client after sending a ping before closing the connection
+        timeout: 20s
         # MinInterval is the minimum permitted time between client pings.
         # If clients send pings more frequently, the peer server will
         # disconnect them
@@ -217,18 +224,16 @@ peer:
             responseTimeout: 3s
             # batchSize the number of blocks to request via state transfer from another peer
             batchSize: 10
-            # blockBufferSize reflect the maximum distance between lowest and
-            # highest block sequence number state buffer to avoid holes.
-            # In order to ensure absence of the holes actual buffer size
-            # is twice of this distance
+            # blockBufferSize reflects the size of the re-ordering buffer
+            # which captures blocks and takes care to deliver them in order
+            # down to the ledger layer. The actually buffer size is bounded between
+            # 0 and 2*blockBufferSize, each channel maintains its own buffer
             blockBufferSize: 100
             # maxRetries maximum number of re-tries to ask
             # for single state transfer request
             maxRetries: 3
 
     # TLS Settings
-    # Note that peer-chaincode connections through chaincodeListenAddress is
-    # not mutual TLS auth. See comments on chaincodeListenAddress for more info
     tls:
         # Require server-side TLS
         enabled:  false
@@ -296,8 +301,6 @@ peer:
             Pin:
             Hash:
             Security:
-            FileKeyStore:
-                KeyStore:
 
     # Path on the file system where peer will find MSP local configurations
     mspConfigPath: msp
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: fabric/bin/lib/common_ledger.sh [hunk 1]
```diff
@@ -131,9 +148,14 @@ ledger_init() {
     try ${CONFIGTXGEN_CMD} -channelID ${CHAN_ID} -profile SampleSingleMSPChannel -outputCreateChannelTx ${CHANNEL_TX}
     # - create genesis block, only by one peer
     try ${PEER_CMD} channel create -o ${ORDERER_ADDR} -c ${CHAN_ID} -f ${CHANNEL_TX} --outputBlock ${CHANNEL_BLOCK}
+    para
+    echo "[+] Channel has been created successful...."
+    echo "[-] JOINING CHANNEL FOR NOW IS DISABLED."
```
mbrandenburger: Could you please elaborate why we don't join the channel?

integration/config/configtx.yaml
```diff
@@ -1,4 +1,3 @@
-# Copyright Intel Corp. All Rights Reserved.
```
mbrandenburger: Please keep it

integration/config/core.yaml
```diff
@@ -1,5 +1,4 @@
 # Copyright IBM Corp. All Rights Reserved.
-# Copyright Intel Corp. All Rights Reserved.
```
mbrandenburger: we should keep this

integration/config/core.yaml
```diff
@@ -458,7 +466,7 @@ vm:
 
         # Enables/disables the standard out/err from chaincode containers for
         # debugging purposes
-        attachStdout: true
+        attachStdout: false
```
mbrandenburger: Let's keep this `true` as we may need it since shim.logger has been removed and we need to handle chaincode logging manually.

integration/config/core.yaml
```diff
@@ -559,44 +576,28 @@ chaincode:
...
 
     # Logging section for the chaincode container
     logging:
       # Default level for all loggers within the chaincode container
+      #level:  info
       level:  debug
       # Override default level for the 'shim' logger
-      shim:   info
+      #shim:   warning
+      shim:  info
```
mbrandenburger: I am wondering why this is still in the config as shim.logger was removed?

integration/config/orderer.yaml
```diff
@@ -1,4 +1,3 @@
-# Copyright Intel Corp. All Rights Reserved.
```
mbrandenburger: Let's keep this

integration/echo_test.sh
```diff
@@ -56,7 +56,7 @@ say "- setup ledger"
 ledger_init
 
 say "- echo test"
-echo_test
+#echo_test
```
mbrandenburger: better add a TODO.

mbrandenburger: Actually why is echo test disabled but auction test not? Echo test fails?

fabric/bin/lib/common_ledger.sh
g2flyer: ledger_init doesn't work yet as ercc would need the decorator yet we disabled the corresponding build. Hence the decorator definition in core.yaml is commented out and here related functionality is commented out.  It probably would be nicer, thoyugh, to only disable the ercc transactions but still install it and also join the channel  ..

integration/echo_test.sh
g2flyer: none of these tests currently work, also as ecc is not yet updated. That said, i also would see let's not disable it here.  I actually forgot to mention to jordan that for ledger setup/teardown there are standalone scripts in `fabric/bin` (`ledger_init.sh` and `ledger_shutdown.sh`), so let's use to test ledger setup and teardown and leave the actual test-scripts untouched ..

integration/config/core.yaml
```diff
@@ -390,8 +390,8 @@ peer:
             name: DefaultDecorator
           -
           # FPC Addition:1
-            name: ERCCDecorator
-            library: ../../ercc/ercc-decorator.so
+          #  name: ERCCDecorator
```
g2flyer: We should mark the stuff we currently (and temporarily) commented out via a corresponding TODO to mark what we later will have to fix once the plugins and tlcc are integrated ...

integration/config/core.yaml
g2flyer: good question. The stanza is definitely still there in the `fabric-samples.git` as it is in `sampleconfig/core.yaml` of `fabric.git` ....

integration/config/orderer.yaml
```diff
@@ -77,24 +68,22 @@ General:
...
-    # initializing the orderer system channel and GenesisMethod is set to
-    # "file". Ignored if GenesisMethod is set to "provisional".
-    GenesisFile: genesisblock
+
+    # Bootstrap method: The method by which to obtain the bootstrap block
+    # system channel is specified. The option can be one of:
+    #   "file" - path to a faile containing the genesis block or config block of system channel
+    #   "none" - allows an orderer to start without a system channel configuration
+    BootstrapMethod: file
+    #BootstrapMethod: none
```
g2flyer: Why did we add this (commented) out line?  Tricked me first in thinking that we changed something from the standard fabric-samples version when we actually didn't.

BTW: we didn't do that in that past (although it would have been good :-) but maybe should do it now: Add a comment at the top of this file (as well as core.yaml and other files we derived from standard fabric files) that this is derived from `config/orderer.yaml` from `fabric-samples.git`. This should help in maintaining these files ....
In fact, only looking closer i realized we started originally from `sampleconfig` in `fabric.git` which also contained the `msp` files => we should also update `msp/config.yaml` to contain the v2 changes ...

fabric/bin/lib/common_ledger.sh
```diff
@@ -112,15 +112,32 @@ ledger_init() {
...
+    try ${CONFIGTXGEN_CMD} -outputBlock genesisblock -profile SampleDevModeSolo -channelID orderer-system-channel
+    sleep 1
+
     ORDERER_GENERAL_GENESISPROFILE=SampleDevModeSolo ${ORDERER_CMD} 1>${ORDERER_LOG_OUT} 2>${ORDERER_LOG_ERR} &
     export ORDERER_PID=$!
     echo "${ORDERER_PID}" > ${ORDERER_PID_FILE}
     sleep 1
     kill -0 ${ORDERER_PID} || die "Orderer quit too quickly: (for log see ${ORDERER_LOG_OUT} & ${ORDERER_LOG_ERR})"
 
+
```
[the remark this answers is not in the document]
Jordy24: As for my testing I guess there are needed..

fabric/bin/lib/common_ledger.sh
```diff
@@ -112,15 +112,32 @@ ledger_init() {
...
 
     # 2. start orderer
+    # try to create orderer-system-channel genesisblock
+    para
+    echo "[+] Creating a genesisblock for orderer-system-channel......."
+    para
+    try ${CONFIGTXGEN_CMD} -outputBlock genesisblock -profile SampleDevModeSolo -channelID orderer-system-channel
+    sleep 1
+
     ORDERER_GENERAL_GENESISPROFILE=SampleDevModeSolo ${ORDERER_CMD} 1>${ORDERER_LOG_OUT} 2>${ORDERER_LOG_ERR} &
```
g2flyer: here we also would have to change solo to raft (see above comment)

fabric/bin/lib/common_ledger.sh
```diff
@@ -112,15 +112,32 @@ ledger_init() {
     try rm -rf ${FABRIC_STATE_DIR}/*
 
     # 2. start orderer
+    # try to create orderer-system-channel genesisblock
+    para
+    echo "[+] Creating a genesisblock for orderer-system-channel......."
+    para
+    try ${CONFIGTXGEN_CMD} -outputBlock genesisblock -profile SampleDevModeSolo -channelID orderer-system-channel
```
g2flyer: Given that solo and kafka is now depricated, we probably should move to Raft (and hence probably the SampleDevModeEtcdRaft profile). From what i understood from Swetha Raft seems also faster that solo, so might be win-win. @MHBauer probably can advise whether just changing the genesis profile is enough and/or whether there are twists with raft which maybe should keep us stick with solo for the time being ...

fabric/bin/lib/common_ledger.sh
```diff
@@ -112,15 +112,32 @@ ledger_init() {
...
     sleep 1
     kill -0 ${ORDERER_PID} || die "Orderer quit too quickly: (for log see ${ORDERER_LOG_OUT} & ${ORDERER_LOG_ERR})"
 
+
+
+
     # 3. start peer
-    LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+"$LD_LIBRARY_PATH:"}${FPC_TOP_DIR}/tlcc/enclave/lib \
-		   ${PEER_CMD} node start 1>${PEER_LOG_OUT} 2>${PEER_LOG_ERR} &
+    para
```
g2flyer: reporting as output that not everything is enabled is a good idea. I would though also add a TODO comment

fabric/bin/lib/common_ledger.sh
```diff
@@ -112,15 +112,32 @@ ledger_init() {
...
+    echo "[-] NO LIBRARY IS BEING LOADED CURRENTLY....."
+    para
+    echo "[INFO] Just starting peer node directly.."
+    para
+    #LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+"$LD_LIBRARY_PATH:"}${FPC_TOP_DIR}/tlcc/enclave/lib \
```
g2flyer: Note the LD_LIBRARY_PATH environment doesn't cause any library to be loaded, only determines where libraries are found.  So you really shouldn't have to comment out this line ..

integration/config/core.yaml
```diff
@@ -498,7 +506,8 @@ chaincode:
         name:
 
     # Generic builder environment, suitable for most chaincode types
-    builder: $(DOCKER_NS)/fabric-ccenv:latest
+    builder: $(DOCKER_NS)/fabric-ccenv:$(TWO_DIGIT_VERSION)
+    #builder: $(DOCKER_NS)/fabric-ccenv:latest
```
g2flyer: i would drop the commented out line ...

fabric/bin/lib/common_ledger.sh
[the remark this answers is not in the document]
g2flyer: _very_ unlikely that additional empty lines have changed anything. In general i agree with marcus, let's try not to add (unnecessary) white-spaces as they make git diff harders to read ...

fabric/bin/lib/common_ledger.sh
mbrandenburger: General concern - How can solo by faster than RAFT? ... taking about protocol not implementation - WIRED

g2flyer: > General concern - How can solo by faster than RAFT? ... taking about protocol not implementation - WIRED

not following ...

integration/config/core.yaml
Jordy24: So by default that option is in the new config by default

fabric/bin/lib/common_ledger.sh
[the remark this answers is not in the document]
Jordy24: Oh I see... will try to reduce white spaces

integration/config/configtx.yaml
Jordy24: sure

integration/config/core.yaml
Jordy24: sure

integration/echo_test.sh
Jordy24: So that would basically mean I would have to copy both ledger_init.sh and ledger_shutdown.sh to the integration directory....so they can access "config"

fabric/bin/lib/common_ledger.sh
Jordy24: In my testing when it is not commented out, it causes issues (breaks) .

integration/config/core.yaml
Jordy24: Yap, that's true...

fabric/bin/lib/common_ledger.sh
Jordy24: sure..

fabric/bin/lib/common_ledger.sh
g2flyer: Can you be a bit more specific with the exact nature of the problem?  And could it maybe be that your FPC_TOP_DIR still points to the old 1.4 setup?

integration/config/configtx.yaml
MHBauer: elsewhere in this file is the field to edit for enabling raft
`OrdererType: solo` to `OrdererType: etcdraft`
There seems to be a lot of extra config in here that's not needed to run.

g2flyer:
> Do these networks run and get set up in travis? Or one of the make targets?

These networks are used by the integration test scripts in `integration/`.  As integration is currently disabled (it doesn't work before #254 is fixed) it is not yet invoked in travis. But certainly should be re-enabled in the future ....

integration/config/configtx.yaml
g2flyer: The approach was to keep as close as possible to the sample config in the fabric repo so it would be easy to see what has changed and upgrade (i.e., a kind of 3-way merge  of `delta new vs old fabric config` + `delta fpc vs fabric config`.

So personally i would leave the somewhat redundant information and just make sure that when we instantiate the network in `fabric/bin/....` we make sure we select the raft profile

fabric/bin/lib/common_ledger.sh
g2flyer: Actually, testing now i realize that the issue has nothing to do with the LD_LIBRARY_PATH but that the peer dies in all cases! The reason why it seemed for you to work is because you removed the '&' and hence started the peer in the foreground. A properly running peer would actually have hang here but because it dies the script proceeds.  But obviously any subsequent peer action would fail!

fabric/bin/lib/common_ledger.sh
g2flyer: btw: seems just simply replacing `SampleDevModeSolo` by `SampleDevModeEtcdRaft` does not work as raft seems to require some additional setup with certs or alike:
```
Error on outputBlock: could not create bootstrapper: could not create channel group: could not create orderer group: cannot marshal metadata for orderer type etcdraft: cannot load client cert for consenter raft0.example.com:7050
```
....

g2flyer: figured out the problem: you didn't properly comment out the decorator by leaving the leading '-'.  Pretty ugly, though, that this causes the peer to crash in a SIGSEG for a invalidly formatted config file !!

MHBauer:
As some reference, here is work @swetharepakula did to go from a 1.4 based setup with solo orderer to a 2.0 based setup with raft orderer. https://github.com/swetharepakula/Fabric101Workshop/commit/b8ad704e981b9dc7426f61826b7d0b11509a1c56

g2flyer:
> As some reference, here is work @swetharepakula did to go from a 1.4 based setup with solo orderer to a 2.0 based setup with raft orderer. [swetharepakula/Fabric101Workshop@b8ad704](https://github.com/swetharepakula/Fabric101Workshop/commit/b8ad704e981b9dc7426f61826b7d0b11509a1c56)

Glancing through it and seeing the references to tls/server.crt makes me remember that Swetha said that raft requires tls.  So i guess the main (only?) hurdle to cross from the error i got above is to create TLS certificates and put them in the right place.

swetharepakula:
Yes Raft requires TLS. Cryptogen will create the TLS certs automatically, so you only need to mount them and configure the peer, orderer, ca to use them. If you look at the commit @MHBauer linked above you should see all the necessary changes needed for the peer, orderer and the ca to enable tls.

mbrandenburger:
@Jordy24 please re-request reviews once you are done with your changes and want us to have another look. Thanks :)

g2flyer:
> Yes Raft requires TLS. Cryptogen will create the TLS certs automatically, so you only need to mount them and configure the peer, orderer, ca to use them. If you look at the commit @MHBauer linked above you should see all the necessary changes needed for the peer, orderer and the ca to enable tls.
Note: this is not for the docker-compose setting, but the run-directly-on-host way the integration tests are run. So no mount and integration tests previously also didn't use any cryptogen ...

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
...
-    if [ -e "${FABRIC_STATE_DIR}/${CHAN_ID}.creator" ]; then
-	try $RUN ${FABRIC_BIN_DIR}/peer chaincode instantiate -n ${ERCC_ID} -v ${ERCC_VERSION} -c '{"args":["init"]}' -C ${CHAN_ID} -V ercc-vscc
-	sleep 4
-	try rm "${FABRIC_STATE_DIR}/${CHAN_ID}.creator"
-    fi
-    #   - get SPID (mostly as debug output)
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n ${ERCC_ID} -c '{"args":["getSPID"]}' -C ${CHAN_ID}
+    say "Installing ercc on channel '${CHAN_ID}' ..."
+    echo "Packaging chaincode ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode package ercc.tar.gz -p ../../ercc --lang golang --label erccv1
```
g2flyer: can we create the package in some tempspace rather than inside the repo? If in repo, then we definitely should remove it after install and/or errors (probably a good idea even if we install in a temp dir).  Maybe then also define a variable so we have the path only at a single place?

integration/config/core.yaml
g2flyer: This still applies to the updated version!

fabric/bin/lib/common_ledger.sh
g2flyer: As discussed on 2020-03-26, conversion to raft is moved out to separate PR. I created a corresponding issue #317

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
...
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode checkcommitreadiness --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1 --output json 
+    para
+    sleep 3
+    echo "Commit chaincode definition...."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode commit -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1 
+    para
+    sleep 3
+    echo "Query commited chaincodes on the channel"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} 
+    #try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} --name ercc
```
g2flyer: this seems dead code? Remove?

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
     try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
 
     # - handle ercc
-    say "Installing & Instantiating ercc on channel '${CHAN_ID}' ..."
-    #   - install ercc
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode install -n ${ERCC_ID} -v ${ERCC_VERSION} -p github.com/hyperledger-labs/fabric-private-chaincode/ercc/cmd
-    sleep 1
-    #   - instantiate ercc iff "channel creation" peer
-    if [ -e "${FABRIC_STATE_DIR}/${CHAN_ID}.creator" ]; then
```
g2flyer: Unless you remove the creator logic completely, you should wrap below `commit` (which is done only by a single peer) as done here ..

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
...
+    para
+    sleep 3
+    echo "In installing chaincode ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode install ercc.tar.gz 
+    para
+    sleep 3
+    echo "Query installed chaincodes.."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode queryinstalled >&log.txt
+    para
+    export PACKAGE_ID=$(awk '/Package ID:/{print}' log.txt | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
```
g2flyer: minor thing, but if you need a variable only locally in the script as here i would drop the `export ` ...

integration/config/configtx.yaml
```diff
@@ -36,19 +44,22 @@ Organizations:
         Policies: &SampleOrgPolicies
             Readers:
                 Type: Signature
-                Rule: "OR('SampleOrg.member')"
```
g2flyer: Why did you change this from the sampleconfig version?  I think they also are functionally identical?  You probably added it in the attempt of getting raft working.  I think we should that though only cleanly in a separate PR with this PR cleanly based on fabric sampleconfig, a half-way conversion to raft mixed with a v1 to v2 conversion . This also applies to many related changes below

integration/config/core.yaml
```diff
@@ -388,10 +388,10 @@ peer:
         decorators:
           -
             name: DefaultDecorator
-          -
+            #-
```
g2flyer: Minor but would be nice to have the for leading '#' aligned

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
...
+    sleep 3
+    echo "Query commited chaincodes on the channel"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} 
+    #try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} --name ercc
+    para
+    sleep 3
+    echo "call chaincode invoke..."
+    try $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke -n ercc -c '{"function":"getSPID","args":[]}' -C ${CHAN_ID} 
     sleep 3
-
```
g2flyer: Don't delete below tlcc stuff, just comment it out with a preceeding TODO comment that we will have to later re-enable it once we have integrated tlcc ..

fabric/bin/lib/common_ledger.sh
```diff
@@ -111,7 +111,12 @@ ledger_init() {
     try mkdir -p ${FABRIC_STATE_DIR}
     try rm -rf ${FABRIC_STATE_DIR}/*
 
+    # this part would be used for raft
+    #try ${FABRIC_BIN_DIR}/cryptogen generate --config=./crypto-config.yaml --output="organizations" 
     # 2. start orderer
+    # - Creating a genesisblock for orderer-system-channel......."
+    try ${CONFIGTXGEN_CMD} -outputBlock genesisblock -profile SampleDevModeSolo -channelID orderer-system-channel
```
g2flyer: Actually, realized only now that this command is also an addition compared to v1.4. The issue right now is that this generates a file (genesisblock) in the repo. So we should ever add that to .gitignore or maybe create it in `/tmp/hyperledger/test`  (where we probably should put ercc.tar.gz, see other comment on temporary files). In the latter case we also would have to probably change the `BootstrapFile` stanza in `orderer.yml`

fabric/bin/peer.sh
```diff
@@ -331,25 +331,39 @@ handle_channel_join() {
...
-    #   - instantiate ercc iff "channel creation" peer
-    if [ -e "${FABRIC_STATE_DIR}/${CHAN_ID}.creator" ]; then
-	try $RUN ${FABRIC_BIN_DIR}/peer chaincode instantiate -n ${ERCC_ID} -v ${ERCC_VERSION} -c '{"args":["init"]}' -C ${CHAN_ID} -V ercc-vscc
-	sleep 4
-	try rm "${FABRIC_STATE_DIR}/${CHAN_ID}.creator"
-    fi
-    #   - get SPID (mostly as debug output)
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n ${ERCC_ID} -c '{"args":["getSPID"]}' -C ${CHAN_ID}
+    say "Installing ercc on channel '${CHAN_ID}' ..."
+    echo "Packaging chaincode ..."
```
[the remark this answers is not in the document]
mbrandenburger: note that when you run this script on a mac ... your terminal starts talking to you :)

integration/config/orderer.yaml
```diff
@@ -1,8 +1,9 @@
-# Copyright Intel Corp. All Rights Reserved.
```
mbrandenburger: Please bring back the copyright notice

mbrandenburger:
Thanks Jordan! Ledger_init and shutdown works in my test setting. Some minor changes still required.

fabric/bin/peer.sh
```diff
@@ -332,39 +332,58 @@ handle_channel_join() {
...
+    say "call chaincode invoke..."
     try $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke -n ercc -c '{"function":"getSPID","args":[]}' -C ${CHAN_ID} 
     sleep 3
 
+    # TODO: Enable once tlcc is integrated
+    ##   - install ercc
+    #try $RUN ${FABRIC_BIN_DIR}/peer chaincode install -n ${ERCC_ID} -v ${ERCC_VERSION} -p github.com/hyperledger-labs/fabric-private-chaincode/ercc/cmd
+    #sleep 1
+    ##   - instantiate ercc iff "channel creation" peer
+    #if [ -e "${FABRIC_STATE_DIR}/${CHAN_ID}.creator" ]; then
```
g2flyer: Hmm, what i meant in my prior comment related to channel-creator was not that copy as-is and comment out but you would have to do the put the `commit` in the same conditional statement as the `instantiate` in v1.4. Remember we want to replicate the same semantics we had before for v1.4 with v2 and `commit` essentially replaced `instantiate` ...

fabric/bin/peer.sh
g2flyer: Not directly resolved in new commit but added separate comment on new version, so resolve this here ...

integration/config/configtx.yaml
```diff
@@ -254,7 +268,7 @@ Orderer: &OrdererDefaults
     # as TLS validation.  The preferred way to specify orderer addresses is now
     # to include the OrdererEndpoints item in your org definition
     Addresses:
-        # - 127.0.0.1:7050
+            #- orderer.example.com:7050
```
g2flyer: This is a superfluos difference (and in context of integration config, orderer.example.com doesn't make sense as the only host existing is localhost ..)

integration/config/configtx.yaml
```diff
@@ -23,11 +23,20 @@ Organizations:
...
+        # this property must always be set to false for orgs included in block
+        # creation.
+        SkipAsForeign: false
+
         # ID is the key by which this org's MSP definition will be referenced.
         # ID can include alphanumeric characters as well as dots and dashes.
         ID: SampleOrg
 
         # MSPDir is the filesystem path which contains the MSP configuration.
+        #MSPDir: organizations/peerOrganizations/sampleorg.example.com/msp
```
g2flyer: I've mentioned that in an earlier review but i guess it got shadowed by me adding the review and you pushing an overlapping update: this comment seems a superfluous difference to the sampleconfig version. Let's keep the files such that differences to the templates only contain essential differences, i.e., if i do a `diff -ur ../../hyperledger/fabric/sampleconfig/ integration/config/ ` i see only differences which matter (with the exception of this and below comment on orderer.example.com i think this is actually true now)

fabric/bin/peer.sh
Jordy24: I never really needed it so that's why I left in the comments.

g2flyer: hmm, but you changed the semantics. We might not necessarily need it but the current commented out code as-is does not make sense (why would old lifecycle command show up in middle of new one) and also break the logic of creator. That's why i mentioned in my earlier code we either retain the original semantics -- where the code tried to do the instantiate (meaning commit now for new lifecycle) exactly once if you would have multiple peers -- or then just remove the creator logic throughout the code.

fabric/bin/peer.sh
```diff
@@ -331,25 +331,46 @@ handle_channel_join() {
...
+    say "In installing chaincode ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode install ${FABRIC_STATE_DIR}/ercc.tar.gz 
+    para
+    sleep 3
+    say "Query installed chaincodes.."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode queryinstalled >&log.txt
+    para
+    PACKAGE_ID=$(awk '/Package ID:/{print}' log.txt | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
+    say "Approve for my org"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode approveformyorg -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --package-id ${PACKAGE_ID} --sequence 1
```
g2flyer: sorry, noticed only now: but why do you hard-code here names like ercc and versions like 1 when we already had pre-defined (and previously used) variables lke ERCC_VERSION and ERCC_ID?   Sequence didn't exist before  but you might want to add a variable for that also (the vars are currently declared in fabric/bin/lib/common_ledger.sh.

fabric/bin/peer.sh
```diff
@@ -331,25 +331,46 @@ handle_channel_join() {
...
+    say "Query installed chaincodes.."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode queryinstalled >&log.txt
+    para
+    PACKAGE_ID=$(awk '/Package ID:/{print}' log.txt | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
+    say "Approve for my org"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode approveformyorg -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --package-id ${PACKAGE_ID} --sequence 1 
+    para
+    sleep 3
+    say "Check for commit readiness"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode checkcommitreadiness --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1 --output json
```
g2flyer: also noticed only now: the instantiate had previously a '-V ercc-vscc' argument.  Right now we don't have the plugin but we still will need it once it is re-enabled. The argument is still the same format and will have to be passed to all of 'approveformyorg', 'checkcommitreadiness' and 'commit'. 
```suggestion
    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode checkcommitreadiness --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1 --output json # -V ercc-vscc   # TODO: re-add validation plugin once they are enabled in peer ...
```

fabric/bin/peer.sh
```diff
@@ -331,25 +331,46 @@ handle_channel_join() {
...
+    say "Approve for my org"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode approveformyorg -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --package-id ${PACKAGE_ID} --sequence 1 
+    para
+    sleep 3
+    say "Check for commit readiness"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode checkcommitreadiness --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1 --output json 
+    para
     sleep 3
+    say "Commit chaincode definition...."
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode commit -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1
```
g2flyer:
```suggestion
    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode commit -o ${ORDERER_ADDR} --channelID ${CHAN_ID} --name ercc --version 1 --sequence 1  # -V ercc-vscc   # TODO: re-add validation plugin once they are enabled in peer ...
```

fabric/bin/peer.sh
```diff
@@ -331,25 +331,46 @@ handle_channel_join() {
...
+    sleep 3
+    say "Query commited chaincodes on the channel"
+    try $RUN ${FABRIC_BIN_DIR}/peer lifecycle chaincode querycommitted --channelID ${CHAN_ID} 
+    para
+    sleep 3
+    say "call chaincode invoke..."
+    try $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke -n ercc -c '{"function":"getSPID","args":[]}' -C ${CHAN_ID} 
+    sleep 3
+
+    # TODO: Enable once tlcc is integrated
```
g2flyer: minor comment, but putting the TODO right in front of the commented out `say` and `try` would seem more obvious than having it here with even an empty line between it and the tlcc block (but that is obviously a subjective take)

fabric/bin/lib/common_ledger.sh
```diff
@@ -48,6 +48,7 @@ define_common_vars() {
     : ${CHAN_ID:="mychannel"}
     : ${ERCC_ID:="ercc"}
     : ${ERCC_VERSION:="0"}
+    : ${SEQUENCE:="1"}
```
g2flyer: for next teim, given that this is an sequence number for ercc, prefixing the var with ERCC_ as for the other variables would have been a bit more consistant

integration/config/core.yaml
```diff
@@ -559,35 +580,18 @@ chaincode:
...
-      #   invokableExternal: true
-      #   invokableCC2CC: true
-      #
-      # FPC Addition:5  
-      - enabled: true
-        name: tlcc
-        path: ../../tlcc/tlcc.so
-        invokableExternal: true
-        invokableCC2CC: true
+        #tlcc: enable  # FPC Addition:4
```
mbrandenburger: ^

integration/config/core.yaml
```diff
@@ -559,35 +580,18 @@ chaincode:
...
 
     # Logging section for the chaincode container
     logging:
       # Default level for all loggers within the chaincode container
+      #level:  info
```
mbrandenburger: ^

fabric/bin/lib/common_ledger.sh
```diff
@@ -48,7 +48,7 @@ define_common_vars() {
     : ${CHAN_ID:="mychannel"}
     : ${ERCC_ID:="ercc"}
     : ${ERCC_VERSION:="0"}
-    : ${SEQUENCE:="1"}
+    : ${ERCC_SEQUENCE:="1"}
```
g2flyer: don't forget to change also the other references to SEQUENCE in peer.sh!
Maybe you are already in the process of doing so, in that case, try to not add a commit for every individual change but try to bundle related changes into a single commit.   You might also cross-check a bit more careful with a git diff/git diff -staged before you commit.  At least not sure whether your commit [33b875c2433fc5](https://github.com/hyperledger-labs/fabric-private-chaincode/pull/307/commits/33b875c2433fc53604551cd27c913dfe819e2a7c) was really intentional (the only thing it added was a single white-space difference comparted to the original version?)

fabric/bin/peer.sh
```diff
@@ -331,25 +331,45 @@ handle_channel_join() {
...
+
 
-    # - ask tlcc to join channel
-    #   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
-    #   it will re-join old channels but tlcc will not!
-    say "Attaching TLCC to channel '${CHAN_ID}' ..."
-    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID}
+    ## - ask tlcc to join channel
+    ##   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
+    ##   it will re-join old channels but tlcc will not!
```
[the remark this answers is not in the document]
Jordy24: this was there on the original file so just put it back as-is

integration/config/core.yaml
Jordy24: what is wrong with this part???

mbrandenburger: missing space but this is sugar!

fabric/bin/peer.sh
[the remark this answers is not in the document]
mbrandenburger: here is some form of double comments;

PR #315, opened 2020-03-25
Enable Mock server debug logging

### LINKED ISSUE #248: Remove fabric v1.4 shim logging
In many places (i.e., ERCC, TLCC, ...) we use `shim.NewLogger(...)`. This has been removed with Fabric 2.0.

For TLCC we can follow qscc and other sys chaincodes and use the following:

```
var logger = flogging.MustGetLogger("tlcc")
```

What about ECC and ERCC? Can we use this there as well or replace my simple fmt.Printf ???

mbrandenburger:
This PR brings back debug logging to the mock server.
As flogging comes with fabric we also it with FPC. The mock server, as
well as other components such as ERCC and ECC can be controlled through
FABRIC_LOGGING_SPEC env variable.

Fixes #248

PR #88 can build on that one here.

```

### DIFF
--- a/demo/client/backend/mock/server.go
+++ b/demo/client/backend/mock/server.go
@@ -59,8 +59,11 @@ func main() {
 	flag.Parse()
 
 	if flagDebug {
-		// TODO find a way to set flogger log level to debug
-		//logger.SetLevel(shim.LogDebug)
+		// We activate our log level here, note that we can also directly control
+		// log level via FABRIC_LOGGING_SPEC. For more fine grained logging we could
+		// also use different log level for loggers.
+		// For example: FABRIC_LOGGING_SPEC=server=INFO:ecc=DEBUG:ecc_enclave=ERROR
+		flogging.ActivateSpec("DEBUG")
 	}
 
 	// deploy
--- a/ecc/cmd/main.go
+++ b/ecc/cmd/main.go
@@ -15,6 +15,11 @@ import (
 var logger = flogging.MustGetLogger("ecc")
 
 func main() {
+
+	// we can control logging via FABRIC_LOGGING_SPEC, the default is FABRIC_LOGGING_SPEC=INFO
+	// For more fine grained logging we could also use different log level for loggers.
+	// For example: FABRIC_LOGGING_SPEC=ecc=DEBUG:ecc_enclave=ERROR
+
 	// create enclave chaincode
 	t := ecc.NewEcc()
 	defer t.Destroy()

### CODE REVIEW

File: ecc/cmd/main.go [hunk 1]
```diff
@@ -15,6 +15,11 @@ import (
 var logger = flogging.MustGetLogger("ecc")
 
 func main() {
+
+	// we can control logging via FABRIC_LOGGING_SPEC, the default is FABRIC_LOGGING_SPEC=INFO
+	// For more fine grained logging we could also use different log level for loggers.
+	// For example: FABRIC_LOGGING_SPEC=ecc=DEBUG:ecc_enclave=ERROR
```
g2flyer: Why not just check whether FABRIC_LOGGIN_SPEC exists as environment variable and take it from there?  This should work both with current container approach as well with external-builder. in the latter case, of course we would have more leeway, e.g., via command options like mock server, but environment variable works in both cases and is also how fabric itself handles it.   Insofar, we might even change mockserver to also use environment variable, at least in addition? (Note that this gives you also more control over logging in case you need it via a more complicated logging spect ...

mbrandenburger: flogger already takes care of reading `FABRIC_LOGGIN_SPEC` from env. There is no need for additional code here. See [this](https://github.com/hyperledger/fabric/blob/master/common/flogging/logging.go#L87).

That is, the loggers used in mock server and ecc will read whatever is specified in `FABRIC_LOGGIN_SPEC`.

- Log levels are defined [here](https://github.com/hyperledger/fabric/blob/master/common/flogging/levels.go#L36)

g2flyer: Oops, i could have sworn i saw here a commented out flogging.ActivateSpec.  I guess too early in the morning.  In any case, even better. Then we should be good to go for (controlled) logging in chaincodes, see my comment on peer forwarding environment variables ...

demo/client/backend/mock/server.go
```diff
@@ -59,8 +59,11 @@ func main() {
 	flag.Parse()
 
 	if flagDebug {
-		// TODO find a way to set flogger log level to debug
-		//logger.SetLevel(shim.LogDebug)
+		// We activate our log level here, note that we can also directly control
```
bvavala: (reading the comments below) if flogger already relies on FABRIC_LOGGING_SPEC, do we still need the flagDebug here?

mbrandenburger: NO - not really needed anymore, however I suggest we keep it as it is and once we decide to revamp the mock server and use it somewhere else for testing we can refactor it and get rid of it.

Issue #269, opened 2020-03-19
cc-builder should be based on local repo rather than on master from github


g2flyer:
cc-builder dockerfile depends on the fpc code checked out from dev which is master from github repo.  To make it properly work, though, it should be based on the current (local) version as done in peer.  So one way fix would be to base it on peer and "nullify" peer-specific. Given the overlap between peer and dev, we might anyway want to consolidate?

Notes: 
- cc-builder is currently disabled
- cc-builder is not on the critical path but is required to get demo running with network setup, certainly a desirable thing to have. That said, while currently added to concept release 2.0 milestone, it certainly can be dropped from that if we don't shoot for dragging the demo along with the rest of the code for now ....
- cc-build might eventually be necessary to be used for the `build` external builder script but that might anyway require some refactor ...

mbrandenburger:
I think this issue is actually about our dev image not cc-builder :D

mbrandenburger:
> Given the overlap between peer and dev, we might anyway want to consolidate?

would it make sense to have a "lightweight" peer image ... that just contains a fpc-custom-build peer binary and the corresponding "plugins" (as long as we still have some). This peer container does not need to build the peer itself right?

mbrandenburger:
the dev container I would like to keep as a full fledge dev environment ... able to build FPC and run FPC chaincode using simulator (of if SGX-hardware is available even use that)

g2flyer:
> > Given the overlap between peer and dev, we might anyway want to consolidate?
> 
> would it make sense to have a "lightweight" peer image ... that just contains a fpc-custom-build peer binary and the corresponding "plugins" (as long as we still have some). This peer container does not need to build the peer itself right?

well somebody should build it and it should be corresponding the "current" (host) code. So you definitely could split into two different images, a builder and a runner, but i'm not sure where that would simplify your life ...

PR #319, opened 2020-03-31
Adapt to recent openssl download reorganization

g2flyer:

Subject says it all.

N/A

Note that travis will fail eventually but it should only be later in the docker build of cc-builder due to the known issue that dev image is built from master, not current branch (see #269).  Also note that the `peer` image currently still won't build but that should also be after the openssl build relevant part during the attempt to do a `make plugins` which is currently not yet working ...


g2flyer:
Hmm, the PR actually had some fixup commits, but i guess it's now a bit later for that ...

Issue #250, opened 2020-03-12
Make custom-vscc for ECC and ERCC Fabric 2.0 compatible


mbrandenburger:
There were rumours that plugin support has been removed from Fabric 2.0.

However, according to the [release notes](https://github.com/hyperledger/fabric/releases/tag/v2.0.0) only system chaincode plugins have been removed (FAB-15343). This does not necessary also mean custom validation plugin is dropped.

mbrandenburger:
- Fabric 2.0 documentation still contains the custom validation plugins section, see [here](https://hyperledger-fabric.readthedocs.io/en/release-2.0/pluggable_endorsement_and_validation.html)

- Validation code still contains plugin dispatcher, see [here](https://github.com/hyperledger/fabric/blob/a282f7132565d509d3ba6d502cd9c07cc5e567f0/core/committer/txvalidator/v20/validator.go#L74)

- AND this plugin dispatcher is also called, see [here](https://github.com/hyperledger/fabric/blob/a282f7132565d509d3ba6d502cd9c07cc5e567f0/core/committer/txvalidator/v20/validator.go#L374)

mbrandenburger:
Note that `github.com/hyperledger/fabric/protos/utils` moved to  `github.com/hyperledger/fabric/protoutil` which is used in `ercc/vscc/ercc_validation_logic.go`

g2flyer:
maybe i'm missing something, but weren't these also go plugin with .so files and have to be called by peer?  So with plugins gone, how does the corresponding code of a specific plugin called?!

mbrandenburger:
This issue was not yet fixed with #260

mbrandenburger:
#322 bring all plugins (ercc-vscc, ecc-vscc, and ercc-decorator) in one single folder and builds it.

Currently, ercc-decorator works nicely with a peer, but ercc-vscc und ecc-vscc still have go dependency issues / version conflicts.

PR #322, opened 2020-04-07
Enable ercc decorator

mbrandenburger:

- This PR brings back ercc-decorator
- Moves all plugins (ecc-vscc, ercc-vscc, and ercc-decorator) to a central `plugins`
- Refactor provider.go to make the attestation component self-contained

Also note that this PR builds on #320 

This PR is related to #250 

Note that `ercc-decorator` must be build from fabric project as described in `plugins/README.md`. 

For testing use same template as used for #320 

```
export FABRIC_LOGGING_SPEC=debug
export FABRIC_CFG_PATH=$(/bin/pwd)/integration/config/ 
export FABRIC_BIN_DIR=$(readlink -f "../../hyperledger/fabric/build/bin")
export PATH=${FABRIC_BIN_DIR}:${PATH}

./fabric/bin/ledger_shutdown.sh

CHAN_ID=mychannel
PEER=./fabric/bin/peer.sh

./fabric/bin/ledger_init.sh

# example chaincode to use ..
# - echo
LBL=echo
VER=0
CC_LANG=fpc-c
CC_PATH=examples/echo/_build/lib/

PKG=/tmp/${LBL}.tar.gz

${PEER} lifecycle chaincode package --lang ${CC_LANG} --label ${LBL} --path ${CC_PATH} ${PKG}
${PEER} lifecycle chaincode install ${PKG}

${PEER} lifecycle chaincode queryinstalled

PKG_ID=$(${PEER} lifecycle chaincode queryinstalled | awk "/Package ID: ${LBL}/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')

${PEER} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${LBL} --version ${VER} # --init-required # extra flag required for auction test
${PEER} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${LBL} --version ${VER}

${PEER} lifecycle chaincode commit -C ${CHAN_ID} --name ${LBL} --version ${VER}  # --init-required # extra flag required for auction test

${PEER} lifecycle chaincode querycommitted -C ${CHAN_ID}

# - echo
${PEER} chaincode invoke -C ${CHAN_ID} -n ${LBL} -c '{"Args": ["echo-1"]}'
#  should work ..

./fabric/bin/ledger_shutdown.sh
```

```

### CODE REVIEW

**Comment by g2flyer** [comment]
the change in build process would also have to be reflected in docker and travis?

File: ercc/README.md [hunk 1]
```diff
@@ -14,9 +14,7 @@ committing state changes.
 
 The enclave registry is implemented as a normal chaincode and comes with a
 custom validation plugin. Before you can install ercc at a peer you have to
-build the vscc plugin by running the following.
-
-    $ make
+build the custom-vscc plugin in `$FPC_PATH/plugins/`.
```
g2flyer, line 17: you might want to explicitly link to the corresponding README given that it has special build instructions? Actually, top-level readme also might have to be correspondingly be updated?

ercc/attestation/mock_ias.go
```diff
@@ -4,20 +4,13 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package mock
-
-import (
-	"github.com/hyperledger-labs/fabric-private-chaincode/ercc/attestation"
-	"github.com/hyperledger/fabric/common/flogging"
```
g2flyer, line 11: Having some logging of the (top-level) attestation provider calls could be useful. That said, it seems we didn't have any before (but then i'm also surprised why go didn't complain about unused imports?)

ercc/attestation/provider.go.template
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
```
g2flyer, line 2: I guess you didn't use `git mv` when moving this file? (Luckily you seem to have done for the other plugin files you moved where retaining the history is a bit more crucial ...)

ercc/go.mod
```diff
@@ -1,29 +1,28 @@
 module github.com/hyperledger-labs/fabric-private-chaincode/ercc
 
-go 1.13
+go 1.14
```
g2flyer, line 3: Could you do a `git mod tidy` before checking in updated go.mods?  Right now all my go.mod/go.sum are updated, even if i do only a `make clean`!  (This is true also for the top-level go.mod ...)

plugins/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The current code base uses custom validation plugins for ecc (`ecc-vscc/`) and ercc (`ercc-vscc/`) to validate transactions. Additionally, ercc uses a decorator plugin (`ercc-decorator/`) to read attestation related data from the peer local filesystem during registration invocation.
+
+## Build plugins
+
+Go plugins are nasty little beasts and often cause pain when building. That is, a go plugin must be build with the same dependencies as used by the application that loads the plugin at runtime. Currently, Fabric still vendors dependencies which make building plugins even harder. So building plugins from the fabric project folder is the easiest way to overcome this at the moment.
+
+```
+$ ln -s $FPC_PATH/plugins $FABRIC_PATH/fpc_plugins
+$ cd $FABRIC_PATH/fpc_plugins; make
```
g2flyer, line 16: i wonder whether that actually works with all shells and/or shell configurations (e.g., tcsh and cwdcmd command ..)

g2flyer: seems also it is important that you do above _after_ you run `make provider` in ercc?

ercc/Makefile
```diff
@@ -23,7 +17,7 @@ IAS_GO_DEFINES+=-DUSE_SGX_HARDWARE_MODE
 endif
 
 TEMPLATE_SUFFIX=.template
-GO_TARGETS=provider.go
```
g2flyer, line 26: Wouldn't it be nicer/cleaner to outsource the template generation also to the `./attestation` subdirectory and a corresponding makefile?

g2flyer:
> Build fails, even after the specific fpc_plugin vs fpc build chicken'n'egg issue mentioned elsewhere. Fails as follows:

Ok, build failure was due to not doing a `make clean` in the _old_ git config rather than in the one of this PR. This meant that `ercc/provider.go` was not deleted. Manually deleting that file solved the build (though still requiring the chicken'n'egg workaround for fpc_plugin build alluded above)

g2flyer:
quick followup: as it builds now, i also did test with echo and helloworld. And indeed, in helloworld it actually does work even with stateful txs despite not having tlcc at all!
BTW: i wanted also to quickly check whether ercc registration was successful. But alas, seems right now you could test with ercc you can do it only with `getAttestationReport` which requires an enclavepk as input and just copy pasting the one returned by an FPC CC query did not work :-( Would be nice to use same format or have a `list registered enclaves` call in ercc or alike. Not that this would be really on the critical path nor worth adding now with ercc anyway soon changing quite a bit ... :-)

plugins/README.md
mbrandenburger: it works at least in the fpc docker dev env. Did someone test with some other shell? if someone experiences a problem here we can create an issue and fix later;

ercc/README.md
mbrandenburger: > BTW: i wanted also to quickly check whether ercc registration was successful. But alas, seems right now you could test with ercc you can do it only with `getAttestationReport` which requires an enclavepk as input and just copy pasting the one returned by an FPC CC query did not work :-( Would be nice to use same format or have a `list registered enclaves` call in ercc or alike. Not that this would be really on the critical path nor worth adding now with ercc anyway soon changing quite a bit ... :-)

Sounds like a good feature for the new ercc! @bvavala

mbrandenburger:
> Never mind, i guess it should have passed only due to the TLCC MockStub?

I think so

ercc/attestation/mock_ias.go
mbrandenburger: ehm the flogging import was indeed used by the `var logger`, however, the logger is no gone as this is not used.

ercc/attestation/provider.go.template
mbrandenburger: uuhh ... nice catch! I can fix that once I squashed the fixups.

ercc/go.mod
mbrandenburger: I actually did that! When I run `go mod tidy` right now it does not change anything?! Not sure what to do here?!

ercc/README.md
bvavala: More than good...it's needed!
Short term: from chaincode id to a one-item list -- the enclave at the designated peer for a specific chaincode.
Medium term: from chaincode id to a list -- the enclaves available in the channel for that chaincode.
Long term: a list of enclaves in the channel which can load and run a chaincode.

ercc/go.mod
mbrandenburger: should I just revert all changes to `go.mod`?

plugins/Makefile
```diff
@@ -0,0 +1,50 @@
...
+ERCC_VSCC_OUT = ercc-vscc.so
+
+ERCC_DECO_SRC = ercc-decorator/decoration.go
+ERCC_DECO_OUT = ercc-decorator.so
+
+FABRIC_FPC_PLUGINS = $(FABRIC_PATH)/fpc_plugins
+
+build:
+	@if [ ! -d $(FABRIC_FPC_PLUGINS) ]; then \
+		ln -s $(PWD) $(FABRIC_FPC_PLUGINS); \
```
mbrandenburger, line 25: this does not work when we run make from root.

plugins/README.md
g2flyer: Oh, as mentioned in btw i'm fine leaving it, just thought i mention that it could lead to some potential problems (which we can fix later ...).
i guess the issue with make provider ordering you are solving now by trying to automate the whole build and symlink process in our makefiles, hence making manual instructions obsolete?

ercc/go.mod
g2flyer: well, i had this issue already beforehand. Thought tidy might solve it but apparently not. Following is the diff i get after a build
```
msteiner@ubusrv1804-dev ~/.../src/github.com/hyperledger-labs/fabric-private-chaincode (pr/322 *)$ git diff
diff --git a/go.mod b/go.mod
index ab422b3..9add882 100644
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
        golang.org/x/net v0.0.0-20200320220750-118fecf932d8 // indirect
        golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
        golang.org/x/sys v0.0.0-20200320181252-af34d8274f85 // indirect
-       golang.org/x/tools v0.0.0-20200323164354-18ea2c8f7359 // indirect
+       golang.org/x/tools v0.0.0-20200408132156-9ee5ef7a2c0d // indirect
        google.golang.org/genproto v0.0.0-20200319113533-08878b785e9c // indirect
        google.golang.org/grpc v1.28.0 // indirect
        gopkg.in/ini.v1 v1.55.0 // indirect
diff --git a/go.sum b/go.sum
index 6a0dd40..470e2fc 100644
--- a/go.sum
+++ b/go.sum
@@ -323,6 +323,7 @@ github.com/xeipuuv/gojsonschema v0.0.0-20180618132009-1d523034197f/go.mod h1:5yf
 github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2/go.mod h1:UETIi67q53MR2AWcXfiuqkDkRtnGDLqkBTpCHuJHxtU=
 github.com/xordataexchange/crypt v0.0.3-0.20170626215501-b2862e3d0a77/go.mod h1:aYKd//L2LvnjZzWKhF00oedf4jCCReLcmhLdhm1A27Q=
 github.com/yuin/goldmark v1.1.25/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
+github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 go.etcd.io/bbolt v1.3.2/go.mod h1:IbVyRI1SCnLcuJnV2u8VeU0CEYM7e686BmAb1XKL+uU=
 go.opencensus.io v0.22.0/go.mod h1:+kGneAE2xo2IficOXnaByMWTGM9T73dGwxeWcUqIpI8=
 go.uber.org/atomic v1.4.0 h1:cxzIVoETapQEqDhQu3QfnvXAV4AlzcvUCxkVUFw3+EU=
@@ -426,6 +427,8 @@ golang.org/x/tools v0.0.0-20191029190741-b9c20aec41a5/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20200323164354-18ea2c8f7359 h1:7R5qvs7puR7tYevClYCMHTOFM3jJbRl2dbKOfmNw7dE=
 golang.org/x/tools v0.0.0-20200323164354-18ea2c8f7359/go.mod h1:Sl4aGygMT6LrqrWclx+PTx3U+LnKx/seiNR+3G19Ar8=
+golang.org/x/tools v0.0.0-20200408132156-9ee5ef7a2c0d h1:2DXIdtvIYvvWOcAOsX81FwOUBoQoMZhosWn7KjXEl94=
+golang.org/x/tools v0.0.0-20200408132156-9ee5ef7a2c0d/go.mod h1:EkVYQZoAsY45+roYkvgYkIh4xh/qjgUK9TdY2XT94GE=
 golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
```

g2flyer: Note, though, this time i don't seem to get it anymore for ercc but only in the top-level. On top-level i have it now since quite a while. I got it also at some point-in-time for ercc but not anymore now. So probably your change is fine. But as you update

mbrandenburger:
> the change in build process would also have to be reflected in docker and travis?

not so sure about that; with the current plugins make I think there is not need to change something in docker and travis. This here is different from the tlcc changes in PR #323

plugins/Makefile
```diff
@@ -0,0 +1,48 @@
...
+ECC_VSCC_OUT = ecc-vscc.so
+
+ERCC_VSCC_SRC = ercc-vscc/ercc_validation_plugin.go ercc-vscc/ercc_validation_logic.go
+ERCC_VSCC_OUT = ercc-vscc.so
+
+ERCC_DECO_SRC = ercc-decorator/decoration.go
+ERCC_DECO_OUT = ercc-decorator.so
+
+build:
+	ln -sfn $$(/bin/pwd) $(FABRIC_FPC_PLUGINS); \
```
g2flyer, line 24: recreating the symlink on each build feels a bit excessive but if it works ... :-)

mbrandenburger: I happy for any suggestion here :D

g2flyer:
> > the change in build process would also have to be reflected in docker and travis?
> 
> not so sure about that; with the current plugins make I think there is not need to change something in docker and travis. This here is different from the tlcc changes in PR #323

Well, this comment was when you still required manual steps.  If everything is automated via make, then of course it doesn't (but then i also wouldn't give any manual instructions in the readme for this as you still do?)

mbrandenburger:
> (but then i also wouldn't give any manual instructions in the readme for this as you still do?)

removed it!

plugins/Makefile
g2flyer: you could make FABRIC_FPC_PLUGINS a target (with the ln as the rule) and add dependency here?  (alternatively, you could test whether it already exists before creating `if [ -x .... `)?

ercc/attestation/provider.go.template
g2flyer: if it is not much effort, then this would be nice. otherwise probably not worth. This might anyway go away sooner or later? (e.g., iff we have to resort to externalbuilder for ercc because we have c-code in it, then we might as well also use build-tags?)

plugins/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The current code base uses custom validation plugins for ecc (`ecc-vscc/`) and ercc (`ercc-vscc/`) to validate transactions. Additionally, ercc uses a decorator plugin (`ercc-decorator/`) to read attestation related data from the peer local filesystem during registration invocation.
+
+## Build plugins
+
+Go plugins are nasty little beasts and often cause pain when building. That is, a go plugin must be build with the same dependencies as used by the application that loads the plugin at runtime. Currently, Fabric still vendors dependencies which make building plugins even harder. So building plugins from the fabric project folder is the easiest way to overcome this at the moment.
+
+We provide a make script that is "smart" enough to build the plugins from within fabric project. Just run 
+```
+$ cd plguins
```
g2flyer, line 16:
```suggestion
$ cd plugins
```

ercc/attestation/provider.go.template
mbrandenburger: ok - let's keep it.

PR #323, opened 2020-04-07
Enable tlcc

### LINKED ISSUE #242: Integration of TLCC without go/plugins and cc2syscc calls
Current code integrates TLCC as a systemchaincode using go plugins and ERCC calls TLCC via chaincode2chaincode calls.

We currently build TLCC using `go build -buildmode=plugin`. See [makefile](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/master/tlcc/Makefile#L13)

TLCC is then hooked into the peer via `core.yaml`
```
chaincode:
    system:
        tlcc: enable

    systemPlugins:
        - enabled: true
        name: tlcc
        path: /path-to/fabric-private-chaincode/tlcc/tlcc.so
        invokableExternal: true
        invokableCC2CC: true
```

However, with Fabric 2.0 [release](https://github.com/hyperledger/fabric/releases/tag/v2.0.0) this is not supported anymore.

```
FAB-15285: Support for invoking system chaincodes from user chaincodes
has been removed.

System chaincodes, for example QSCC, are intended to be
invoked by a client rather than by a user chaincode. Invoking from a user
chaincode caused deadlocks in prior releases.
```

```
FAB-15343: System Chaincode Plugins have been removed.

As part of a general
move away from go plugins as an extension mechanism for Fabric, the ability to
add system chaincodes via go plugins has been removed. Users wishing to extend
Fabric with custom system chaincodes may rebuild the peer binary with the
system chaincode built into the binary. This system chaincode should then be
defined and initialized like any other user chaincode would be. This new model
is very similar to the plugin model (which required that the plugin to be built
at the same exact commit of Fabric), and addresses the significant shortcomings
around the lifecycle and validation of system chaincode transactions.
```

mbrandenburger:

This PR brings back TLCC for Fabric v2.

- Add patch to integrate TLCC as syscc and compile it with the peer
- Remove old TLCC plugin build target
- Remove TLCC mock from ECC

Fixes #242 

This PR depends on #320

```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "tlcc_enclave/protos/fabric"]
+	path = tlcc_enclave/protos/fabric
+	url = https://github.com/hyperledger/fabric-protos.git
+	branch = release-2.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ services:
 addons:
   apt:
     packages:
-      # TODO upgrade to protobuf 3.5 or higher
       - protobuf-compiler
       - python-protobuf
 
@@ -45,6 +44,12 @@ env:
     - SGX_REPO=https://download.01.org/intel-sgx/linux-2.6/ubuntu18.04-server
     - SGX_LIB_COMMON=libsgx-enclave-common_2.6.100.51363-bionic1_amd64.deb
     - SGX_SDK_BIN=sgx_linux_x64_sdk_2.6.100.51363.bin
+    # new protobuf
+    - PROTO_VERSION=3.11.4
+    - PROTO_ZIP=protoc-${PROTO_VERSION}-linux-x86_64.zip
+    - PROTO_REPO=https://github.com/google/protobuf/releases/download
+    - PROTO_DIR=/usr/local/proto3
+    - PROTOC_CMD=${PROTO_DIR}/bin/protoc
 
 before_install:
   # SGX SDK
@@ -62,6 +67,8 @@ before_install:
   # NANOPB
   - git clone https://github.com/nanopb/nanopb.git ${NANOPB_PATH}
   - pushd ${NANOPB_PATH} && git checkout nanopb-${NANOPB_VERSION} && cd generator/proto && make && popd
+  # Newer Protobuf
+  - wget ${PROTO_REPO}/v${PROTO_VERSION}/${PROTO_ZIP}; sudo unzip ${PROTO_ZIP} -d ${PROTO_DIR}
   # Fabric
   - git clone --branch $FABRIC_VERSION https://github.com/hyperledger/fabric.git $FABRIC_PATH
   - pushd $FABRIC_PATH
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ services:
 addons:
   apt:
     packages:
+      # TODO upgrade to protobuf 3.5 or higher
       - protobuf-compiler
       - python-protobuf
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,9 @@ env:
     - PROTOC_CMD=${PROTO_DIR}/bin/protoc
 
 before_install:
+  # git settings
+  - git config --global user.email "fpc_tester@fpc"
+  - git config --global user.name "fpc_tester"
   # SGX SDK
   - mkdir -p /opt/intel && mkdir -p /etc/init
   - pushd /opt/intel; wget ${SGX_REPO}/${SGX_LIB_COMMON}
@@ -70,10 +73,8 @@ before_install:
   # Newer Protobuf
   - wget ${PROTO_REPO}/v${PROTO_VERSION}/${PROTO_ZIP}; sudo unzip ${PROTO_ZIP} -d ${PROTO_DIR}
   # Fabric
-  - git clone --branch $FABRIC_VERSION https://github.com/hyperledger/fabric.git $FABRIC_PATH
-  - pushd $FABRIC_PATH
-  - git am $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/*.patch
-  - GO_TAGS=pluginsenabled make peer orderer cryptogen configtxgen && popd
+  - git clone https://github.com/hyperledger/fabric.git $FABRIC_PATH
+  - pushd $FABRIC_PATH; git checkout tags/$FABRIC_VERSION -b "$FABRIC_VERSION-fpc"; popd;
 
 script:
   # travis doesn't check out the branch itself but for our make to work we
--- a/README.md
+++ b/README.md
@@ -248,7 +248,9 @@ Make sure that you have the following required dependencies installed:
 
 * Docker 18.x
 
-* Protocol Buffers 3.5.x or higher and [Nanopb](http://github.com/nanopb/nanopb) 0.3.9.2
+* Protocol Buffers 
+    - Protocol Buffers 3.0.x needed for the Intel SGX SDK
+    - Protocol Buffers 3.11.x or higher and [Nanopb](http://github.com/nanopb/nanopb) 0.3.9.2
 
 * SGX SDK v2.6 for [Linux](https://github.com/intel/linux-sgx)
 
@@ -296,6 +298,14 @@ For more detailed information consult the official nanopb documentation http://g
     $ cd generator/proto && make
 
 Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Chaincode.
+
+Moreover, in order to build Fabric protobufs we also require a newer Protobuf compiler than what is provided as standard ubuntu package and is used to build the
+Intel SGX SDK. For this reason you will have to download and install another version and use it together with Nanopb. Do not install the new protobuf, though, such that it is not found in your standard PATH but instead define the `PROTOC_CMD`, either as environment variable or via `config.override.mk` to point to the new `protoc` binary
+
+    $ wget https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-x86_64.zip
+    $ unzip protoc-3.11.4-linux-x86_64.zip -d /usr/local/proto3
+    $ export PROTOC_CMD=/usr/local/proto3/bin/protoc
+
 
 #### Clone Fabric Private Chaincode
 
--- a/README.md
+++ b/README.md
@@ -313,38 +313,24 @@ Clone the code and make sure it is on your `$GOPATH`. (Important: we assume in t
 
     $ git clone --recursive https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 
-#### Patch Fabric 
+#### Prepare Fabric
 
-Next we need to patch the Fabric peer and rebuild it in order to enable Fabric Private Chaincode support. 
+Fabric Private Chaincode support requires to re-build the Fabric peer. 
 
-Checkout Fabric 2.0.1 release and apply our patch using the following commands:
+Checkout Fabric 2.0.1 release using the following commands:
 
     $ export FABRIC_PATH=${GOPATH}/src/github.com/hyperledger/fabric
-    $ git clone --branch v2.0.1 https://github.com/hyperledger/fabric.git $FABRIC_PATH
-    $ cd $FABRIC_PATH
-    $ git am ../../hyperledger-labs/fabric-private-chaincode/fabric/*.patch
+    $ git clone https://github.com/hyperledger/fabric.git $FABRIC_PATH
+    $ cd $FABRIC_PATH; git checkout tags/v2.0.1 -b v2.0.1-fpc
     
-Note that this patch does currently not work with the Fabric master branch, therefore make sure you use the Fabric
-v2.0.1 branch.
+Note that Fabric Private Chaincode currently does not work with the Fabric `master` branch, therefore make sure you use the Fabric
+`v2.0.1` branch and create a local branch named `v2.0.1-fpc`. This is important as our build script applies some patches
+to the fabric peer to enable FPC support. Make sure the source of Fabric is in your ``$GOPATH``.
 
 ### Build Fabric Private Chaincode
 
-Once you have your development environment up and running (i.e., using our docker-based setup or install all all dependencies on your machine) you can build FPC and start developing your own FPC application.
+Once you have your development environment up and running (i.e., using our docker-based setup or install all dependencies on your machine) you can build FPC and start developing your own FPC application.
 
-Make sure Fabric is in your ``$GOPATH`` and you enable the plugin feature using `GO_TAGS=pluginsenabled`. Simply run
-
-    $ cd $FABRIC_PATH
-    $ GO_TAGS=pluginsenabled make
-
-Building Fabric may take a while and it's time to get a coffee. Also, be not surprised if unit tests fail. In order to
-just build the peer you can run the following command:
-```
-$ GO_TAGS=pluginsenabled make peer
-```
-Please make sure that the peer is _always_ built with GO_TAGS, otherwise our custom validation plugins will (silently!) be
-ignored by the peer, despite the settings in ``core.yaml``.
-
-Once you have setup your development environment including Intel SGX SDK and SSL, nanopb, and successfully patched and built Fabric you can build the Fabric Private Chaincode framework.  
 ```
 $ cd $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 $ make
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ Make sure that you have the following required dependencies installed:
 
 * Docker 18.x
 
-* Protocol Buffers 3.x and [Nanopb](http://github.com/nanopb/nanopb) 0.3.9.2
+* Protocol Buffers 3.5.x or higher and [Nanopb](http://github.com/nanopb/nanopb) 0.3.9.2
 
 * SGX SDK v2.6 for [Linux](https://github.com/intel/linux-sgx)
 
@@ -301,7 +301,7 @@ Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Ch
 
 Clone the code and make sure it is on your `$GOPATH`. (Important: we assume in this documentation and default configuration that your `$GOPATH` has a _single_ root-directoy!)
 
-    $ git clone https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
+    $ git clone --recursive https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 
 #### Patch Fabric 
 
--- a/config.mk
+++ b/config.mk
@@ -15,6 +15,11 @@ export SGX_BUILD ?= PRERELEASE
 
 export FABRIC_PATH ?= ${GOPATH}/src/github.com/hyperledger/fabric
 
+# Give the option to overload by custom protoc
+# e.g. this is overloaded by travis and docker dev as we use protoc 3.11.4 to build
+# protos in tlcc_enclave but use protoc 3.0.x to build SGX SDK and SSL
+export PROTOC_CMD ?= protoc
+
 JAVA ?= java
 PLANTUML_JAR ?= plantuml.jar
 PLANTUML_CMD ?= $(JAVA) -jar $(PLANTUML_JAR)
--- a/ecc/enclave_chaincode.go
+++ b/ecc/enclave_chaincode.go
@@ -57,14 +57,6 @@ func CreateMockedECC() *EnclaveChaincode {
 
 // Init sets the chaincode state to "init"
 func (t *EnclaveChaincode) Init(stub shim.ChaincodeStubInterface) pb.Response {
-	mrenclave, err := t.enclave.MrEnclave()
-	if err != nil {
-		return shim.Error(err.Error())
-	}
-	logger.Debugf("Init: chaincode [mrenclave=%s]", mrenclave)
-	if err := stub.PutState(utils.MrEnclaveStateKey, []byte(mrenclave)); err != nil {
-		return shim.Error(err.Error())
-	}
 	return shim.Success(nil)
 }
 
@@ -112,6 +104,32 @@ func (t *EnclaveChaincode) setup(stub shim.ChaincodeStubInterface) pb.Response {
 		return shim.Error(errMsg)
 	}
 
+	// write mrenclave to ledger
+	mrenclave, err := t.enclave.MrEnclave()
+	if err != nil {
+		return shim.Error(err.Error())
+	}
+
+	// TODO this will change in the future; we will need to fetch MRENCLAVE from the chaincode definition
+	// check if MRENCLAVE is already on the ledger; if not, write it; if exists compare with enclave;
+	exists, err := stub.GetState(utils.MrEnclaveStateKey)
+	if err != nil {
+		return shim.Error(err.Error())
+	}
+
+	if len(exists) == 0 {
+		logger.Debugf("Set chaincode with mrenclave=%s", mrenclave)
+		if err := stub.PutState(utils.MrEnclaveStateKey, []byte(mrenclave)); err != nil {
+			return shim.Error(err.Error())
+		}
+	} else {
+		if mrenclave != string(exists) {
+			errMsg := fmt.Sprintf("ecc: MRENCLAVE has already been defined for this chaincode. Expected %s but got %s", string(exists), mrenclave)
+			logger.Errorf(errMsg)
+			return shim.Error(errMsg)
+		}
+	}
+
 	//get spid from ercc
 	spid, err := t.erccStub.GetSPID(stub, erccName, channelName)
 	if err != nil {
@@ -129,12 +147,6 @@ func (t *EnclaveChaincode) setup(stub shim.ChaincodeStubInterface) pb.Response {
 
 	enclavePkBase64 := base64.StdEncoding.EncodeToString(enclavePk)
 	quoteBase64 := base64.StdEncoding.EncodeToString(quoteAsBytes)
-
-	// we just add mrenclave to the proposal readset
-	_, err = stub.GetState(utils.MrEnclaveStateKey)
-	if err != nil {
-		return shim.Error(err.Error())
-	}
 
 	// register enclave at ercc
 	if err = t.erccStub.RegisterEnclave(stub, erccName, channelName, []byte(enclavePkBase64), []byte(quoteBase64)); err != nil {
--- a/ecc/enclave_chaincode.go
+++ b/ecc/enclave_chaincode.go
@@ -38,9 +38,7 @@ func NewEcc() *EnclaveChaincode {
 	logger.Debug("NewEcc")
 	return &EnclaveChaincode{
 		erccStub: &ercc.EnclaveRegistryStubImpl{},
-		// TODO disable TLCC for the moment
-		//tlccStub: &tlcc.TLCCStubImpl{},
-		tlccStub: &tlcc.MockTLCCStub{},
+		tlccStub: &tlcc.TLCCStubImpl{},
 		enclave:  enclave.NewEnclave(),
 		verifier: &crypto.ECDSAVerifier{},
 	}
--- a/fabric/0001-Integrate-tlcc-syscc.patch
+++ b/fabric/0001-Integrate-tlcc-syscc.patch
@@ -1,0 +1,47 @@
+From 0b29c14088c10e64ecaf918c9a96e5444d106e47 Mon Sep 17 00:00:00 2001
+From: Marcus Brandenburger <bur@zurich.ibm.com>
+Date: Tue, 7 Apr 2020 15:07:09 +0200
+Subject: [PATCH] Integrate tlcc syscc
+
+Loads tlcc system chaincode on peer startup.
+
+Signed-off-by: Marcus Brandenburger <bur@zurich.ibm.com>
+---
+ internal/peer/node/start.go | 6 +++++-
+ 1 file changed, 5 insertions(+), 1 deletion(-)
+
+diff --git a/internal/peer/node/start.go b/internal/peer/node/start.go
+index 340d6da26..909708e3c 100644
+--- a/internal/peer/node/start.go
++++ b/internal/peer/node/start.go
+@@ -74,6 +74,7 @@ import (
+ 	"github.com/hyperledger/fabric/core/scc/cscc"
+ 	"github.com/hyperledger/fabric/core/scc/lscc"
+ 	"github.com/hyperledger/fabric/core/scc/qscc"
++	"github.com/hyperledger/fabric/core/scc/tlcc"
+ 	"github.com/hyperledger/fabric/core/transientstore"
+ 	"github.com/hyperledger/fabric/discovery"
+ 	"github.com/hyperledger/fabric/discovery/endorsement"
+@@ -566,6 +567,7 @@ func serve(args []string) error {
+ 		"qscc":       {},
+ 		"cscc":       {},
+ 		"_lifecycle": {},
++		"tlcc":       {},
+ 	}
+ 
+ 	lsccInst := &lscc.SCC{
+@@ -719,8 +721,10 @@ func serve(args []string) error {
+ 		Metrics:                endorser.NewMetrics(metricsProvider),
+ 	}
+ 
++	tlccInst := tlcc.New(peerInstance)
++
+ 	// deploy system chaincodes
+-	for _, cc := range []scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC} {
++	for _, cc := range []scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC, tlccInst} {
+ 		if enabled, ok := chaincodeConfig.SCCWhitelist[cc.Name()]; !ok || !enabled {
+ 			logger.Infof("not deploying chaincode %s as it is not enabled", cc.Name())
+ 			continue
+-- 
+2.26.0
+
--- a/fabric/Makefile
+++ b/fabric/Makefile
@@ -1,0 +1,92 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+TOP = ..
+include $(TOP)/build.mk
+
+FABRIC_VERSION ?= v2.0.1
+FABRIC_PEER_BIN = $(FABRIC_PATH)/build/bin/peer
+FABRIC_TLCC = $(FABRIC_PATH)/core/scc/tlcc
+
+FABRIC_CURRENT_BRANCH = $(shell cd $(FABRIC_PATH); git rev-parse --abbrev-ref HEAD)
+FABRIC_CURRENT_TAG = $(shell cd $(FABRIC_PATH); git describe --tags)
+FABRIC_PATCHED_BRANCH = $(FABRIC_VERSION)-fpc
+FABRIC_PATCHED = $(FABRIC_PATH)/.fpc_patched
+FABRIC_TRACKED_CHANGES = $(shell cd $(FABRIC_PATH); git status -s | grep -v '\?')
+FABRIC_COMMITS = $(shell cd $(FABRIC_PATH); git log $(FABRIC_VERSION)..HEAD --oneline)
+FABRIC_LOG_DIFF = $(shell diff $(FABRIC_PATCHED) <(FABRIC_COMMITS))
+
+build: native
+
+patch: $(FABRIC_PATCHED)
+
+$(FABRIC_PATCHED):
+	@echo "Patching Fabric ..."
+	@if [ "$(FABRIC_CURRENT_BRANCH)" != "$(FABRIC_PATCHED_BRANCH)" ]; then \
+		echo "Aborting! Current branch is $(FABRIC_CURRENT_BRANCH) but want $(FABRIC_PATCHED_BRANCH)"; exit 1; \
+	fi
+
+	@if [ "$(FABRIC_CURRENT_TAG)" != "$(FABRIC_VERSION)" ]; then \
+		echo "Aborting! Current branch HEAD does not match tag/$(FABRIC_VERSION)! $(FABRIC_CURRENT_TAG) but want $(FABRIC_VERSION)"; exit 1; \
+	fi
+
+	cd $(FABRIC_PATH) && \
+	git am $(FPC_PATH)/fabric/*.patch && \
+	git log $(FABRIC_VERSION)..HEAD --oneline > $@
+
+clean-patch:
+	@echo "Cleaning Fabric patches ..."
+	@if [ "$(FABRIC_CURRENT_BRANCH)" != "$(FABRIC_PATCHED_BRANCH)" ]; then \
+		echo "Aborting! Current branch is $(FABRIC_CURRENT_BRANCH) but want $(FABRIC_PATCHED_BRANCH)"; \
+		echo "Please checkout $(FABRIC_PATCHED_BRANCH) and come back."; \
+		exit 1; \
+	fi
+
+	@if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
+		echo "Aborting! Nothing to do here ... it seems you haven't applied the patches yet"; exit 1; \
+	fi
+
+	@if [ ! -z "$(FABRIC_TRACKED_CHANGES)" ]; then \
+		echo "Aborting! It seems you have some changes in your branch:"; \
+		echo \\t$(FABRIC_TRACKED_CHANGES); \
+		echo "Please clean/stash your changes and try again."; \
+		exit 1; \
+	fi
+
+	@cd $(FABRIC_PATH) && git log $(FABRIC_VERSION)..HEAD --oneline > $(FABRIC_PATCHED).tmp
+	@if [ "$(diff $(FABRIC_PATCHED) $(FABRIC_PATCHED).tmp)" ]; then \
+  		echo "You have other commits than the patches:"; \
+   		cd $(FABRIC_PATH) && git log $(FABRIC_VERSION)..HEAD --oneline; \
+   		echo "Please check and resolve manually!"; \
+   		exit 1; \
+  	fi
+
+	@echo "Removing the following commits on branch ${FABRIC_PATCHED_BRANCH} in $(FABRIC_PATH)"
+	@cd $(FABRIC_PATH) && git log $(FABRIC_VERSION)..HEAD --oneline
+
+	cd $(FABRIC_PATH) && \
+	git reset --hard $(FABRIC_VERSION)
+	rm -f $(FABRIC_PATCHED) $(FABRIC_PATCHED).tmp
+
+$(FABRIC_TLCC):
+	ln -sfn $$(/bin/pwd)/../tlcc $(FABRIC_TLCC)
+
+peer: patch $(FABRIC_PEER_BIN)
+
+$(FABRIC_PEER_BIN): $(FABRIC_TLCC)
+	cd $(FABRIC_PATH) && \
+	GO_TAGS=pluginsenabled $(MAKE) peer
+
+native: peer
+	cd $(FABRIC_PATH) && \
+	$(MAKE) -j orderer cryptogen configtxgen
+
+clean: clean-native clean-patch
+	rm -f $(FABRIC_TLCC)
+
+clean-peer:
+	rm -rf $(FABRIC_PEER_BIN)
+
+clean-native:
+	rm -rf $(FABRIC_PATH)/build/bin/
--- a/fabric/README.md
+++ b/fabric/README.md
@@ -1,0 +1,37 @@
+<!---
+Licensed under Creative Commons Attribution 4.0 International License
+https://creativecommons.org/licenses/by/4.0/
+--->
+
+# Build Fabric with FPC support
+
+To build the Fabric peer manually make sure you have built [tlcc_enclave](../tlcc_enclave) before.
+Then run the following command:
+
+    $ cd $FPC_PATH/fabric
+    $ make
+
+## Common errors
+
+### TLCC not built yet
+
+```
+# github.com/hyperledger/fabric/cmd/peer
+/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
+/usr/bin/ld: cannot find -ltl
+collect2: error: ld returned 1 exit status
+```
+
+Seems that the Trusted Ledger has not been built before.
+Try to run `pushd $FPC_PATH/tlcc_enclave; make; popd;` followed by `make` again.
+
+
+### Wrong Fabric version
+```
+Patch Fabric ...
+Aborting! Current branch is master but want v2.0.1-fpc
+Makefile:22: recipe for target '/project/src/github.com/hyperledger/fabric/.fpc_patched' failed
+```
+
+Seems that your Fabric is on the wrong branch.
+Try to run `pushd $FABRIC_PATH; git checkout tags/v2.0.1 -b v2.0.1-fpc; popd;` followed by `make` again.
--- a/fabric/bin/peer.sh
+++ b/fabric/bin/peer.sh
@@ -378,8 +378,8 @@ handle_channel_join() {
     # - ask tlcc to join channel
     #   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
     #   it will re-join old channels but tlcc will not!
-    #say "Attaching TLCC to channel '${CHAN_ID}' ..." # TODO: Enable once tlcc is integrated.....
-    #try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID} # TODO: Enable once tlcc is integrated
+    say "Attaching TLCC to channel '${CHAN_ID}' ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID}
 
     # - exit (otherwise main function will invoke operation again!)
     exit 0
--- a/tlcc/Makefile
+++ b/tlcc/Makefile
@@ -6,15 +6,8 @@ TOP = ..
 include $(TOP)/build.mk
 
 LD_LIBRARY_PATH := LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$$(pwd)/enclave/lib
-SRC := tlcc.go
-OUTPUT := tlcc.so
 
-$(OUTPUT): $(SRC)
-	$(LD_LIBRARY_PATH) $(GO) build -o $@ -buildmode=plugin $<
-
-build: $(OUTPUT)
-
-test: build
+test:
 	$(LD_LIBRARY_PATH) $(GO) $@ -test.v ./enclave
 
 clean:
--- a/tlcc/README.md
+++ b/tlcc/README.md
@@ -7,37 +7,16 @@ https://creativecommons.org/licenses/by/4.0/
 Before your continue here make sure you have built ``tlcc_enclave``.
 We refer to [tlcc_enclave/README.md](../tlcc_enclave).
 
-First we build tlcc as system chaincode plugin. What is that?
-See [here](https://hyperledger-fabric.readthedocs.io/en/release-1.4/systemchaincode.html).
-
-    $ make
-
-This builds `tlcc.so`. Using this plugin when running the peer inside
-Docker most probably will not work out-of-the-box, thus, not supported
-right now.
-
 ## Integrate with Fabric
 
-Add `tlcc` as system chaincode plugin to your `core.yaml`. Example:
-
-```
-chaincode:
-    system:
-        tlcc: enable
-
-    systemPlugins:
-        - enabled: true
-        name: tlcc
-        path: /path-to/fabric-private-chaincode/tlcc/tlcc.so
-        invokableExternal: true
-        invokableCC2CC: true
-```
+TLCC is integrated into a fabric peer as system chaincode and thus must be built into the peer binary.
+We provide a peer target in [fabric/](../fabric) that builds the peer with tlcc integration.
 
 ## Starting the peer
 
 When starting the peer make sure that `LD_LIBRARY_PATH` points to the enclave lib.
 
-    $ LD_LIBRARY_PATH=/path-to/fabric-private-chaincode/tlcc/enclave/lib build/bin/peer node start
+    $ LD_LIBRARY_PATH=$FPC_PATH/tlcc/enclave/lib build/bin/peer node start
 
 ## Join the channel
 
--- a/tlcc_enclave/Makefile
+++ b/tlcc_enclave/Makefile
@@ -6,12 +6,13 @@ TOP = ..
 include $(TOP)/build.mk
 
 PROTO_BUILD_DIR := enclave/protos
+FABRIC_PROTOS := protos/fabric
 BUILD_DIR := _build
 
 .PHONY: protos
 protos: $(PROTO_BUILD_DIR)
 
-$(PROTO_BUILD_DIR):
+$(PROTO_BUILD_DIR): $(FABRIC_PROTOS)
 	./generate_protos.sh $@
 
 .PHONY: protos-clean
--- a/tlcc_enclave/Makefile
+++ b/tlcc_enclave/Makefile
@@ -6,13 +6,12 @@ TOP = ..
 include $(TOP)/build.mk
 
 PROTO_BUILD_DIR := enclave/protos
-FABRIC_PROTOS := protos/fabric
 BUILD_DIR := _build
 
 .PHONY: protos
 protos: $(PROTO_BUILD_DIR)
 
-$(PROTO_BUILD_DIR): $(FABRIC_PROTOS)
+$(PROTO_BUILD_DIR):
 	./generate_protos.sh $@
 
 .PHONY: protos-clean
--- a/tlcc_enclave/README.md
+++ b/tlcc_enclave/README.md
@@ -16,9 +16,8 @@ the blockchain state.
 ## Protos
 
 We use nanopb for proto buffers inside the trusted ledger enclave.
-We can generate the proto files by using ``make protos``. Check that
-you export the environment variables `NANOPB_PATH` and
-`FABRIC_PATH` pointing to Fabric source.
+We can generate the proto files by using ``make protos``. Fabric proto files
+are located in `protos/fabric` and fetched from https://github.com/hyperledger/fabric-protos.
 
 ## Build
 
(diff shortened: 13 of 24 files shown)

### CODE REVIEW

**Comment by mbrandenburger** [comment]
Current build fails because in order to build the peer with tlcc support we need to run build tlcc_enclave first. ARGH.

**Comment by mbrandenburger** [comment]
Ouch we have a protobuf issue here with tlcc.
```
root@7dbf8e13a678:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave# make
./generate_protos.sh enclave/protos
find: '/project/src/github.com/hyperledger/fabric/protos': No such file or directory
find: '/project/src/github.com/hyperledger/fabric/protos': No such file or directory
find: '/project/src/github.com/hyperledger/fabric/protos': No such file or directory
find: '/project/src/github.com/hyperledger/fabric/protos': No such file or directory
find: '/project/src/github.com/hyperledger/fabric/protos': No such file or directory
sed: can't read enclave/protos/ledger/rwset/rwset.pb.h: No such file or directory
sed: can't read enclave/protos/ledger/rwset/rwset.pb.c: No such file or directory
```

**Comment by g2flyer** [comment]
> Ouch we have a protobuf issue here with tlcc.

Do you see something more fundamental than just updating the path in generate_protos to the new proto github repo (and updating docu/docker/travis to also checkout that repo)? (The latter of course is some work, but not any complex hurdle to cross)

File: .travis.yml [hunk 1]
```diff
@@ -65,6 +65,7 @@ before_install:
   - git clone --branch $FABRIC_VERSION https://github.com/hyperledger/fabric.git $FABRIC_PATH
   - pushd $FABRIC_PATH
   - git am $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/*.patch
+  - ln -s $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc $GOPATH/src/github.com/hyperledger/fabric/core/scc/tlcc
```
MHBauer, line 68: so this must be done here, because tlcc is compiled into the peer?

fabric/0001-Integrate-tlcc-syscc.patch
```diff
@@ -0,0 +1,47 @@
...
+
+diff --git a/internal/peer/node/start.go b/internal/peer/node/start.go
+index 340d6da26..5fd9ebba3 100644
+--- a/internal/peer/node/start.go
++++ b/internal/peer/node/start.go
+@@ -74,6 +74,7 @@ import (
+ 	"github.com/hyperledger/fabric/core/scc/cscc"
+ 	"github.com/hyperledger/fabric/core/scc/lscc"
+ 	"github.com/hyperledger/fabric/core/scc/qscc"
++	tlcc "github.com/hyperledger/fabric/core/scc/tlcc"
```
MHBauer, line 21: kind of weird it gets imported like this.

fabric/bin/peer.sh
```diff
@@ -254,6 +254,12 @@ handle_channel_join() {
     # - call real peer so channel is joined
     try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
 
+    # - ask tlcc to join channel
+    #   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
+    #   it will re-join old channels but tlcc will not!
+    say "Attaching TLCC to channel '${CHAN_ID}' ..."
+    try $RUN ${FABRIC_BIN_DIR}/peer chaincode query -n tlcc -c '{"Args": ["JOIN_CHANNEL"]}' -C ${CHAN_ID}
```
MHBauer, line 261: query causes a join?

tlcc/README.md
```diff
@@ -7,37 +7,22 @@ https://creativecommons.org/licenses/by/4.0/
...
 ## Integrate with Fabric
 
-Add `tlcc` as system chaincode plugin to your `core.yaml`. Example:
-
-```
-chaincode:
-    system:
-        tlcc: enable
+TLCC is integrated into a fabric peer as system chaincode and thus must be build together with the peer binary.
+We provide a simple Fabric for the integration that can be used as follows:
```
MHBauer, line 13: Question answered.

.travis.yml
g2flyer: i guess. But the more problematic issue is that right now it doesn't work (see travis build): either the shell ignores the symlink and thinks it is in the fabric tree but then the relative paths to other fpc code fails or it knows about symlink the relative links to fpc work but then the whole reason to run it from fabric would be kindof moot and/or bring gopath problems?

fabric/0001-Integrate-tlcc-syscc.patch
g2flyer: shouldn't it also work without the `tlcc` prefix?

fabric/bin/peer.sh
g2flyer: curious, why did you change order of ercc and tlcc add/init?

fabric/bin/peer.sh
```diff
@@ -254,6 +254,12 @@ handle_channel_join() {
     # - call real peer so channel is joined
     try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
 
+    # - ask tlcc to join channel
+    #   IMPORTANT: right now a join is _not_ persistant, so on restart of peer,
+    #   it will re-join old channels but tlcc will not!
```
g2flyer, line 259: given that you already moved it (see below) and shows up as a diff for the whole section: should we maybe extend the comment that (presumably) manually calling the JOIN_CHANNEL after a peer restart presumably should make it work, though?

tlcc/README.md
```diff
@@ -7,37 +7,22 @@ https://creativecommons.org/licenses/by/4.0/
...
 
-    systemPlugins:
-        - enabled: true
-        name: tlcc
-        path: /path-to/fabric-private-chaincode/tlcc/tlcc.so
-        invokableExternal: true
-        invokableCC2CC: true
-```
+    $ cd $FABRIC_PATH
+    $ git am $FPC_PATH/fabric/0001-Integrate-tlcc-syscc.patch
```
g2flyer, line 16: Hmm, don't we do elsehwere already a git am $FPC_PATH/fabric/*, so this should fail here?

mbrandenburger: right - I just wanted to document that tlcc integration relies on that patch! Let me try to make this more clear

fabric/0001-Integrate-tlcc-syscc.patch
mbrandenburger: right! will change

fabric/bin/peer.sh
mbrandenburger: tlcc should first join the channel and thereby register for new blocks/transactions. So this must happen as first action of a FPC Peer when joining a channel; otherwise tlcc will not see any transactions related to ercc lifecycle/installation. Indeed, those lifecycle transactions are currently not really processed within tlcc.

mbrandenburger: Does this make sense?

mbrandenburger:
> > Ouch we have a protobuf issue here with tlcc.
> 
> Do you see something more fundamental than just updating the path in generate_protos to the new proto github repo (and updating docu/docker/travis to also checkout that repo)? (The latter of course is some work, but not any complex hurdle to cross)

- getting the protos from the new repo is easy
- new protos use reserve keyword (https://github.com/hyperledger/fabric-protos/blob/master/common/common.proto#L28) which is not supported with protobuf 3.0.x but requires protobuf 3.5.x or higher; it seems we need to upgrade protobuf inside travis and
- small additional edit on the nanopb options file for new lifecycle protos needed

fabric/bin/peer.sh
g2flyer: Hmm, i thought we always read from the genesis block, so it wouldn't matter unless ercc requires tlcc?  If what you say is indeed true, it would also mean that what we claim that we currently support peer restart would be wrong?!

g2flyer:
> > > Ouch we have a protobuf issue here with tlcc.
> > 
> > 
> > Do you see something more fundamental than just updating the path in generate_protos to the new proto github repo (and updating docu/docker/travis to also checkout that repo)? (The latter of course is some work, but not any complex hurdle to cross)
> 
> * getting the protos from the new repo is easy
> * new protos use reserve keyword (https://github.com/hyperledger/fabric-protos/blob/master/common/common.proto#L28) which is not supported with protobuf 3.0.x but requires protobuf 3.5.x or higher; it seems we need to upgrade protobuf inside travis and
> * small additional edit on the nanopb options file for new lifecycle protos needed

That means a new version of nanopb?  If so, didn't @bvavala look for other reasons also on newer version of nanopb (or potentially standard google protobuf as alternatives?)

mbrandenburger:
> > > > Ouch we have a protobuf issue here with tlcc.
> > > 
> > > 
> > > Do you see something more fundamental than just updating the path in generate_protos to the new proto github repo (and updating docu/docker/travis to also checkout that repo)? (The latter of course is some work, but not any complex hurdle to cross)
> > 
> > 
> > 
> > * getting the protos from the new repo is easy
> > * new protos use reserve keyword (https://github.com/hyperledger/fabric-protos/blob/master/common/common.proto#L28) which is not supported with protobuf 3.0.x but requires protobuf 3.5.x or higher; it seems we need to upgrade protobuf inside travis and
> > * small additional edit on the nanopb options file for new lifecycle protos needed
> 
> That means a new version of nanopb? If so, didn't @bvavala look for other reasons also on newer version of nanopb (or potentially standard google protobuf as alternatives?)

the nanopb version we are using seem to work fine. At least I did not see any issues here with, thus, in this PR I would not update nanopb. Switching to google protobuf I see in a different PR as it will also touch tlcc implementation.

fabric/bin/peer.sh
mbrandenburger: let me double check -

tlcc/Makefile
```diff
@@ -6,15 +6,25 @@ TOP = ..
...
-$(OUTPUT): $(SRC)
-	$(LD_LIBRARY_PATH) $(GO) build -o $@ -buildmode=plugin $<
+FABRIC_TLCC = $(FABRIC_PATH)/core/scc/tlcc
+FABRIC_PEER_BIN = $(FABRIC_PATH)/build/bin/peer
 
-build: $(OUTPUT)
+build: fabric-peer
 
-test: build
+fabric-peer: $(FABRIC_PEER_BIN)
```
bvavala, line 15: This does not belong here. Any reason to have the peer built in our build process?
Besides, the peer patch is documented but not applied here -- it's applied in travis and docker.

I see that it could be convenient to have this step in the build process. In this case it should probably be an `fpc-peer` target in the top-level makefile, which also applies all the patches we have -- and we should remove this step in travis and docker.

Additionally, to make sure the peer is right, we might add a capability check in tlcc and tests to test whether a query to tlcc fails -- oth. it will badly fail later when it does not find the system chaincode.

MHBauer:
rebased to make sure I didn't accidentally blow it up with tools.go

.travis.yml
```diff
@@ -15,6 +15,7 @@ services:
 addons:
   apt:
     packages:
+      # TODO upgrade to protobuf 3.5 or higher
```
MHBauer, line 18: todo complete as of below using protobuf 3.11.4?

.travis.yml
```diff
@@ -61,11 +66,13 @@ before_install:
...
   - git clone https://github.com/nanopb/nanopb.git ${NANOPB_PATH}
   - pushd ${NANOPB_PATH} && git checkout nanopb-${NANOPB_VERSION} && cd generator/proto && make && popd
+  # Newer Protobuf
+  - wget ${PROTO_REPO}/v${PROTO_VERSION}/${PROTO_ZIP}; unzip ${PROTO_ZIP} -d protoc3; sudo mv protoc3/bin/* /usr/local/bin/; sudo mv protoc3/include/* /usr/local/include/
   # Fabric
   - git clone --branch $FABRIC_VERSION https://github.com/hyperledger/fabric.git $FABRIC_PATH
   - pushd $FABRIC_PATH
   - git am $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/*.patch
-  - GO_TAGS=pluginsenabled make peer orderer cryptogen configtxgen && popd
+  - make orderer cryptogen configtxgen && popd
```
MHBauer, line 75: so now is peer happening in the make process below?

fabric/0001-Integrate-tlcc-syscc.patch
```diff
@@ -0,0 +1,47 @@
...
++
+ 	// deploy system chaincodes
+-	for _, cc := range []scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC} {
++	for _, cc := range []scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC, tlccInst} {
+ 		if enabled, ok := chaincodeConfig.SCCWhitelist[cc.Name()]; !ok || !enabled {
+ 			logger.Infof("not deploying chaincode %s as it is not enabled", cc.Name())
+ 			continue
+-- 
+2.26.0
+
```
MHBauer, line 47: it's really weird looking at a diff in the diff-looker.

fabric/bin/peer.sh
MHBauer: Still confused about this, maybe a design issue to resolve later. I would think we need to write the tlcc config to the chain to trust it.

Not confused about the order. Agree that this means peer restart does not work unless the txn is re-run.

tlcc/Makefile
MHBauer: I can go either way on this.

On a principal of owning what we're doing, we should build the peer as we modify it with a patch.

On the other hand, someone running this does need to know to manually swap the peer binary out.

So by automating this step are we fooling later users into skipping a step? It is something which affirmatively needs to be done manually.

Until there are outside users, I think automating the process is helpful, wherever it occurs.

tlcc/README.md
```diff
@@ -7,37 +7,22 @@ https://creativecommons.org/licenses/by/4.0/
...
-
 ## Integrate with Fabric
 
-Add `tlcc` as system chaincode plugin to your `core.yaml`. Example:
-
-```
-chaincode:
-    system:
-        tlcc: enable
+TLCC is integrated into a fabric peer as system chaincode and thus must be build together with the peer binary.
```
MHBauer, line 12: `build` -> `built` 

alternatively, "must be built into the peer binary. This means the peer must be compiled and the resulting binaries replace those from a standard fabric distribution."

Feel free to keep as much or as little of this as you want. I can make the typo change barring objections.

tlcc/README.md
MHBauer: Maybe seems to need some resolution. I supposed the "can be used" wording above implies it is not necessary to do. 

However, people will see terminal commands and blindly copy them. It does help that this is made difficult by having the leading `$`.

tlcc_enclave/Makefile
```diff
@@ -6,12 +6,16 @@ TOP = ..
...
 PROTO_BUILD_DIR := enclave/protos
+FABRIC_PROTOS := protos/fabric
 BUILD_DIR := _build
 
 .PHONY: protos
 protos: $(PROTO_BUILD_DIR)
 
-$(PROTO_BUILD_DIR):
+$(FABRIC_PROTOS):
+	git clone https://github.com/hyperledger/fabric-protos.git $@
```
MHBauer, line 16: I hesitantly suggest fabric-protos as a submodule, but if this works, I'm fine with it.

tlcc_enclave/.gitignore
```diff
@@ -4,3 +4,4 @@ enclave/protos
 tags
 enclave/*_t.*
 trusted_ledger/*_u.*
+protos/fabric
```
MHBauer, line 7: NIT: a comment of "to ignore the cloned fabric-protos repo"

utils/docker/dev/Dockerfile
```diff
@@ -65,8 +66,21 @@ RUN git clone https://github.com/nanopb/nanopb.git $NANOPB_PATH \
  && cd generator/proto \
  && make
 
+# Install new protobuf
+# TODO having this here helps indeed to build new protos but may not be compatible with SGX SDK
+# That is, newer protobuf should be installed in base image but
+ARG PROTO_VERSION=3.11.4
+ARG PROTO_ZIP=protoc-${PROTO_VERSION}-linux-x86_64.zip
```
MHBauer, line 73: I am highly amused that this is provided as a zip.

tlcc/Makefile
```diff
@@ -6,15 +6,25 @@ TOP = ..
...
+
+$(FABRIC_PEER_BIN): $(FABRIC_TLCC)
+	cd $(FABRIC_PATH) && \
+	GO_TAGS=pluginsenabled $(MAKE) peer
```
MHBauer, line 19: peer build moved here!

utils/docker/peer/Dockerfile
```diff
@@ -90,11 +90,13 @@ RUN cd $FABRIC_PATH \
...
  && git am $FPC_PATH/fabric/*patch \
  && git config --unset user.email \
- && GO_TAGS=pluginsenabled make peer orderer cryptogen configtxgen
+ && make orderer cryptogen configtxgen
 
 # Build FPC plugins for peer
 RUN cd $FPC_PATH \
  && make plugins \
+ && make tlcc_enclave \
+ && make tlcc \
```
MHBauer, line 99: is there a reason to make these individually vs all in one invocation (and maybe with `-j`) ?

tlcc_enclave/protos/fabric.options
```diff
@@ -102,3 +105,6 @@ kvrwset.KVWrite.key type:FT_POINTER
 kvrwset.KVWrite.value type:FT_POINTER
 
 kvrwset.KVRead.key type:FT_POINTER
+
+lifecycle.StateData.Bytes type:FT_POINTER
+lifecycle.StateData.String type:FT_POINTER
```
MHBauer, line 110: I don't know what to think of any of this file change.

g2flyer:
ok, did now some quick tests, but while ledger_init did work, the commit of helloworld failed with
```
20-04-09 14:05:28.115 PDT [grpc] createTransport -> DEBU 025 grpc: addrConn.createTransport failed t
o connect to {127.0.0.1:7051 0  <nil>}. Err :connection error: desc = "transport: error while dialing:
 dial tcp 127.0.0.1:7051: connect: connection refused". Reconnecting...
2020-04-09 14:05:28.116 PDT [grpc] HandleSubConnStateChange -> DEBU 026 pickfirstBalancer: HandleSubCo
nnStateChange: 0xc000453340, TRANSIENT_FAILURE
Error: error getting endorser client for query: endorser client failed to connect to 127.0.0.1:7051: f
ailed to create new connection: connection error: desc = "transport: error while dialing: dial tcp 127
.0.0.1:7051: connect: connection refused"
+ die 'test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledge
r/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":["__getEncl
avePk"]}'
++ basename ./fabric/bin/peer.sh
+ recho 'peer.sh: test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com
/hyperledger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":
["__getEnclavePk"]}'
+ echo 'peer.sh:' test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com
/hyperledger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c '{"Args"
:["__getEnclavePk"]}'
peer.sh: test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperled
ger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":["__getEn
clavePk"]}
+ exit 111

```
and /tmp/hyperledger/test/peer.err shows a tlcc crash
```
rusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:560) [Enclave] Ledger:                     \-> key = _lifecycle.namespaces/fields/hello/Sequence version = blockNum: 5; txNum: 0
2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8d DEBUG  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:560) [Enclave] Ledger:                     \-> key = hello.MRENCLAVE version = blockNum: 0; txNum: 0
2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8e DEBUG  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:582) [Enclave] Ledger:             \-> Writes:
2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8f ERROR  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:625) [Enclave] Ledger: >>> NO MRENCLAVE found for: ^T.MRENCLAVE
fatal: morestack on g0
SIGTRAP: trace trap
PC=0xae1872 m=4 sigcode=128

goroutine 0 [idle]:
runtime.abort()
        /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:859 +0x2
runtime.morestack()
        /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:416 +0x25

goroutine 175 [syscall]:
runtime.cgocall(0x1576c80, 0xc0029e7f28, 0x17ed340)
        /home/msteiner/go/go1.14.linux-amd64/src/runtime/cgocall.go:133 +0x5b fp=0xc0029e7ef8 sp=0xc0029e7ec0 pc=0xa7d03b
github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave._C2func_tlcc_send_block(0x1bd500000002, 0x7f48380154d0, 0x1227, 0x7f4800000000, 0x0, 0x0)
        _cgo_gotypes.go:86 +0x55 fp=0xc0029e7f28 sp=0xc0029e7ef8 pc=0x1547615
github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave.(*StubImpl).NextBlock(0xc0002e0c08, 0xc002e6c000, 0x1227, 0x1227, 0x0, 0x0)
        /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave/enclave_stub.go:132 +0xa2 fp=0xc0029e7f80 sp=0xc0029e7f28 pc=0x15487e2
github.com/hyperledger/fabric/core/scc/tlcc.(*TrustedLedgerCC).readBlocks(0xc0003bb500, 0x1b4ab60, 0xc0003baea0)
        /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/core/scc/tlcc/tlcc.go:148 +0xa7 fp=0xc0029e7fc8 sp=0xc0029e7f80 pc=0x154b787
runtime.goexit()
        /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:1373 +0x1 fp=0xc0029e7fd0 sp=0xc0029e7fc8 pc=0xae1f61
created by github.com/hyperledger/fabric/core/scc/tlcc.(*TrustedLedgerCC).joinChannel
        /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/core/scc/tlcc/tlcc.go:133 +0x28c

goroutine 1 [chan receive, 49 minutes]:
github.com/hyperledger/fabric/internal/peer/node.serve(0x277e610, 0x0, 0x0, 0x0, 0x0)
        /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/internal/peer/node/start.go:863 +0x4aa4
github.com/hyperledger/fabric/internal/peer/node.glob..func6(0x264c1c0, 0x277e610, 0x0, 0x0, 0x0, 0x0)
        /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/in
```

tlcc/Makefile
mbrandenburger: I am fine with calling it fpc-peer.

At least we need this target for us to build the peer for our integration tests. If you give this to the end users we can discuss. Would it make more sense to put the patched fpc peer binary somewhere in the fpc folder?! $FPC_PATH/fabric/build/bin ??

bvavala: The target name is important but secondary.
Yes, we need it for the integration tests, because the we run those in the concept release.

The main points are "where" it is and "what" it does.
The travis file answers to this https://github.com/hyperledger-labs/fabric-private-chaincode/blob/169cfe756de05d1495aef8175a831cff149744fe/.travis.yml#L67.
What we do in travis is to: (1) apply the patches, (2) recompile the peer, (3) run make (with the integration tests).
So if you create an `fpc-peer` target in the top-level makefile:
* you can move (1) and (2) from travis to the new target
* you automate the procedure for the end user, because now the user still has to apply the patches manually whenever he starts from scratch
* you can make sure you build the "patched" peer  -- in the current form you might build the unpatched peer and only discover later that tlcc is not there.

mbrandenburger: Sounds good to me

tlcc_enclave/generate_protos.sh
```diff
@@ -26,20 +26,19 @@ rm -rf $BUILD_DIR
 mkdir -p $BUILD_DIR
 
 # compile google protos (timestamp)
-$(protoc "$PROTOC_OPTS" --proto_path="protos" --nanopb_out=$BUILD_DIR protos/google/protobuf/*.proto)
+protoc "$PROTOC_OPTS" --proto_path="protos" --nanopb_out=$BUILD_DIR protos/google/protobuf/*.proto
```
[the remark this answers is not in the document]
mbrandenburger: I like that idea ... let me try this

.travis.yml
mbrandenburger: peer patching and compilation moved into the `fabric/Makefile`

fabric/bin/peer.sh
mbrandenburger: aha! @MHBauer I think you refer to the MRENCLAVE of TLCC itself. How do we get this into TLCC; this critical feature is not implemented yet.

fabric/bin/peer.sh
mbrandenburger: moved it back where it was

tlcc/README.md
mbrandenburger: I revised all readme files

tlcc/Makefile
mbrandenburger: I created this peer target in `fabric/Makefile`. It patches with peer with all patches we have and compiles it; moreover, we have a target to build the other fabric binaries (orderer, confitxgen ....;

.travis.yml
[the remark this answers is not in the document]
mbrandenburger: removed

utils/docker/dev/Dockerfile
```diff
@@ -65,8 +66,21 @@ RUN git clone https://github.com/nanopb/nanopb.git $NANOPB_PATH \
  && cd generator/proto \
  && make
 
+# Install new protobuf
```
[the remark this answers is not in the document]
mbrandenburger: I moved it to base but actually only in dev it is needed; currently the peer image also builds the peer but ideally we could build the peer using the dev container and just copy the peer binary into the peer container; with the current docker layout it seems that dev and peer images are almost identical

tlcc/Makefile
mbrandenburger: > ok, did now some quick tests, but while ledger_init did work, the commit of helloworld failed with
> 
> ```
> 20-04-09 14:05:28.115 PDT [grpc] createTransport -> DEBU 025 grpc: addrConn.createTransport failed t
> o connect to {127.0.0.1:7051 0  <nil>}. Err :connection error: desc = "transport: error while dialing:
>  dial tcp 127.0.0.1:7051: connect: connection refused". Reconnecting...
> 2020-04-09 14:05:28.116 PDT [grpc] HandleSubConnStateChange -> DEBU 026 pickfirstBalancer: HandleSubCo
> nnStateChange: 0xc000453340, TRANSIENT_FAILURE
> Error: error getting endorser client for query: endorser client failed to connect to 127.0.0.1:7051: f
> ailed to create new connection: connection error: desc = "transport: error while dialing: dial tcp 127
> .0.0.1:7051: connect: connection refused"
> + die 'test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledge
> r/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":["__getEncl
> avePk"]}'
> ++ basename ./fabric/bin/peer.sh
> + recho 'peer.sh: test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com
> /hyperledger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":
> ["__getEnclavePk"]}'
> + echo 'peer.sh:' test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com
> /hyperledger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c '{"Args"
> :["__getEnclavePk"]}'
> peer.sh: test failed: /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperled
> ger/fabric/build/bin/peer chaincode query -o localhost:7050 -C mychannel -n hello -c {"Args":["__getEn
> clavePk"]}
> + exit 111
> ```
> 
> and /tmp/hyperledger/test/peer.err shows a tlcc crash
> 
> ```
> rusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:560) [Enclave] Ledger:                     \-> key = _lifecycle.namespaces/fields/hello/Sequence version = blockNum: 5; txNum: 0
> 2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8d DEBUG  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:560) [Enclave] Ledger:                     \-> key = hello.MRENCLAVE version = blockNum: 0; txNum: 0
> 2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8e DEBUG  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:582) [Enclave] Ledger:             \-> Writes:
> 2020-04-09 14:05:28.035 PDT [tl-enclave] golog -> INFO 2e8f ERROR  (/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc_enclave/enclave/ledger.cpp:625) [Enclave] Ledger: >>> NO MRENCLAVE found for: ^T.MRENCLAVE
> fatal: morestack on g0
> SIGTRAP: trace trap
> PC=0xae1872 m=4 sigcode=128
> 
> goroutine 0 [idle]:
> runtime.abort()
>         /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:859 +0x2
> runtime.morestack()
>         /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:416 +0x25
> 
> goroutine 175 [syscall]:
> runtime.cgocall(0x1576c80, 0xc0029e7f28, 0x17ed340)
>         /home/msteiner/go/go1.14.linux-amd64/src/runtime/cgocall.go:133 +0x5b fp=0xc0029e7ef8 sp=0xc0029e7ec0 pc=0xa7d03b
> github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave._C2func_tlcc_send_block(0x1bd500000002, 0x7f48380154d0, 0x1227, 0x7f4800000000, 0x0, 0x0)
>         _cgo_gotypes.go:86 +0x55 fp=0xc0029e7f28 sp=0xc0029e7ef8 pc=0x1547615
> github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave.(*StubImpl).NextBlock(0xc0002e0c08, 0xc002e6c000, 0x1227, 0x1227, 0x0, 0x0)
>         /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave/enclave_stub.go:132 +0xa2 fp=0xc0029e7f80 sp=0xc0029e7f28 pc=0x15487e2
> github.com/hyperledger/fabric/core/scc/tlcc.(*TrustedLedgerCC).readBlocks(0xc0003bb500, 0x1b4ab60, 0xc0003baea0)
>         /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/core/scc/tlcc/tlcc.go:148 +0xa7 fp=0xc0029e7fc8 sp=0xc0029e7f80 pc=0x154b787
> runtime.goexit()
>         /home/msteiner/go/go1.14.linux-amd64/src/runtime/asm_amd64.s:1373 +0x1 fp=0xc0029e7fd0 sp=0xc0029e7fc8 pc=0xae1f61
> created by github.com/hyperledger/fabric/core/scc/tlcc.(*TrustedLedgerCC).joinChannel
>         /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/core/scc/tlcc/tlcc.go:133 +0x28c
> 
> goroutine 1 [chan receive, 49 minutes]:
> github.com/hyperledger/fabric/internal/peer/node.serve(0x277e610, 0x0, 0x0, 0x0, 0x0)
>         /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/internal/peer/node/start.go:863 +0x4aa4
> github.com/hyperledger/fabric/internal/peer/node.glob..func6(0x264c1c0, 0x277e610, 0x0, 0x0, 0x0, 0x0)
>         /home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger/fabric/in
> ```

now we integration tests enabled, travis also produces this error.

mbrandenburger:
TLCC crashes because it tried to read MRENCLAVE from a readset when validating a ercc transaction; however, with PR #330 we removed init that writes MRENCLAVE to the ledger. I propose to move it for now to the _setup as we will fetch MRENCLAVE from the chaincode definition.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "tlcc_enclave/protos/fabric"]
+	path = tlcc_enclave/protos/fabric
+	url = https://github.com/hyperledger/fabric-protos.git
+	branch = release-2.0
```
MHBauer, line 4: ooohhhh, module

.travis.yml
```diff
@@ -61,11 +70,10 @@ before_install:
   # NANOPB
   - git clone https://github.com/nanopb/nanopb.git ${NANOPB_PATH}
   - pushd ${NANOPB_PATH} && git checkout nanopb-${NANOPB_VERSION} && cd generator/proto && make && popd
+  # Newer Protobuf
+  - wget ${PROTO_REPO}/v${PROTO_VERSION}/${PROTO_ZIP}; sudo unzip ${PROTO_ZIP} -d ${PROTO_DIR}
```
MHBauer, line 74: sudo?

README.md
```diff
@@ -301,40 +301,24 @@ Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Ch
 
 Clone the code and make sure it is on your `$GOPATH`. (Important: we assume in this documentation and default configuration that your `$GOPATH` has a _single_ root-directoy!)
 
-    $ git clone https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
+    $ git clone --recursive https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
```
MHBauer, line 304: :+1: --recursive for modules

README.md
```diff
@@ -301,40 +301,24 @@ Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Ch
...
-    $ git am ../../hyperledger-labs/fabric-private-chaincode/fabric/*.patch
     
-Note that this patch does currently not work with the Fabric master branch, therefore make sure you use the Fabric
-v2.0.1 branch.
+Note that Fabric Private Chaincode currently not work with the Fabric master branch, therefore make sure you use the Fabric
+v2.0.1 branch. Make sure Fabric is in your ``$GOPATH``.
 
 ### Build Fabric Private Chaincode
 
 Once you have your development environment up and running (i.e., using our docker-based setup or install all all dependencies on your machine) you can build FPC and start developing your own FPC application.
```
MHBauer, line 320: repeated all, "all all"

config.mk
```diff
@@ -15,6 +15,8 @@ export SGX_BUILD ?= PRERELEASE
 
 export FABRIC_PATH ?= ${GOPATH}/src/github.com/hyperledger/fabric
 
+export PROTOC_CMD ?= protoc
```
MHBauer, line 18: overloaded by travis config.

fabric/REAME.md
```diff
@@ -0,0 +1,18 @@
+<!---
```
MHBauer, line 1: name of file, missing D, REAME vs README

tlcc/README.md
```diff
@@ -7,37 +7,16 @@ https://creativecommons.org/licenses/by/4.0/
...
-        tlcc: enable
-
-    systemPlugins:
-        - enabled: true
-        name: tlcc
-        path: /path-to/fabric-private-chaincode/tlcc/tlcc.so
-        invokableExternal: true
-        invokableCC2CC: true
-```
+TLCC is integrated into a fabric peer as system chaincode and thus must be build together with the peer binary.
```
MHBauer, line 12: build -> built

I favor "built into the peer binary".

tlcc_enclave/protos/fabric.options
MHBauer: still same.

utils/docker/Makefile
```diff
@@ -23,7 +23,7 @@ DOCKER_DEV_BUILD_OPTS ?=
 ifdef DOCKER_DEV_IMAGE_APT_ADD__PKGS
 	DOCKER_DEV_BUILD_OPTS += --build-arg APT_ADD_PKGS=$(DOCKER_DEV_IMAGE_APT_ADD__PKGS)
 endif
-
+DOCKER_DEV_BUILD_OPTS += --build-arg GIT_USER_NAME=$(shell git config user.name) --build-arg GIT_USER_EMAIL=$(shell git config user.email)
```
MHBauer, line 26: needed because we're doing the apply inside of docker, where git is unconfigured? 

kind of obvious, but could use a comment.

utils/docker/base/Dockerfile
```diff
@@ -47,3 +48,13 @@ ENV SGX_SDK=/opt/intel/sgxsdk
...
+
+# install custom protoc
+ARG PROTO_VERSION=3.11.4
+ARG PROTO_ZIP=protoc-${PROTO_VERSION}-linux-x86_64.zip
+ARG PROTO_REPO=https://github.com/google/protobuf/releases/download
+ARG PROTO_DIR=/usr/local/protoc3
+ENV PROTOC_CMD=${PROTO_DIR}/bin/protoc
+
+RUN wget ${PROTO_REPO}/v${PROTO_VERSION}/${PROTO_ZIP} \
+ && unzip ${PROTO_ZIP} -d ${PROTO_DIR}
```
MHBauer, line 60: Remember, we're limited to 125 layers. :-D

utils/docker/dev/Dockerfile
```diff
@@ -66,7 +72,7 @@ RUN git clone https://github.com/nanopb/nanopb.git $NANOPB_PATH \
  && make
 
 # Go tools we need
-RUN go get golang.org/x/tools/cmd/goimports 
+RUN go get golang.org/x/tools/cmd/goimports
```
MHBauer, line 75: Should be `go install` to pick up the requirements from module config.

.travis.yml
mbrandenburger: yap! cause we unzip to ${PROTO_DIR} = /usr/local/proto3 

maybe better location?

README.md
mbrandenburger: good catch!

utils/docker/base/Dockerfile
mbrandenburger: oh? never experience that limit! what is the best way to merge all those ARG and ENV layers?! :D

utils/docker/dev/Dockerfile
mbrandenburger: This line seems to be not needed at all. Let's remove it in a separate PR.

MHBauer: some discussion on the side happened. I retract this change request.

I'm wrong here because we're not in the fpc path when this is executed. This could be resolved by moving it below WORKDIR directive.
Additionally, as goimports is part of the make process, it will automatically be created during that part of the build. This can be resolved by totally removing the RUN here.

fabric/README.md
```diff
@@ -0,0 +1,18 @@
+<!---
+Licensed under Creative Commons Attribution 4.0 International License
+https://creativecommons.org/licenses/by/4.0/
+--->
+
+# Build Fabric with FPC support
+
+Make sure you have built [tlcc](../tlcc),[tlcc_enclave](../tlcc_enclave), and [plugins](../plugins)
```
g2flyer, line 8: Maybe mention that a top-level build is supposed to do this automatically. If we think have to do this manual, it might be worth adding a target for above list?

fabric/bin/peer.sh
g2flyer: > TLCC crashes because it tried to read MRENCLAVE from a readset when validating a ercc transaction; however, with PR #330 we removed init that writes MRENCLAVE to the ledger. I propose to move it for now to the _setup as we will fetch MRENCLAVE from the chaincode definition.

From what we talked in phoenix, the plan was to bake it into ercc code.  But then with the new protocols the bootstrapping will anyway also require changes in ecc and elsewhere.  In any case, you also mean `ecc`, not `ercc`.  In which case moving it to `__setup` seems a good idea ...

utils/docker/Makefile
```diff
@@ -23,7 +23,8 @@ DOCKER_DEV_BUILD_OPTS ?=
 ifdef DOCKER_DEV_IMAGE_APT_ADD__PKGS
 	DOCKER_DEV_BUILD_OPTS += --build-arg APT_ADD_PKGS=$(DOCKER_DEV_IMAGE_APT_ADD__PKGS)
 endif
-
+# needed because we're doing the apply of fabric patches inside of docker, where git is unconfigured
+DOCKER_DEV_BUILD_OPTS += --build-arg GIT_USER_NAME=$(shell git config user.name) --build-arg GIT_USER_EMAIL=$(shell git config user.email)
```
MHBauer, line 27: These need to be quoted. My name is "Morgan Bauer" and the space breaks this.

--build-arg GIT_USER_NAME="$(shell git config user.name)" --build-arg GIT_USER_EMAIL="$(shell git config user.email)"

tlcc/Makefile
```diff
@@ -6,15 +6,8 @@ TOP = ..
...
-SRC := tlcc.go
-OUTPUT := tlcc.so
 
-$(OUTPUT): $(SRC)
-	$(LD_LIBRARY_PATH) $(GO) build -o $@ -buildmode=plugin $<
-
-build: $(OUTPUT)
-
-test: build
+test:
```
g2flyer, line 10: test has no dependency?  That said, i guess i guess go test handles them all?

tlcc_enclave/Makefile
```diff
@@ -6,12 +6,13 @@ TOP = ..
...
 
 PROTO_BUILD_DIR := enclave/protos
+FABRIC_PROTOS := protos/fabric
 BUILD_DIR := _build
 
 .PHONY: protos
 protos: $(PROTO_BUILD_DIR)
 
-$(PROTO_BUILD_DIR):
+$(PROTO_BUILD_DIR): $(FABRIC_PROTOS)
```
g2flyer, line 15: hmm, wouldn't the dependency have to be rather on the actual proto files than the directory?  (e.g., something along the lines of `$(shell find tlcc_enclave/protos/fabric/ -name '*.proto')`?

utils/docker/Makefile
g2flyer: Needed but alas buggy: You have to quote these variables ....
```suggestion
DOCKER_DEV_BUILD_OPTS += --build-arg GIT_USER_NAME="$(shell git config user.name)" --build-arg GIT_USER_EMAIL="$(shell git config user.email)"
```

utils/docker/base/Dockerfile
```diff
@@ -47,3 +48,13 @@ ENV SGX_SDK=/opt/intel/sgxsdk
 ENV PATH=$PATH:$SGX_SDK/bin:$SGX_SDK/bin/x64
 ENV PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SGX_SDK/pkgconfig
 ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$SGX_SDK/sdk_libs
+
+# install custom protoc
+ARG PROTO_VERSION=3.11.4
+ARG PROTO_ZIP=protoc-${PROTO_VERSION}-linux-x86_64.zip
+ARG PROTO_REPO=https://github.com/google/protobuf/releases/download
+ARG PROTO_DIR=/usr/local/protoc3
```
g2flyer, line 56: Matter of taste but i think it's nicer to have all build-arg together at the beginning of a docker-file ....

utils/docker/peer/Dockerfile
mbrandenburger: I made it now all in one invocation; `-j` does not work here as tlcc depends on tlcc_enclave; so order matters here; though plugins and tlcc_enclave could be built in parallel

tlcc_enclave/Makefile
mbrandenburger: I add this check to `generate_protos.sh`.

tlcc/Makefile
mbrandenburger: yep, go test will build everything needed to run `tlcc_test.go`

tlcc_enclave/README.md
```diff
@@ -16,9 +16,8 @@ the blockchain state.
 ## Protos
 
 We use nanopb for proto buffers inside the trusted ledger enclave.
-We can generate the proto files by using ``make protos``. Check that
-you export the environment variables `NANOPB_PATH` and
```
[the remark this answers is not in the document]
mbrandenburger: ack - the sentence is back

fabric/Makefile
```diff
@@ -0,0 +1,48 @@
...
+$(FABRIC_TLCC):
+	ln -sfn $$(/bin/pwd)/../tlcc $(FABRIC_TLCC)
+
+peer: patch $(FABRIC_PEER_BIN)
+
+$(FABRIC_PEER_BIN): $(FABRIC_TLCC)
+	cd $(FABRIC_PATH) && \
+	GO_TAGS=pluginsenabled $(MAKE) peer
+
+native:
```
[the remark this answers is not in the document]
mbrandenburger: fabric uses the same name for the target see [here](https://github.com/hyperledger/fabric/blob/master/Makefile#L28).

fabric/Makefile
```diff
@@ -0,0 +1,48 @@
...
+patch: $(FABRIC_PATCHED)
+
+$(FABRIC_PATCHED):
+	cd $(FABRIC_PATH) && \
+	git am $(FPC_PATH)/fabric/*.patch > $@
+	cat $@
+
+clean-patch:
+	cd $(FABRIC_PATH) && \
+	git reset --hard $(FABRIC_VERSION)
```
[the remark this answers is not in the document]
mbrandenburger: I agree that this is a bit too aggressive. I propose the following:

- we checkout fabric v2.0.1 and create a branch called v2.0.1-fpc
- we only apply the patches on that branch
- we only apply the patches if HEAD points to tag v2.0.1
- we only perform cleaning if we are on that branch
- we remove clean-patch from clean target but keep it for local development

.travis.yml
mbrandenburger: This part of the main README.md will be very confusing ... in particular, we now need two different versions of protocol buffers.

1. protobuf 3.0.x to build the SGX SDK

1. protobuf 3.11.x or higher to build the fabric protos

However, the tricky part is that we build nanopb using the python bindings for 3.0.x ... argh ... need more testing to make this much easier.

tlcc/Makefile
MHBauer: I mean, in theory the dependency is all the go files, but good luck with making a makefile target for that mess. 

Best to let the go toolchain handle it!

fabric/Makefile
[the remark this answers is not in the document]
g2flyer: a few thoughts:
- dropping clean-patch gives you trouble when you add patches and/or switch between different versions. Doing as now is robust (though dangerous).
- to make the clean-branch less dangerous, we could actually two slightly different things than above
  - instead of (or maybe in addition to) branching on v2.0.1 we could add a branch (or tag?) _after_ applying the patches and make sure that we do a clean-branch only if we are on that branch, e.g., test with something along the lines of, ` if [ "fpc-patched-branch" == "$(git rev-parse --abbrev-ref HEAD)" ]; then .... fi`
  - make sure we do not have any modifications which get overriden, e.g., along the lines of `if [ -z "$(git status -s | grep -v '\?')" ]; then....  fi`

fabric/README.md
mbrandenburger: actually we only need to build `tlcc_enclave`. Updated

README.md
```diff
@@ -301,40 +301,24 @@ Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Ch
...
 
     $ export FABRIC_PATH=${GOPATH}/src/github.com/hyperledger/fabric
     $ git clone --branch v2.0.1 https://github.com/hyperledger/fabric.git $FABRIC_PATH
-    $ cd $FABRIC_PATH
-    $ git am ../../hyperledger-labs/fabric-private-chaincode/fabric/*.patch
     
-Note that this patch does currently not work with the Fabric master branch, therefore make sure you use the Fabric
-v2.0.1 branch.
+Note that Fabric Private Chaincode currently not work with the Fabric master branch, therefore make sure you use the Fabric
+v2.0.1 branch. Make sure Fabric is in your ``$GOPATH``.
```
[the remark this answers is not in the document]
mbrandenburger: please see new version

fabric/Makefile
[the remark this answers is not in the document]
g2flyer: actually, better idea instead of add branch/tag after applying patches, is simply to add the commit id to the (currently empty) patched-file and then read that in the clean and compare with the current commit. (obviously you still want to do the no-modification test ...)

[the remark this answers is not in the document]
g2flyer: I notice that you actually ask now interactively. This is robust but interaction during a makefile is not super elegant and might cause some potential trouble when make is run in scripts?

fabric/Makefile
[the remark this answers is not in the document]
g2flyer: not very intuitive. But then i guess it could be worse :-)

tlcc_enclave/Makefile
g2flyer: hmm, you missed my point: i thought you wanted to add this dependency so generate_protos.sh is always called when protobuf files have changed, not only the very first time when the proto_build_dir does not exist yet (as it will be without any dependency).

tlcc_enclave/enclave/ledger.cpp
```diff
@@ -611,6 +611,13 @@ int parse_endorser_transaction(
                         LOG_ERROR("Ledger: ercc expects only a single write ");
                         valid_tx = false;
                     }
+
+                    // TODO get MRENCLAVE from read/writeset to perform attestation validation once enabled
+                    // this code worked before as we wrote mrenclave during chaincode init; now we write mrenclave
+                    // during setup and thus mrenclave is part of fpc chaincode writeset. This code needs some
+                    // refactoring to resolve this issue; quick and dirty fix to prevent tlcc crashing, this code is
+                    // commented out
+                    /*
```
g2flyer, line 620: I'm pretty sure the problem that you do not properly identify fpc CC as following code on line 473 in ledger.cpp does not work anymore

            // currently private chaincodes need to have ecc prefix
            if (cc_name.compare(0, 3, "ecc") == 0)
            {
                ecc_namespace = cc_name;
            }
the external-builder version (which you approved ;-) actually does not rename chaincode anymore from foo to ecc_foo! I thought we used that only to remember in the wrapper which CC is fpc CC and forgot that also tlcc relied on this naming convention.
All that said, i think we should leave the commented-out code (although probably with more background as discussed here ...); this code should change sooner than later anyway and re-adding the renaming in the peer wrapper doesn't seem warranted, at least i would hope very much we can do that using this ugly naming convention. That said, in the back of the mind in the past i saw this as a non-problem going away with a proper chaincode type. However, now in v2.0 the chaincode type in the CC-definition does not exist anymore! So it's a good guestion how we are going to handle that. We might at some point revisit the protocol discussion from phoenix (hopefully on updated protocols diagrams reflecting that discussion!) and revisit from whatever we learned in the meantime about fabric v2.0 ...

(Btw: if we really want to keep prefixes, we might use `fpc_` or `fpc-c_` as it would encode the type rather a concept which the user might have no idea about?

tlcc_enclave/protos/fabric.options
MHBauer: so, this is a nanopb config of some sort.

tlcc_enclave/enclave/ledger.cpp
MHBauer: I defer on this point.

fabric/0001-Integrate-tlcc-syscc.patch
```diff
@@ -0,0 +1,47 @@
+From 0b29c14088c10e64ecaf918c9a96e5444d106e47 Mon Sep 17 00:00:00 2001
```
MHBauer, line 1: Thinking a small amount forward, modules plus a patch against upstream is probably difficult.

It might be possible with a vendor directory, but requires more investigation in the future.

fabric/Makefile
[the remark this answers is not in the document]
MHBauer: sounds like this mechanism still needs work.

.travis.yml
mbrandenburger: removed

tlcc_enclave/Makefile
mbrandenburger: we use modules now :D

tlcc_enclave/enclave/ledger.cpp
mbrandenburger: added more background as comment in the code.

fabric/Makefile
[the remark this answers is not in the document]
mbrandenburger: I added the modification test and check that patch commits are available on that branch.

I suggest that we stop making this is more robust and create a separate issue/PR for this as it seems to be beyond the scope if this PR.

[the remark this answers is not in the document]
MHBauer: I concur.

fabric/0001-Integrate-tlcc-syscc.patch
mbrandenburger: Good point! I think the patch is not a long term version, we may try to find / create better integrations points in the future.

Once modules in Fabric are upstream we may use `replace` to point to our local patched version?!

mbrandenburger:
Let's try to finish this one! @g2flyer, @MHBauer, @bvavala  please have a final look and check if you have happy with the way I worked in your feedback. Let's collect two green checkmarks on this PR, so I can rebase/squash and @MHBauer can merge

README.md
```diff
@@ -297,44 +299,38 @@ For more detailed information consult the official nanopb documentation http://g
 
 Make sure that you set `$NANOPB_PATH` as it is needed to build Fabric Private Chaincode.
 
+Moreover, in order to build Fabric protobufs we also require a newer Protobuf compiler than used to build the
+Intel SGX SDK. For this reason we recommend to get another installation and use it together with Nanopb.
```
g2flyer, line 303:
```suggestion
Moreover, in order to build Fabric protobufs we also require a newer Protobuf compiler than what is provided as standard ubuntu package and is used to build the
Intel SGX SDK. For this reason you will have to download and install another version and use it together with Nanopb. Do not install the new protobuf, though, such that it is not found in your standard PATH but instead define the `PROTOC_CMD`, either as environment variable or via `config.override.mk` to point to the new `protoc` binary
```

fabric/Makefile
```diff
@@ -0,0 +1,92 @@
...
+	@echo "Cleaning Fabric patches ..."
+	@if [ "$(FABRIC_CURRENT_BRANCH)" != "$(FABRIC_PATCHED_BRANCH)" ]; then \
+		echo "Aborting! Current branch is $(FABRIC_CURRENT_BRANCH) but want $(FABRIC_PATCHED_BRANCH)"; \
+		echo "Please checkout $(FABRIC_PATCHED_BRANCH) and come back."; \
+		exit 1; \
+	fi
+
+	@if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
+		echo "Aborting! Nothing to do here ... it seems you haven't applied the patches yet"; exit 1; \
+	fi
```
g2flyer, line 48: hmm, exit 1 for this condition is ugly, you shouldn't terminate make if you run it in an already cleaned setup? i think it should be the inversion of the test and then everything subsequent inside?

fabric/Makefile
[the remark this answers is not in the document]
g2flyer: hmm, not sure whether it is really beyond the scope given that this just part of the whole refactoring of the fabric build.  That said, I guess instead of going forth and back here, i guess we can table it and i will push a corresponding PR with my proposal once it is merged ...

fabric/0001-Integrate-tlcc-syscc.patch
g2flyer: hmm, not sure how modules or vendor will help regarding the patch itself?  It only would help making the building of the patch (potentially) easier/cleaner as the symlinking we currently do?

fabric/Makefile
mbrandenburger: I agree! This would be better - I've learnt that a exit 0 does not work in this case an there is not wait to abort without an error. As agreed we can improve this in another PR.

mbrandenburger:
rebased and squashed

fabric/0001-Integrate-tlcc-syscc.patch
g2flyer: @MHBauer What i meant is i doubt we get away with _any_ patch.  This patch is as minimal as it gets and nothing with modules or vendor aleviates the need for the patch itself (only could simplify make cleaner tying in and build the related FPC code) ...

ecc/enclave_chaincode.go
```diff
@@ -112,6 +102,32 @@ func (t *EnclaveChaincode) setup(stub shim.ChaincodeStubInterface) pb.Response {
 		return shim.Error(errMsg)
 	}
 
+	// write mrenclave to ledger
```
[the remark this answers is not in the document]
mbrandenburger: this is still the way how tlcc and ecc-vscc can get access to mrenclave

PR #325, opened 2020-04-08
tools.go for goimports

MHBauer:
https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md

Signed-off-by: Morgan Bauer <mbauer@us.ibm.com>

I think it may help with some of the go.mod changing.

Not sure if there is an existing issue created for it.
Fixes #

Please see the linked documentation describing this as the suggested path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ linter: gotools build
 
 gotools:
 	# install goimports if not present
-	$(GO) get golang.org/x/tools/cmd/goimports
+	$(GO) install golang.org/x/tools/cmd/goimports
 
 godeps: gotools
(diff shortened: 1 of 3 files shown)


MHBauer:
I feel like I dreamed doing this. Please try it again, I think maybe differing versions of go are behaving slightly differently.

MHBauer:
rebased for check with recent changes.

g2flyer:
given that i was the one complaining about this issue, you can make me the assignee, although a second review might still be nice?

PR #327, opened 2020-04-09
Make peer.sh also work for non-fpc code

g2flyer:

This PR improves #320 to allow peer.sh also be used for non-fpc chaincode (and hence allow upgrading the deployment integration test) 

No explicit issue but addresses limitation mentioned in #320 

See #320 for hints on how to test. For non-fpc chaincode i used `marbles02`, i.e., extended steps from #320 with
 ```
      # marbles02
      LBL=marbles02
      VER=0
      CC_LANG=golang
      CC_PATH=github.com/hyperledger/fabric-samples/chaincode/marbles02/go
```

it should not break backwards compatability; it does add though now also --help documentation on the --sgx-mode/-s flags for package introduced in #320

### DIFF
--- a/fabric/bin/peer.sh
+++ b/fabric/bin/peer.sh
@@ -64,6 +64,16 @@ handle_lifecycle_chaincode_package() {
 	    # one arg (true as of v2.0.1) and then the remaining one is the
 	    # output file ...
 	    -h|--help)
+		# with help, no point to continue but run it right here ..
+		try $RUN ${FABRIC_BIN_DIR}/peer "${ARGS_EXEC[@]}"
+		# .. as well as augment it with additiona -s/--sgx-mode arg
+		echo ""
+		echo "Flags for fpc-c chaincode:"
+		echo "  -s, --sgx-mode string                SGX-mode to run with"
+		exit 0 
+		;;
+
+	    --clientauth)
 		OTHER_ARGS+=( "$1" )
 		shift
 		;;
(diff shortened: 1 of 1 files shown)


Jordy24:
`panic: Could not find plugin at path ../../plugins/ercc-decorator.so: stat ../../plugins/ercc-decorator.so: no such file or directory` this is the error am getting... or maybe am not testing correct ?

mbrandenburger:
> `panic: Could not find plugin at path ../../plugins/ercc-decorator.so: stat ../../plugins/ercc-decorator.so: no such file or directory` this is the error am getting... or maybe am not testing correct ?

doube check that there is the ercc decorator `ercc-decorator.so` in `$FPC_PATH/plugins/` when you run the test.

make sure you run the test after a fresh `make clean && make`.

Jordy24:
could you show the command you are using for testing?? am using the script from #320 and thats not working for me

g2flyer:
> I am using the script from #320 and thats not working for me

Can you be a bit more specific on what you did and what did not work?  Note the code fragment i added to #320 are not a script to run automatically but just a toolbox of commands to easily test manually.   E.g., i would do ledger_init, then pick one of the example chaincode and set there variables and then one by one do the installation and try the invoke corresponding to that example chaincode (as indicated by the comment above the invoke). then i might try another chaincode do installation and test invoke, and finally do the shutdown. All of that essentially should work.   But of course it also depends that you started from a clean build first (e.g., by doing what marcus has said or an even "cleaner" `make clobber all`

Issue #165, opened 2019-09-19
Change our examples and demo to fabric 2.0 lifecycle commands


g2flyer:
move from old lifecycle commands (peer chaincode install/instantiate) to new lifecycle (peer lifecycle chaincode package/approvefororg/commit).

Probably best doing in conjunction with #98 as motivation of this issue is to enable _lifecycle_ endorsement policies which can require multiple organizations (technically, we are currently limited to have (transaction) endorsements only by a single peer (TEE/SGX enclave attestation) endorsement.

Sub-work-items:
- [x] example conversion
- [X] demo (via mockserver)
- [ ] demo (via fabric network)
- [ ] demo (via fabric network and migrated from fabric-client to fabric-network)

g2flyer:
Precondition: resolution of issue #254 and #247

g2flyer:
Note: the demo with docker-compose is postponed to concept-release 2.1

g2flyer:
Note: According to [fabric node-sdk migration guide](https://hyperledger.github.io/fabric-sdk-node/master/tutorial-migration.html):
- the fabric-client apis we use in `demo` (and also in `utils/docker-compose/node-sdk`) are now depricated and one should use fabric-network in turns, and
- "SDK no longer aims to provide administrative and management capability. The command-line interface should be used for these operations."
This means that we should eventually migrate the demo client side code to the new APIs.  Right now i've just added a separate checkbox above but we probably will eventually have to "outsource" checkboxes  into separate smaller work-item(s) ...

PS: Also note that demo with fabric networks depend on resolution of #256

g2flyer:
Note: the example02 used in deployment test doesn't exist anymore, probably easiest to just replace with an example from fabric samples, e.g., marbles02 ...

mbrandenburger:
#330 converts examples

PR #330, opened 2020-04-09
Remove init chaincode API and enables auction and echo tests

### LINKED ISSUE #326: Remove "init" from fpc shim
**Description**
With new lifecycle, init is generally deprecated in fabric and also difficult to support for us.  Hence, we should remove it
- [ ] from `shim.h` (and related c-code ...)
- [ ] examples using it (e.g., auction example)

**Link to feature or bug**
#165 depends on this being done

**Link to dependent issues**
no dependency

mbrandenburger:

- Remove init in echo example
- Refactor init in auction example and make normal invokable function
- Remove init in demo chaincode
- Remove init in hell world tutorial
- Remove init from shim.h; ecc and ecc_enclave components
- Enable auction and echo tests
- Deployment test still disabled until #327 is merged

Fixes #326 and partially addresses #165

```
cd integration
make
```

Remove init interface from FPC Chaincode; This breaks compatibility with concept-release-v1 chaincodes.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -360,19 +360,20 @@ Add the following content to the function, `helloworld_test()` in test.sh.  Plea
 ```
 helloworld_test() {
     say "- do hello world"
-    # install and instantiate helloworld  chaincode
 
-    # builds the docker image; creates the docker container and enclave;
+    # install helloworld  chaincode
     # input:  CC_ID:chaincode name; CC_VERS:chaincode version;
     #         ENCLAVE_SO_PATH:path to build artifacts
     say "- install helloworld chaincode"
-    try ${PEER_CMD} chaincode install -l fpc-c -n ${CC_ID} -v ${CC_VERS} -p ${ENCLAVE_SO_PATH}
-    sleep 3
+    PKG=/tmp/${CC_ID}.tar.gz
+    try ${PEER_CMD} lifecycle chaincode package --lang fpc-c --label ${CC_ID} --path ${CC_PATH} ${PKG}
+    try ${PEER_CMD} lifecycle chaincode install ${PKG}
 
-    # instantiate helloworld chaincode
-    say "- instantiate helloworld chaincode"
-    try ${PEER_CMD} chaincode instantiate -o ${ORDERER_ADDR} -C ${CHAN_ID} -n ${CC_ID} -v ${CC_VERS} -c '{"args":[]}' -V ecc-vscc
-    sleep 3
+    PKG_ID=$(${PEER_CMD} lifecycle chaincode queryinstalled | awk "/Package ID: ${CC_ID}/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
+
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VERS}
+    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VERS}
+    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VERS}
 
     # store the value of 100 in asset1
     say "- invoke storeAsset transaction to store value 100 in asset1"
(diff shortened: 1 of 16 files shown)


examples/README.md
```diff
@@ -421,19 +360,20 @@ Add the following content to the function, `helloworld_test()` in test.sh.  Plea
...
 
-    # builds the docker image; creates the docker container and enclave;
+    # install helloworld  chaincode
     # input:  CC_ID:chaincode name; CC_VERS:chaincode version;
     #         ENCLAVE_SO_PATH:path to build artifacts
     say "- install helloworld chaincode"
-    try ${PEER_CMD} chaincode install -l fpc-c -n ${CC_ID} -v ${CC_VERS} -p ${ENCLAVE_SO_PATH}
-    sleep 3
+    PKG=/tmp/${CC_ID}.tar.gz
+    try ${PEER_CMD} lifecycle chaincode package --lang ${CC_LANG} --label ${CC_ID} --path ${CC_PATH} ${PKG}
```
g2flyer, line 369: CC_PATH is not defined? Neither is CC_LANG?

mbrandenburger: Good catch! will fix that

mbrandenburger:
> Additionally, does deployment test work with #327? If so, could you use that as input to a review for #327? And maybe already enable it here as incentives for others to review #327 ;-?

I tried deployment test with #327 but did not work - not sure if it's a problem with the changes on deployment test or something in #327. More investigation is needed. 

We could also merge this one here before #327 and deployment test could be used to check if #327 actually works. WDYT?

g2flyer:
> We could also merge this one before #327 and then use deployment test to check if #327 actually works. WDYT?

Hmm, i pushed first and the initial bug was on your end, not on mine ;-)  So i would propose we merge #327 first and then do this (with deployment test enabled)

g2flyer:
> Hmm, i pushed first and the initial bug was on your end, not on mine ;-) So i would propose we merge #327 first and then do this (with deployment test enabled)

btw: i've rebased #327 onto this PR and enabled deployment_test.sh and (after correctly uncommenting it _and_ realizing that part of the proxy patch got somewhere lost^1 :-) I could successfully run `make clean all` with all tests passing successfully.   So seems with #330 and #327 we have Concept Release 2.0!

If also noticed, that go.mod/go.sum changes after build. Seems because you added new dependencies but didn't do a `go mod tidy`?

^1: see separate upcoming PR ..

integration/deployment_test.sh
```diff
@@ -20,34 +20,49 @@ CC_VERS=0
...
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name auction_test --version ${CC_VERS}
+    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name auction_test --version ${CC_VERS}
+    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name auction_test --version ${CC_VERS}
+
+    # install something non-fpc
+    PKG=/tmp/marbles02.tar.gz
+    try ${PEER_CMD} lifecycle chaincode package --lang golang --label example02 --path github.com/hyperledger/fabric-samples/chaincode/marbles02/go ${PKG}
+    try ${PEER_CMD} lifecycle chaincode install ${PKG}
+    PKG_ID=$(${PEER_CMD} lifecycle chaincode queryinstalled | awk "/Package ID: example02/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name example02 --version ${CC_VERS}
```
g2flyer, line 38: Minor thing but given that we use now marbles02, we might name the chaincode also marbles02 rather than examples02 (which was the code name of the old chaincode)?

mbrandenburger:
I enabled deployment test, so travis should fail. Once #327 is merged, we re-run travis and then this PR completes the mileston.

g2flyer:
FYI: I've updated branch now that #327 is merged. @mbrandenburger i think you can do an autosquash assuming travis build works fine. Unless anybody else (@bvavala @MHBauer @Jordy24 ) wants to show this PR some love in terms of review, i will merge it then later today ...

PR #331, opened 2020-04-10
Complete proxy patch (earlier version contained for unknown reason on…

g2flyer:

The earlier proxy PR #266 had for some unknown reason only half of the necessary proxy patches in the peer.  Discovered testing #330 with #327

Without this PR, #330 will fail trying to build ercc or marbles behind a proxy will fail due to a timeout. WIth this patch applied, #330 (with or without #327 applied) will nicely build and run integration tests also behind a proxy ...

### DIFF
--- a/fabric/0001-add-proxy-support-for-docker-in-peer.patch
+++ b/fabric/0001-add-proxy-support-for-docker-in-peer.patch
@@ -1,13 +1,65 @@
-From 0d9a0a7954c79d17d48e4ba0fee881027b6ec25b Mon Sep 17 00:00:00 2001
+From ffabe25186d43af74bf8dd351d5237865cadc68c Mon Sep 17 00:00:00 2001
 From: Michael Steiner <michael.steiner@intel.com>
 Date: Wed, 18 Mar 2020 19:42:55 -0700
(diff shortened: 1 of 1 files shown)


g2flyer:
> I don't recall the ercc container requiring to download any packages before.

It requires dependent modules; also some refactoring in Fabric v2 code seemed to require me changing two places instead of one. (I had figured that already in the old PR; alas the second half of the patch got lost in the format patch.).   Also note unless you remove the `ercc` docker images, it might not be rebuild and you might not phase the problem!   In any case, i've rebased now to latest master.  If you are happy with it, you can merge (given that you are the only proxy user, i guess there is not much point to ask for a second review unless somebody wants to dig into the actual fabric code to disect the patch itself ...)

bvavala:
> It requires dependent modules; also some refactoring in Fabric v2 code seemed to require me changing two places instead of one. (I had figured that already in the old PR; alas the second half of the patch got lost in the format patch.).

Even before, the patch touched 2 places: docker controller and chaincode support.
I quickly looked at the code and, even in 1.4, the code under `platforms` creates a container. But that part was not patched. I speculate it is due to the newer `ccenv` image but I don't think it is worth investigating this any further.

PR #333, opened 2020-04-20
component/registration diagrams

bvavala:
revised drafts of component and registration diagrams


docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
-
-note left ERCC1
-  FPC chaincode has three different associated endorsement policies (EP):
-  - lifecycle transactions ("normal fabric": approve chaincode definition (define application transaction EP), upgrade, etc ...)
-  - application transactions
-  - FPC-lifecycle transactions (register, key-gen, key-dist, ...)
-
-  Decision: We use a **separate** ERCC.
-  We realized FPC-liflecycle as application transaction EP of ERCC chaincode;
-  should ideally match lifecycle EP which is 'Channel/Application/LifecycleEndorsement' and by default majority of organizations)
```
g2flyer, line 57: as discussed in phoenix, we should discuss somewhere the involved endorsement policies and the requirement of consistency between the EP errc uses and the Lifecyle EP used by fpc chaincode creation.  Insofar some of these comments would still be useful?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
-
-  |||
-
-  create participant "ECC_Enclave" as Enclave1 order 30 #99FF99
-  ECC1 -> Enclave1 ++: create (SPID, CC_NAME, CC_VER, Org_1, TLCC_MR, CH_id)
-  note left
-    this is a cgo call
-    Org_1 denotes the peer's MSP_ID
-  end note
+  Admin -> ECC1 ++: invoke create(SPID)
```
g2flyer, line 40: as discussed in phoenix, we probably should either separate user action from client (code) function or just have external triggering events to characterize what the end-user has to do.  E.g., there doesn't seem any reason why would want to have the human do the later ias interaction and ercc registration steps.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
+  'ECC1 -> TLCC1 ++: getInfo
+  'note over TLCC1
+  '  This information could also be retrieved elsewhere if easier. At this point in time
+  '  it does not have to be authenticated (and is anyway passed through untrusted ECC),
+  '  consistency will be verified later in registration (Remember that TLCC_MR is defined
+  '  and agreed by human consensus in the ERCC chaincode definition)
+  'end note
+  'return TLCC_MR, CH_id
+
+  ECC1 -> Enclave1 ++: invoke create(SPID)
```
g2flyer, line 51: as discussed last week, i think it would be clean (as in simpler from a overall protocol perspective and better in supporting also dcap) when the spid owning and ias interaction is done peer-side.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
+
+  Enclave1 -> ECC1 ++: get_signed_proposal
```
g2flyer, line 53: As already mentioned by me and marcus, it would be nicer to have explicit enclave lifecycle/mgmt functions in the edl rather than just splicing through the shim commands.  I.e., much closer to what we had in the old version than what you have now.  Certainly, to get the parameters for the calls we might use the shim functions you listed but the would be all inside the untrusted part of the ecc.  On one hand i think this makes understanding the protocol much easier and allow the reasoning of the security and also allows much more easily eventually tighter integration with dedicated fabric cli admin commands or alike.  


btw: given that you do not really need the signature and i think the shim can give you all values consumed, i would replace the get_signed_proposal call with individual getters for the required fields.  The signature here is really confusing when you look at it from a security protocol perspective as we don't really care about this signature at all ...

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
-  note left Enclave1
-    ECC_ID is //short// unique identifier
-    ECC_ID = Hash(ECC_S_PK)
+    keyGen generates the keys stored in the key ring,
+    and described in the Credential protobuf
   end note
 
-  Enclave1  -> ECC1 ++: binding (ECC_S_PK, ECC_E_PK)
+  group Org-Enclave binding (to be discussed)
+  Enclave1  -> ECC1 ++: binding (ECC_VK)
```
g2flyer, line 72: any reason you dropped the encryption key from the binding?  Also i would leave what is exactly signed (which should be more than only ECC_VK but also context). I haven't checked, if you added a corresponding data struct to the other diagram, just a reference to that datastructure should do as well (e.g., simular to what you have done for Credential)

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
+    Enclave_SK, Enclave_DK, CC_DK,
+    TLCC_MR
   end note
-  return ATTESTATION, ECC_S_PK / error
-
-  deactivate Admin
-  |||
+  end group
+
+  return Credentials protobuf
```
g2flyer, line 103: See above why i think below should be done peer-side and not client-side.  See also the related high-level calls i've outline in my comment around attestation in the old diagram ...

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
+  end ref
+  activate ERCC2
 
-    ERCC1 -> ERCC1 : check consistency of quote (report-body)
-    note left
-      ECC_MR (mrenclave of ecc) is defined in ECC definition;
-      TLCC_MR (mrenclave of tlcc) is defined in ERCC definition
-      CH_Id is read from ledger or defined in ERCC definition
-    end note
+  ERCC1 -> ERCC1 : check verification report (through hard-coded certificate)
```
g2flyer, line 142: what seems missing here is the check of the consistency of the params in the registration and the chaincode definition. Ideally ercc still would be able to get that directly locally through some cctosyscall or alike.  The replication of the whole approval/commit handshake (and hence "outsourcing" the consistency check to a quorum of admins) could be a (last) alternative. But it definitely should be explicit as this is crucial for security.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -12,261 +12,155 @@
...
-  Decision: We use a **separate** ERCC.
-  We realized FPC-liflecycle as application transaction EP of ERCC chaincode;
-  should ideally match lifecycle EP which is 'Channel/Application/LifecycleEndorsement' and by default majority of organizations)
-  Note that with the **integrated** approach we have deal with a mix of applications transactions EP and FPC-lifecycle EP.
-end note
-
-ref over User1, ERCC1
-  Create channel; each peer joins channel; each peer launches TLCC, which will  "imprint" itself with CH_id (content-addressable channel identifier, hash of gensis block)
+ref over ERCC1, ERCC2
+  It is assumed that: channel has already been created; ERCC has been installed.
```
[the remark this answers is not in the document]
bvavala: > We should also mention here assumption/constraints from the calls to standard CC lifecycle functions to create a CC.

I have put a comment below.

> Conceptually, the credential _creation_ part can be earliest with `approveformyorg` as only then all the parameters are defined. With external builder there is the additional constraint that the (somewhat persistant) `run` script/process will only be triggered after commit for peer where commit is created (not sure when run is called for other peers ...)

For the registration, we can simplify the whole process by performing it after the chaincode commit -- a separate script run by the admin; I have added a comment.
Hopefully we can later embed this in `run` -- endorsing peers will have to run this sooner or later, possibly on first invocation.


> The `registration` of the credentials can/should be only after successful `commit` (as otherwise we might have enclaves for non-existing contracts)

I was confused by this. Avoiding enclaves for non-existing contracts is not really a priority for us. However, as you mentioned below, making sure that the registration and chaincode definition (specifically mrenclave) match _is_ critical.

docs/design/fabric-v2+/fpc-registration.puml
bvavala: added a comment.

docs/design/fabric-v2+/fpc-registration.puml
bvavala: any suggestion on how to do the separation?
I have added a comment the these operations can be scripted/automated and that human intervention is sufficient but not necessary.

docs/design/fabric-v2+/fpc-registration.puml
bvavala: * a peer component (e.g., enclave manager) should take care of that; and we would require all the control APIs up to the enclave
* if we can embed these ops in the external builder scripts, then those would be peer-side
(no strong opinion on that... as long as it does not delay the implementation)

docs/design/fabric-v2+/fpc-registration.puml
bvavala: > As already mentioned by me and marcus, it would be nicer to have explicit enclave lifecycle/mgmt functions in the edl rather than just splicing through the shim commands I.e., much closer to what we had in the old version than what you have now.

this was _necessary_ for the old version. In this version, it does not make much difference.

> Certainly, to get the parameters for the calls we might use the shim functions you listed but the would be all inside the untrusted part of the ecc. On one hand i think this makes understanding the protocol much easier and allow the reasoning of the security and also allows much more easily eventually tighter integration with dedicated fabric cli admin commands or alike. btw: given that you do not really need the signature and i think the shim can give you all values consumed, i would replace the get_signed_proposal call with individual getters for the required fields. The signature here is really confusing when you look at it from a security protocol perspective as we don't really care about this signature at all ...

This is more an implementation detail than design. Sure we can have a different call. The point is whether we should implement something additional when we already have something that works -- the `get_signed_proposal` is part of the fabric shim, which we are already supposed to expose, and which already returns the necessary information (though, as you say, more than that).

docs/design/fabric-v2+/fpc-registration.puml
bvavala: no reason, I tagged this as TBD.
I am actually not sure anymore about this step. This is for the Org to certify (i.e., taking accountability of) the enclave, and then let the enclave attest that. Do we still need this?
The certification can be done peer-side, then the certificate check is up to TLCC.

docs/design/fabric-v2+/fpc-registration.puml
bvavala: > what seems missing here is the check of the consistency of the params in the registration and the chaincode definition. Ideally ercc still would be able to get that directly locally through some cctosyscall or alike.

Good catch, I have added it.

> The replication of the whole approval/commit handshake (and hence "outsourcing" the consistency check to a quorum of admins) could be a (last) alternative. But it definitely should be explicit as this is crucial for security.

We can avoid this alternative on registration -- we'll revisit this for chaincode upgrades.

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: the cleanest would be to separate the admin/client _human_ and _sw_ as two seperate entitites.  An easier (though maybe visually more confusing) would be to add the human triggering the action as a note (ideally with a different color as the other ones)

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: your last point is exactly where i was heading it.  We need somewhere a protocol specification which leaves out irrelevant implementation details so one can easily see the overall flow and reason about the security.   The implementation details you added not only blur the bigger picture but imho also force certain implementation choices (e.g., they force a query-embedding of the management command and preclude a separate clean cli extensions).  Additionally, leaving low-level implementation details out of the UML also will increase the propability that the diagrams will also be uptodate over time

docs/design/fabric-v2+/fpc-registration.puml
bvavala: I have tried to give it a shot

docs/design/fabric-v2+/fpc-registration.puml
bvavala: I have tried to remove many details and to separate the management part.
Ok to remove the blur, and let's move on with the implementation -- the fastest one is clearly by embedding the management part in queries.

Ok to leave out implementation details, but there is nothing wrong (and it's actually highly desirable imho) if we update the diagrams over time as we learn!

docs/design/fabric-v2+/fpc-registration.puml
mbrandenburger: I would keep it as it was unless you can argue that we don't need that binding anymore.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,191 @@
...
+  Enclave1 -> Enclave1 : Credentials <- <CC_Parameters, Public_Keys>
 
-  Enclave1  -> ECC1 ++: binding (ECC_S_PK, ECC_E_PK)
+  group Org-Enclave binding (to be discussed -- not implemented)
+  Enclave1  -> ECC1 ++: binding (ECC_VK)
   ECC1 -> ECC1 : SIG_Org_1 <- create signature
-  note left
+  note left #FFaa00
     Org_1 takes "ownership" of this enclave
-    sign <CC_NAME, CC_VER, Org_1, ECC_S_PK, ECC_E_PK> with Org_1_S_SK
```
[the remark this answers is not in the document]
bvavala: it is, though it has been changed to implement (more realistically) the certificate signing request mechanism.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
+        The interaction with TLCC is not relevant. Rather, it is convenient to
+        access the chaincode definition.
+    end note
+    create "TLCC Peer1_Org1" as TLCC1 order 45
+    ERCC1 -> TLCC1  ++: approve consistency of Credentials
+    TLCC1 -> TLCC1    : check Credentials against chaincode definition
+    note over TLCC1 #FFaa00
+        This check ensures that:
+        1) the chaincode version (i.e., the MREnclave) matches the version of the
+            approved chaincode.
```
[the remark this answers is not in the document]
bvavala: (interesting point)
Not sure about this. I rather think we do not, here, and for multiple reasons:
1.  `TLCC_MREnclave` is hard-coded in ERCC, and already checked by ERCC on registration;
2.  ERCC is our root-of-trust for this value, since many participants have to agree on it
3. in a heterogeneous deployment, some peers may not have TLCC (and ERCC would check the chaincode definition by pulling it somehow from Fabric); still, the ERCC check is sufficient because it ensures that the chaincode will not establish a secure channel with a different TLCC -- if this is actually sound, it would be compelling for network with SGX-enabled designated FPC peers mixed with legacy Fabric peers.

docs/design/fabric-v2+/fpc-components.puml
```diff
@@ -7,36 +7,133 @@
...
+end note
+note right of CC_Parameters::CC_Version #FFaa00
+    initialized from Create transaction proposal
+end note
+note right of CC_Parameters::Channel_Id #FFaa00
+    initialized from shim (need check from TLCC?)
+end note
+note right of CC_Parameters::MSPID #FFaa00
+    initialized from shim (need check from TLCC?)
+end note
```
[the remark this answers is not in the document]
g2flyer: hmm, the potentially wrong parameter you have in all cases as the info comes from untrusted space, only verification during registration really ensures validation? Regarding duplication, this is an artifact of the short-term realization of using standard invoke (and hijacking part of the corresponding namespace). Long-term you certainly would want to get rid of that as well.  But i guess this is the same instance of where we just seem to fundamentally disagree on importance in these diagram -- close to short-term implementation vs protocol-readability and more long-term outlook.  So i would leave it at the 'agree to disagree' as above.   

That said, the current text `initialized from shim (need check from TLCC?)` still seems to me unclear.  First i guess it's really about `retrieve' and not `check`?  Also regarding the info, MSP and channelid you can get from the shim (in fact, getting channelid from tlcc gets you in a chicken'n'egg problem  as there is only one tlcc but potentially multiple channels, at least in long run ...), What i'm not sure on how to from the context in the shim is version and sequence number. What for sure you wouldn't get from shim (and i didn't see it right away in signedproposal or alike -- rather difficult to explore given the way the protobufs are defined as lots of byte fields which then recursively have to manually extract the type and the corresponding protobuf ... -- ) is the tlcc_mrenclave. Getting it from tlcc of course would be one way.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
+  end group
 
-  Manager1 -> storage1 ++ : store sealed state
-  note right
-    sealed state is stored under a
-    unique identifier consisting of
-    CC_NAME, CC_VER, CC_ID, ECC_S_PK, ECC_E_PK
-  end note
-  return
+  Enclave1 -> Enclave1 : Credentials <- <Verb="CREATE", Attested_Data, Evidence>
```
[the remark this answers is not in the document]
g2flyer: i didn't get an error in running puml but it showed `<Verb="CREATE", Attested_Data, Evidence>` verbatim and i couldn't make any sense of the Verb="CREATE" and thought that might be some puml-ism which went wrong.  I see now that you do have a verb attribute in `Credential` but i have zero intuition what the semantics is and why we have that.  If we really need that, can we give it a more semantically-meaningful name and given that it seems to be an enum, also define it as an enumeration?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
+'    - EnclavePersist: returns sealed state (important for security that this is called only after
+'      EnclaveProof [which we might want to enforce using some state-machine?] )
+'  end note
+'  return ATTESTATION
 
 end
 
-group register attestation
+group Org-Enclave binding/certification (to be implemented)
+  Admin -> E_Manager ++: certify(CC_Id, Cert_Parameters)
```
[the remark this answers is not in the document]
g2flyer: Hmm, isn't the reason of the cert that also the peer (org) claims owner ship of `CC_Parameters ` (and not only `CC_Id`) as the enclave does via the attestation? So `CC_Parameters` would seem to me to have to be part of the CSR (with the involved enclave PK of course as the to-be-certified key).  And the role is probably crucial as it affects the endorsement policies (where by default you would expect a certificate with role `peer`?)  So an empty Cert_Parameter seems strange to me ....

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: when i read it i saw the comment of everything which has to be verified during registration and verification of TLCC_MREnclave would be a crucial part.  Now whether that is done in ercc or tlcc that is an orthogonal question (and to some extent irrelevant from a protocol perspective where tlcc comes in only as the potential source of some information but not really from the usual 'tlcc-as-trusted-leger-for-ecc-in-hostile-environment' context^1 . Personally i think it is cleaner and easier to understand if all checks are done at a single place (and preferably ercc) but i guess that might be another of the 'agree-to-disagree' points. 

^1: given that we all agreed that for situations like these the tlcc enclave (which is really the heart of tlcc) is not involved, i think we should actually add a separate system-chaincode/component to make clear that this has nothing to do with the usual goal/functionality of tlcc ....

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
-
-    ERCC1 -> ERCC1 : Retrieve ECC definition for <CC_NAME, CC_VER>
+  ref over ERCC2
+    (same as ERCC1)
+  end ref
+  activate ERCC2
 
-    ERCC1 -> ERCC1 : Verify SIG_Org_1
+  ERCC1 -> ERCC1 : check Credentials.Evidence
+  ERCC1 -> ERCC1 : check Credentials.Attested_Data.CC_Parameters
```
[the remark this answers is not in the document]
g2flyer: Hmm, chaincode definition is not fabric internal secret, it is just not exposed to chaincode but is exposed to admins (which define it in the first place).   I still would prefer to understand right away all checks performed when looking at ercc code but if everybody else is happy with outsourcing the check rather than only the retrieval of the chaincode definition, then i will consider myself overruled :-)

docs/design/fabric-v2+/fpc-components.puml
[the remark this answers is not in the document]
bvavala: I have removed all the comments in `CC_Parameters`.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: it's a constant string - I have added a comment

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
+    -Org_1 takes "ownership" of this enclave,
+    -the certificate should be made available to the chaincode endorsement plugin,
+     so to include it in the responses (for eventual validation) 
+  end note
+end group
 
+group enclave registration
   activate Admin
+  Admin -> E_Manager ++: register
+  E_Manager -> ERCC1 ++: register Credentials
```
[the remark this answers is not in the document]
bvavala: maybe, but we should deal with this later because:
1. we don't have the certificate
2. I am not sure about that. If the certificate is usually appended to responses (or anywhere else clients/peers can find it), do we also need it here? -- unless we replace the public key.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: This binding is far from "to be implemented". I am not going to discuss it now, and rather remove it if necessary.
BTW, `CC_Parameters` is already bound to the enclave's public key.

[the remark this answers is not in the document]
g2flyer: I know that CC_Parameters contains the pk, the point was that for the certificate the same pk of course should be the certified key of the X509 certificate.   

Regarding removing. I thought these diagrams  should be primarily the protocol definition. And from what we discussed in the past i thought it was pretty clean what we want to achieve here,  conceptually a "dual-signature" on the contract enclave context (essentially CC_Parameters) by both the enclave as well as the organization running the peer on which this enclave resides?  Baring some gory implementation details (e.g., which x509 extension field to use or alike) even from a design perspective X509 CSR more or less defines what you will have to do.    

That said, maybe the bigger question you are touching upon is whether in the protocol we want to document the X509-based clean validation path which will almost certainly not be in the MVP in Q3 but sooner or later would have to be moved to.  If we want to cover only Q3 MVP, then we should drop any mentioning of certificates.  Personally, i would prefer to specify the clean protocol and maybe mention the initial short-cuts.  That said, the crucial part is to agree "which version" we want to document and then doing that consistently (and from a protocol perspective, completely)?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-    verify and store
-  end ref
-  return register endorsement
+  return register endorsement /' E_Manager -> ERCC1 '/
 
-  Admin -> ERCC3 ++: register (CC_NAME, CC_VER, MSP_ID_ORG_1, SIG_Org_1, ECC_S_PK, ECC_E_PK, ATTESTATION)
-  ref over ERCC3
-    verify and store
+  ref over E_Manager
+    submit transaction
```
mbrandenburger: I think only two orgs is boring! :)

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: as mentioned above, either we do certificates and then do the whole thing or we drop them completely. Putting them only half-way doesn't seem useful but rather confusing?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
   end ref
-  return register endorsement
-
-  Admin -> Admin : create register transaction
 
-  Admin -> Orderer : submit register transaction
+  return result
```
mbrandenburger: skipping this will not complete the enclave registration.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: would be more useful to add the comment in the components file? Also why call it `verb`?  That doesn't provide any intuition from what you added as comment.  And from that comment i still do understand why we need it in the first place? What does "reuse a protobuf mean"?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-
-ref over User1, ERCC1
-  Install and approve ERCC chaincode. ERCC defines TLCC_MR and IAS_PK in chaincode definition.
-  (alternatively, TLCC_MR and IAS_PK could be in channel configuration; crucial is that they are approved by humans (admins) on sufficiently many organizations)
-  ERCC also reads CH_id from ledger or, if that does not work, also embeds it into chaincode
-
-  see fpc-lifecycle-v2.puml
-end ref
-|||
-
```
mbrandenburger, line 74: I think it makes it easier to understand when the enclave registration happens when we keep the reference to other diagrams here. So why removed?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-
-group enclave creation and attestation generation
-  
+group enclave creation
```
mbrandenburger, line 34: we still have attestation generation in this box happening ... why not keeping it in the group title?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
 
-  Manager1 -> ECC1 ++: create (SPID)
-  note left: this is a cc2cc call
+    PS: in a future tighter integration with fabric we could remove the first constraint
+    and could potentially piggyback the steps before the ercc registration (enclave and
+    evidence generation) in the `approveformyorg` and the registration as part of
+    `commit` without having to change the protocol flows.
+  end note
 
+  create E_Manager order 15
```
mbrandenburger, line 49: as mentioned already above ... UML experts may complain as this means that e_manager instance is created here ....

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: * clarified the parameters that ercc checks,
* added a separate component and a comment to emphasize no need of enclave

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
mbrandenburger: sorry, I don't get that. `Cert_Paras` are just here for (maybe) later use? I am ok with this. But to be consistent with `create` we should remove CC_id and read it from cc_parms where it is needed. where is it actually needed?

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: AFAIK, Fabric does not expose the chaincode definition data structure.
If this holds, then the chaincode definition **is** an internal secret and is **not** exposed to admins. In fact, the admins only specify some parameters, and it's the Fabric documentation that says that those are the chaincode definition.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-  |||
 
 end
 
-group register attestation
+group Org-Enclave binding/certification (to be implemented)
+  Admin -> E_Manager ++: certify(CC_Id, Cert_Parameters)
+  E_Manager -> ECC1 ++: getCSR(Cert_Parameters)
+  ECC1 -> Enclave1 ++: getCSR(Cert_Parameters)
+  return csr /' ECC1 -> Enclave1 '/
```
mbrandenburger, line 130: here some action is missing ... like `csr <- generate_csr(some inputs? cc_params???)`

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
+  return csr /' E_Manager -> ECC1 '/
+  E_Manager -> E_Manager : generate peer certificate for enclave (certified by Org_1)
+  return certificate
+  note over Admin, E_Manager #FFaa00
+    -Org_1 takes "ownership" of this enclave,
+    -the certificate should be made available to the chaincode endorsement plugin,
+     so to include it in the responses (for eventual validation) 
+  end note
```
mbrandenburger, line 138: what would the admin now do with the certificate? send again to the peer who stores it on disk in some folder where endorsement plugin can access it at runtime? if we do so ... what are we using to map this cert to an enclave such that the endorsement plugin can pick the right one?!

Instead of returning to the admin, maybe the manager can keep the cert and store it locally (at the peer)?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
 
-  Admin -> ERCC1 ++: register (CC_NAME, CC_VER, MSP_ID_ORG_1, SIG_Org_1, ECC_S_PK, ECC_E_PK, ATTESTATION)
-  group verify and store
-
-    ERCC1 -> ERCC1 : check attestation signature & status
-    note left
-      - IAS verification key defined in ERCC definition
-      - status denotes the attestation result
+  ref over ERCC2
+    (same as ERCC1)
```
mbrandenburger, line 148: I liked the previous reference to an actual group more. A pitty that `verify and store` group is not there anymore :/

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: > I still would prefer to understand right away all checks performed when looking at ercc code

I really don't see how this would hinder your understanding, given that the one-line would be `if cc_def_exists(id, ver, seq) then ok, else abort`, where `cc_def_exists` is the call to TLCC.

> but if everybody else is happy with outsourcing the check rather than only the retrieval of the chaincode definition, then i will consider myself overruled 

If you mean "retrieving the protobuf", I would be very unhappy.
Otherwise... feel free to update it.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
+    -the certificate should be made available to the chaincode endorsement plugin,
+     so to include it in the responses (for eventual validation) 
+  end note
+end group
 
+group enclave registration
   activate Admin
+  Admin -> E_Manager ++: register
+  E_Manager -> ERCC1 ++: register Credentials
+  E_Manager -> ERCC2 : register Credentials
```
mbrandenburger, line 145: I would stick with a sequential view of actions here. Even though register Credentials can be invoked in parallel to multiple peers this makes reading this document not easier. Better just add a note that this invocation can also happen concurrently.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: > AFAIK, Fabric does not expose the chaincode definition data structure.

But we don't really care about the internal data structure, just the information which we also get from `peer lifecycle chaincode querycommitted` (which is probably a subset of the information in the channel definition?)

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-    - EnclaveProof: gets org endorsement-sig, creates report, then quote
-      and, if epid, gets IAS report. 
-    - EnclavePersist: returns sealed state (important for security that this is called only after
-      EnclaveProof [which we might want to enforce using some state-machine?] )
-  end note
-  return ATTESTATION
+  return Credentials, Sealed_Credentials
 
-  |||
+  E_Manager -> E_Manager : store Sealed_Credentials
```
mbrandenburger, line 118: I am wondering where this is stored .... with recent discussion what the manager actually is, in particular, if the manger is part of a peer or part of a client.
1. If the manager is part of the peer, store will dump the sealed credentials to peers local storage
1. if the manager is part of a client, where is it stored?

docs/design/fabric-v2+/fpc-registration.puml
mbrandenburger: aha ... now I see in line 195 the `submit transaction` action .... this makes me wondering what the manager actually is ... part of the peer or part of a client ... see related question above also.

1. If the manager is part of the peer, it cannot submit transactions, in particular, it cannot send transaction proposals to other endorsers, collect endorsements and submit a tx. It just has no client credentials for that. 
1. If the manager is part of a client (admin). it could indeed perform the actions just mentioned using the admin credentials ...

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: > would be more useful to add the comment in the components file? 

if somebody is unhappy looking at `verb` it can look it up and will find the comment.

> Also why call it verb?

Because it's a "verb", and a convention we have successfully used in PDO.

> That doesn't provide any intuition from what you added as comment.

I have explained in the comment what and why.

> And from that comment i still do understand why we need it in the first place?

(we discussed this several times, but here we go again)
The expectation is that an admin will request re-attestations in the future. These will not go through the `create` but rather through another api. Yet, the attestation block will be the same (or similar). The verb used in the re-attestation will be different, so we can distinguish the two cases.
 
> What does "reuse a protobuf mean"?

It means that the protobuf message will be generated "also" in other parts of the code. So, we will not change the protobuf, but rather simply the verb.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: > I really don't see how this would hinder your understanding, given that the one-line would be `if cc_def_exists(id, ver, seq) then ok, else abort`, where `cc_def_exists` is the call to TLCC.

If you rephrase it like that i would feel much better as it spells out the actual check.  It raises though another question: is checking for existence sufficient or should we check it is also currently _active_?   Issue probably primarily related to revocation? For dealing with confidentiality & revocation is a big rathole which also might require some re-encryption or more complicated mitigation/damage control but from an integrity perspective, probably purging pre-existing outdated enclaves from the registry and preventing any future registration of such outdated enclaves should be sufficient. So an "active-check" could be useful.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: > Because it's a "verb", and a convention we have successfully used in PDO.

well, you could replace `Credentials' by `Noun` and it would also be correct but not very intuitive.   The fact that it is used in PDO doesn't help anybody who is look at FPC to understand it.

What about just calling it version (with the first version having to be the one created when the seal-key is created)

> (we discussed this several times, but here we go again)
The expectation is that an admin will request re-attestations in the future. These will not go through the create but rather through another api. Yet, the attestation block will be the same (or similar). The verb used in the re-attestation will be different, so we can distinguish the two cases.

Then why not just write exactly this in the comment?  I agree that we have to distinguish the initial version from later ones but lets describe it with a more meaningful name and let's indicate the issue explicitly in the comment?  I knew the issue (i think i even brought it up in the first place) but the fact that i couldn't connect the dots makes it kind of obvious that the name is not intuitive?

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: i have the same question further up (in fact, i don't think it will actually work to have it at the client with the current flows ...)

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
 
-  ECC1 -> TLCC1 ++: getInfo
-  note over TLCC1
-    This information could also be retrieved elsewhere if easier. At this point in time
-    it does not have to be authenticated (and is anyway passed through untrusted ECC),
-    consistency will be verified later in registration (Remember that TLCC_MR is defined
-    and agreed by human consensus in the ERCC chaincode definition)
+  note over Admin, E_Manager #FFaa00
+    The E_Manager component hides some operations that do not necessarily need to be
+    performed by the Admin -- they could be scripted or embedded in the peer process.
```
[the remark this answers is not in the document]
g2flyer: > For now we have already assumed that client and peer go together.
Hmm, this is kind of news to me and i don't think it is really a good idea (for reasons discussed above).

docs/design/fabric-v2+/fpc-registration.puml
mbrandenburger: @g2flyer  agreed, but if the manager is part of the peer, we have the problem that the manager cannot invoke "register transactions" (as I commented here also somewhere)

g2flyer: Nothing prevents you from doing a transaction on the peer?

mbrandenburger: ok - I agree that nothing prevents you from setup a client (some client credentials are needed) on the same node where a peer runs. Luckily the `peer chaincode` command also provides all the tool one need to run this from a peer host.

I would like to keep the logical separation of client, peer, and orderer here. 

- Client: invokes chaincode (endorsements), assembles transactions based on endorsements, submit transactions, (a special client called admin) performs special mgn operations such as install chaincode/approve/createchannel/etc...

- Peer: executes chaincodes invoked by clients and produces an endorsement, maintains a ledger of blocks of transactions, verifies transactions received from an ordering service; 

- Orderer: receives transactions, append to block, broadcast block, access control 

With the separation in mind I see the manager as part of the peer. If I think about an implementation of that, the admin calls the create operation and receives back a blob (attestation, cc_params?!), which is then used as input when the admin calls register as "normal" ercc invocation.

Let's quickly play this through ... (we extend the peer cli with fpc functions ... maybe using the wrapper) and let the admin/client use this.

the admin calls ...
```
$ peer fpc create -cc_name MyChaincode -ch_name MyChannel --peerAddresses 
172.23.45:7050
```

create will return an `attestation.blob` or similar 
```
$ cat attestation.blob
Attestation, Evidence, CC_Parms ....
$ peer chaincode invoke -cc_name ERCC -ch_name MyChannel --peerAddresses 
172.23.45:7050 172.23.55:7050 172.23.65:7050 -input attestation.blob
```

This completes creation of an enclave.  

This is a two step action, very similar how a new channel is created. There the admin first submits a transaction to the orderer and gets back a block, which is then used as input at a peer with the join command.

Alternatively, if we think about adding client functionality to the manager and host it together with the peer, we add an additional requirement in the network deployment that every peer must be provisioned with "some client" credentials that allows to invoke ercc. Setting up a fabric network with multiple peers and clients is already complicated enough I think, so this approach would make it even harder.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
mbrandenburger: regarding chaincode definition. Clients can get the contents of the chaincode definition in human readable form by invoking lifecycle commands.

See example here: https://hyperledger-fabric.readthedocs.io/en/release-2.0/commands/peerlifecycle.html#peer-lifecycle-chaincode-querycommitted-example

The corresponding proto message you can find here: https://github.com/hyperledger/fabric-protos/blob/9944f8f73cfe0585a477f4b7d56a090374d0c7b6/peer/lifecycle/lifecycle.proto#L173

As a system chaincode we have the chance to retrieve a chaincode definition from the peer by following the implementation of the lifecycle command, see here: https://github.com/hyperledger/fabric/blob/master/core/chaincode/lifecycle/scc.go#L421

Some more references
- https://github.com/hyperledger/fabric/blob/9277b0eb3e6c4dadc37c69602e8e8f9c14af36ab/core/chaincode/lifecycle/scc.go#L95
- https://github.com/hyperledger/fabric/blob/9277b0eb3e6c4dadc37c69602e8e8f9c14af36ab/core/chaincode/lifecycle/scc.go#L421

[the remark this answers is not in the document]
g2flyer: > regarding chaincode definition. Clients/Admins can get the contents of the approved chaincode definition in human readable form by invoking lifecycle commands.

That's what i wrote just two comments above in the same thread ;-)
 
> The corresponding proto message you can find here: https://github.com/hyperledger/fabric-protos/blob/9944f8f73cfe0585a477f4b7d56a090374d0c7b6/peer/lifecycle/lifecycle.proto#L173

Note this though almost certainly _not_ the chaincode definition on the ledger: there is no code notarized on the ledger with Fabric v2.  That said, it does contain the information which we care about and that part should correspond to what is on the ledger.

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: > I would like to keep the logical separation of client, peer, and orderer here.
> 
> * Client: invokes chaincode (endorsements), assembles transactions based on endorsements, submit transactions, (a special client called admin) performs special mgn operations such as install chaincode/approve/createchannel/etc...

I think we should be careful, though, not to mix client in a sense of humans (e.g., admins) and client programs as complement to a server program. E.g., the external builder is a client to the peer as well but it is clearly not the same kind of client as an admin working from his laptop.

> * Peer: executes chaincodes invoked by clients and produces an endorsement, maintains a ledger of blocks of transactions, verifies transactions received from an ordering service;
> * Orderer: receives transactions, append to block, broadcast block, access control

> With the separation in mind I see the manager as part of the peer. If I think about an implementation of that, the admin calls the create operation and receives back a blob (attestation, cc_params?!), which is then used as input when the admin calls register as "normal" ercc invocation.
> 
> Let's quickly play this through ... (we extend the peer cli with fpc functions ... maybe using the wrapper) and let the admin/client use this.
> 
> the admin calls ...
> 
> ```
> $ peer fpc create -cc_name MyChaincode -ch_name MyChannel --peerAddresses 
> 172.23.45:7050
> ```

I think we should place that under `peer lifecycle chaincode ...` as it is really part of the lifecycle?

> create will return an `attestation.blob` or similar
> 
> ```
> $ cat attestation.blob
> Attestation, Evidence, CC_Parms ....
> ```
> 
> next the admin uses a normal chaincode invocation ...

Why do you even want to expose the blob to the admin (user), this could (imho should) all be part of above call?  I see your example of channel creation, but there it is as there multiple ways to create the channel definition (e.g., seems that the imagined "production-way" is not configtxgen but build it from  https://github.com/hyperledger/fabric-config; there was recently some related discussion on the fabric ML ....).  I don't think this would apply in our case?

> ```
> $ peer chaincode invoke -cc_name ERCC -ch_name MyChannel --peerAddresses 
> 172.23.45:7050 172.23.55:7050 172.23.65:7050 -input attestation.blob
> ```

If we decide that the human should do two commands instead of one as i suggest above, i would do this still as a (from the admin perspective) management command in the same namespace as above `create`.    This hides the current implementation (using invoke) and allows seemless tighter integration in the future.

> This completes creation of an enclave.
> 
> This is a two step action, very similar how a new channel is created. There the admin first submits a transaction to the orderer and gets back a block, which is then used as input at a peer with the join command.
> 
> Alternatively, if we think about adding client functionality to the manager and host it together with the peer, we add an additional requirement in the network deployment that every peer must be provisioned with "some client" credentials that allows to invoke ercc. Setting up a fabric network with multiple peers and clients is already complicated enough I think, so this approach would make it even harder.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
mbrandenburger: > That's what i wrote just two comments above in the same thread ;-)

so I just confirmed what you wrote! :)

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,191 @@
...
+  'end group
 
-  note left Enclave1
-    ECC_ID is //short// unique identifier
-    ECC_ID = Hash(ECC_S_PK)
-  end note
+  Enclave1 -> Enclave1 : generate Private_Keys <- <Enclave_SK, Enclave_DK, CC_DK>
+  Enclave1 -> Enclave1 : Public_Keys <- Public(Private_Keys)
+
+  Enclave1 -> Enclave1 : Credentials <- <CC_Parameters, Public_Keys>
```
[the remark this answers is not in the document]
bvavala: 867f9dee60e760dfb1a60ffe5d169c1fb621e154

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,191 @@
...
   end note
   return SIG_Org_1
-  |||
+  end group
 
-  Enclave1 -> Enclave1 : generate quote
-  note left
-    quote is defined with reportbody =
-    Hash(<CC_NAME, CC_VER, Org_1, SIG_Org_1, ECC_S_PK, ECC_E_PK, TLCC_MR, CH_id>)
+  Enclave1 -> Enclave1 : Attestation <- attest(Credentials)
```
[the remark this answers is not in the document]
bvavala: this was already moved into a separate box.
Regarding the signature, it will be part of the certificate.
Also, there seems to be no need to include it in the attestation since the user and/or validation procedure can verify the certificate chain.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,191 @@
...
+    Enclave_SK, Enclave_DK, CC_DK,
+    TLCC_MR
   end note
+  end group
 
-  return ECC_S_PK, ECC_E_PK, sealed state, quote
+  return Credentials, Attestation
 
-  return ECC_S_PK, ECC_E_PK, sealed state, quote, SIG_Org_1
+  return Credentials, Attestation
```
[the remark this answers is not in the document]
bvavala: sealed state is now Sealed_Credentials -- unimplemented.
Org signature was moved to org-enclave binding box

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,191 @@
...
-    end note
-
-    ERCC1 -> ERCC1 : Retrieve ECC definition for <CC_NAME, CC_VER>
+  ref over ERCC2
+    (same as ERCC1)
+  end ref
+  activate ERCC2
 
-    ERCC1 -> ERCC1 : Verify SIG_Org_1
+  ERCC1 -> ERCC1 : check Credentials
```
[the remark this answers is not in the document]
bvavala: 54014519c479fb5789679720d6a06ed26cc80b62

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: af26f472796ebeaf9248fc028485010daae83a7e
yes, you brought it up in the first place

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,200 @@
...
+'      and, if epid, gets IAS report. 
+'    - EnclavePersist: returns sealed state (important for security that this is called only after
+'      EnclaveProof [which we might want to enforce using some state-machine?] )
+'  end note
+'  return ATTESTATION
 
 end
 
-group register attestation
+group Org-Enclave binding/certification (to be implemented)
```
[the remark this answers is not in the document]
bvavala: same as above, it could be scripted now in the E_Manager and handled through invokes.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: 8c4fdd99521515f98fc0e6abce90548dc289d73f

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
bvavala: b9be319f68fe8af7ccf582098fd1ea06602809de

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-!pragma teoz true
-
-hide footbox
-title Foot Box removed
-
-title Fabric Private Chaincode - Enclave registry - New approach
-
-actor "User Org1" as User1 order 9
 actor "Admin Org1" as Admin order 10
-participant "ECC_Manager (syscc) Peer1_Org1" as Manager1 order 20
```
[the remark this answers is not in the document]
bvavala: d51cb9d66b29c0828bf7f6040eabf02c37b5c7cf

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
+    if necessary, a pre-validation can be performed through tlcc.
+    Another alternative could be to pass these parameters as part of the create call.
   end note
+  Enclave1 -> Enclave1: Require: CC_Parameters
 
-  note left Enclave1
-    ECC_ID is //short// unique identifier
-    ECC_ID = Hash(ECC_S_PK)
-  end note
+  Enclave1 -> Enclave1 : generate Private_Keys <- <Enclave_SK, Enclave_DK>
```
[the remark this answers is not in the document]
bvavala: 64d8c6cc73f427be68be1b782316072d76d8271c

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
+    This additionally requires some ocalls and helper code in untrusted environment.
+    For convenience, the enclave now returns the attestation, and the evidence
+    (i.e., the verification report) is retrieved from untrusted environment.
+  end note
+  Enclave1 -> IAS ++: Attestation
+  return Evidence <- Verification_Report
+end group
 
-  |||
+  Enclave1 -> Enclave1 : Sealed_Credentials <- seal(CC_Parameters, Private_Keys)
```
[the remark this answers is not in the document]
bvavala: correct

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,198 @@
...
-    ECC_E_SK : Enc priv key
+  note over Enclave1 #FFaa00
+    The requirement of CC_Paramenters can be met in practice as follows:
+    cc id, version, MSPID and channel id can be retrieved through Fabric shim functions;
+    tlcc mrenclave can be retrieved through ercc;
+    the validation of these parameters occurs when the registration tx is processed by ercc;
+    if necessary, a pre-validation can be performed through tlcc.
+    Another alternative could be to pass these parameters as part of the create call.
   end note
+  Enclave1 -> Enclave1: Require: CC_Parameters
```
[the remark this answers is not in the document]
bvavala: a73c270394eb354b4c56d804ce6208e2d4374a71

docs/design/fabric-v2+/fpc-registration.puml
bvavala: 2285d1c0e8775289a66f534f16e8ba45b0f9d102

docs/design/fabric-v2+/fpc-registration.puml
bvavala: b9be319f68fe8af7ccf582098fd1ea06602809de

docs/design/fabric-v2+/fpc-registration.puml
bvavala: 2b716a44f090dadd7906f75f2153becf2a9269a3

docs/design/fabric-v2+/fpc-registration.puml
bvavala: 2b716a44f090dadd7906f75f2153becf2a9269a3

docs/design/fabric-v2+/fpc-registration.puml
bvavala: for now the sealing part is not implemented, and it could be easily implemented as/with the create, storing the sealed blob at the return point. This is not worth any additional time.

bvavala:
pass complete

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
mbrandenburger: Sorry, I don't like the resolution to search in another document which is also very complex just to understand what *_DK, and *_SK, means ... thus is I suggested to just keep a note here. Anyway, I will not reject this.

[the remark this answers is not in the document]
bvavala: I understand we have different tastes here. Personally I would avoid replicating text already written elsewhere.
Anyway, I am fine if you want to step in and add a commit here.

docs/design/fabric-v2+/fpc-registration.puml
mbrandenburger: Thanks Bruno, I added a minor comment to this commit what I think is missing.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -7,266 +7,221 @@
...
-group register attestation
+group Org-Enclave binding/certification (to be implemented)
+  Admin -> E_Manager ++: certify
+  E_Manager -> ECC1 ++: getCSR
+  ECC1 -> Enclave1 ++: getCSR
+  Enclave1 -> Enclave1 : csr <- generate_csr(Private_Keys.Enclave_SK)
+  return csr /' ECC1 -> Enclave1 '/
+  return csr /' E_Manager -> ECC1 '/
+  E_Manager -> E_Manager : generate peer certificate for enclave (certified by Org_1)
+  return Certificate
```
mbrandenburger, line 140:
```suggestion
  E_Manager -> E_Manager : generate peer certificate for enclave (certified by Org_1)
  E_Manager -> E_Manager : store Certificate
  return
```

mbrandenburger: The manager should keep the cert for later use (ie. when calling register)

g2flyer:
Remaining unresolved issues (e.g., split on E_Manager into cli and peer as decided 5/22 and also done in lifecycle diagrams) to be handled in a separate follow-up PR

PR #335, opened 2020-04-20
Improved fabric patch/clean

g2flyer:
Simplified and more robust fabric patching

This resolves some remaining discussion items from PR #323


README.md
```diff
@@ -321,10 +321,10 @@ Checkout Fabric 2.0.1 release using the following commands:
 
     $ export FABRIC_PATH=${GOPATH}/src/github.com/hyperledger/fabric
     $ git clone https://github.com/hyperledger/fabric.git $FABRIC_PATH
-    $ cd $FABRIC_PATH; git checkout tags/v2.0.1 -b v2.0.1-fpc
```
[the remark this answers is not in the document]
mbrandenburger: I also agree that we don't need this fpc branch name to make our `fabric/Makefile` more robust and should rely on it to check if patches are applied. But still good to have this dedicated branch :)

fabric/Makefile
```diff
@@ -9,64 +9,55 @@ FABRIC_VERSION ?= v2.0.1
...
-	fi
-
-	@if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
-		echo "Aborting! Nothing to do here ... it seems you haven't applied the patches yet"; exit 1; \
-	fi
-
-	@if [ ! -z "$(FABRIC_TRACKED_CHANGES)" ]; then \
-		echo "Aborting! It seems you have some changes in your branch:"; \
+	@echo "Cleaning Fabric patches ..."; \
+	cd $(FABRIC_PATH); \
```
mbrandenburger, line 33: I would move this line closer where it is needed, that is, just in front of `$(FABRIC_CURRENT_DIFF_COMMITS_CMD)`

fabric/Makefile
```diff
@@ -9,64 +9,55 @@ FABRIC_VERSION ?= v2.0.1
...
-		exit 1; \
-	fi
-
-	@if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
-		echo "Aborting! Nothing to do here ... it seems you haven't applied the patches yet"; exit 1; \
-	fi
-
-	@if [ ! -z "$(FABRIC_TRACKED_CHANGES)" ]; then \
-		echo "Aborting! It seems you have some changes in your branch:"; \
+	@echo "Cleaning Fabric patches ..."; \
```
mbrandenburger, line 32: is this really needed to have almost everything in this target in a single line? we have not established common syntax for writing makefiles but there we more semicolons and backslashes than in other places.

README.md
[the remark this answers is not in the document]
g2flyer: personally, i felt the branch a PITA as it was easy to forget to create it and it didn't seem to add any value.  Not sure in what respect it helps with the switch?  I was wondering whether a branch could help if you actively work with some fabric patches but in it's current form it wouldn't really have helped either.  But I agree for uniformity there is no point to have the branch in .travis and dockerfile.
In any case, i would lean to remove the branch but i removed it also as i assumed you added it only because initially i was suggesting to use a branch.  If you feel strongly about having the branch, i'll add it back ...

fabric/Makefile
```diff
@@ -9,64 +9,55 @@ FABRIC_VERSION ?= v2.0.1
...
+	cd $(FABRIC_PATH); \
+	if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
+		echo "Nothing to do here ... it seems you haven't applied the patches yet";\
+		exit 0; \
```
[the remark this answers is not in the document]
g2flyer: me culpa: to retain output of executed commands, i split the commands on last minutes into two and that of course annuls effect of exit 0 ...

fabric/Makefile
g2flyer: ok, i'll moved it. Note that to handle above mea culpa but to leave still the '@' to ignore general pre-amble output, i've added now some 'set -x' before the actual state-changing commands and as this command doesn't make sense unless you know the `cd`, i've repeated it now also later below (just pro-activelly "fighting" a later comment ;-)

fabric/Makefile
g2flyer: yes, actually _everything_ has to be in a single execution or the exit 0 does not work ...

README.md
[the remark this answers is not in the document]
MHBauer: I don't see the need for the branch, but I'll check below.

fabric/Makefile
MHBauer: Yea, difficult to review. 

My understanding here is that all of this needs to take place within the context of the fabric path, which is difficult since we can't place our own makefile instructions inside to call out to it.
The result here is that everything has to happen in the same shell execution.

Any consideration of this special target?
https://www.gnu.org/software/make/manual/html_node/One-Shell.html

README.md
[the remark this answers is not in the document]
g2flyer: personally, i felt the branch a PITA as it was easy to forget to create it and it didn't seem to add any value.  Not sure in what respect it helps with the switch?  I was wondering whether a branch could help if you actively work with some fabric patches but in it's current form it wouldn't really have helped either.  But I agree for uniformity there is no point to have the branch in .travis and dockerfile.
In any case, i would lean to remove the branch but i removed it also as i assumed you added it only because initially i was suggesting to use a branch.  If you feel strongly about having the branch, i'll add it back ...

fabric/Makefile
g2flyer: I didn't know about `.ONESHELL`.  But it seems to have major drawbacks as it applies to all targets and, more importantly, now you have to do error checks on every but last step or join everything via '&&' ...

README.md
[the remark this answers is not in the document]
mbrandenburger: let me clarify. I tried to think about this git checkout like someone who is actively working on fabric. In their daily job they create branches, switch from one branch to another, do PRs reviews, etc ... creating the branch with checkout -b allows this person to easily switch back to there other work, but also come back and continue with FPC.

We could argue that in this situation people are responsible but creating (or suggesting it in the readme) a named branch makes life (I think) easier. For people who would never checkout Fabric otherwise this named branch would not add any value, agreed.

I don't have a strong opinion here so Morgan can decide :)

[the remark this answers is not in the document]
g2flyer: note that as-is nobody stops you to add that branch, just that nobody forces it ...

fabric/Makefile
```diff
@@ -19,18 +19,19 @@ build: native
 patch: $(FABRIC_PATCHED)
 
 $(FABRIC_PATCHED):
-	@echo "Patching Fabric ..."
-	@if [ "$(FABRIC_CURRENT_TAG)" != "$(FABRIC_VERSION)" ]; then \
-		echo "Aborting! Tag on current HEAD ($(FABRIC_CURRENT_TAG)) does not match expected tag/$(FABRIC_VERSION)!"; exit 1; \
+	echo "Patching Fabric ..."; \
```
mbrandenburger, line 22: I would bring the @echo back to avoid output the entire if statement the console

mbrandenburger, line 22:
```suggestion
	@echo "Patching Fabric ..."; \
```

fabric/Makefile
```diff
@@ -9,64 +9,58 @@ FABRIC_VERSION ?= v2.0.1
...
+	if [ ! -f $(FABRIC_PATCHED) ]; then \
+		echo "You didn't apply patches automatically. Please manually reset to ${FABRIC_VERSION}"; \
+		exit 1; \
+	fi; \
+	cd $(FABRIC_PATH); \
+	$(FABRIC_CURRENT_DIFF_COMMITS_CMD) > $(FABRIC_PATCHED).tmp; \
+	if [ ! -z "$$(diff $(FABRIC_PATCHED) $(FABRIC_PATCHED).tmp)" ]; then \
   		echo "You have other commits than the patches:"; \
-   		cd $(FABRIC_PATH) && git log $(FABRIC_VERSION)..HEAD --oneline; \
+		echo "- expected: "; cat  $(FABRIC_PATCHED); \
```
mbrandenburger, line 53:
```suggestion
		echo "- expected: "; cat $(FABRIC_PATCHED); \
```

fabric/Makefile
```diff
@@ -9,64 +9,58 @@ FABRIC_VERSION ?= v2.0.1
...
-	@echo "Removing the following commits on branch ${FABRIC_PATCHED_BRANCH} in $(FABRIC_PATH)"
-	@cd $(FABRIC_PATH) && git log $(FABRIC_VERSION)..HEAD --oneline
-
-	cd $(FABRIC_PATH) && \
-	git reset --hard $(FABRIC_VERSION)
+  	fi; \
+	\
+	echo "Removing the following fpc commits in $(FABRIC_PATH)"; \
+	$(FABRIC_CURRENT_DIFF_COMMITS_CMD); \
+	set -x;\
```
mbrandenburger, line 61:
```suggestion
	set -x; \
```

fabric/Makefile
```diff
@@ -9,64 +9,58 @@ FABRIC_VERSION ?= v2.0.1
...
-	@if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
-		echo "Aborting! Nothing to do here ... it seems you haven't applied the patches yet"; exit 1; \
-	fi
-
-	@if [ ! -z "$(FABRIC_TRACKED_CHANGES)" ]; then \
-		echo "Aborting! It seems you have some changes in your branch:"; \
+	# Note: for below to work with exit 0, this has to be a single shell command!
+	@echo "Cleaning Fabric patches ..."; \
+	if [ "$(FABRIC_CURRENT_TAG)" = "$(FABRIC_VERSION)" ]; then \
+		echo "Nothing to do here ... it seems you haven't applied the patches yet";\
```
mbrandenburger, line 36:
```suggestion
		echo "Nothing to do here ... it seems you haven't applied the patches yet"; \
```

fabric/Makefile
g2flyer: oops, that was actually the original intention. got lost in testing ...

mbrandenburger:
Looks good to merge! Please rebase/squash. Thank you!

g2flyer:
> Looks good to merge! Please rebase/squash. Thank you!

just did ...

PR #337, opened 2020-04-21
pdo-crypto-lib integration

bvavala:
it add the Private Data Objects project as a module to compile and integrate its crypto library.

In this PR, the crypto lib is only compiled and linked in some executables, but not actually used.
Packages using this library will be pushed in separate PRs.

no


g2flyer:
question: glancing through the files, presumably, private-data-objects repo will be added as a sub-module under common/crypto/pdo and you will add a corresponding `.gitmodules` stanza?

bvavala:
> question: glancing through the files, presumably, private-data-objects repo will be added as a sub-module under common/crypto/pdo and you will add a corresponding `.gitmodules` stanza?

Yes it/I will, but it's not right now -- the PR is *WIP*; as PDO adapts, I will add the submodule.

cmake/Init.cmake
```diff
@@ -7,3 +7,9 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_VERBOSE_MAKEFILE ON)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+
+if("${FPC_PATH} " STREQUAL " ")
```
g2flyer, line 11: As part of build there shouldn't be really a requirement to have FPC_PATH env defined, we should be able to determine that via TOP?  Also note that in doc we do not really require FPC_PATH to be defined, we just informally use it to characterize examples and alike ..

that said, i notice that it seems Marcus already "sinned" in the last fabric build PR which also requires now FPC_PATH.  If we want to rely on it, i think we should define it in `config.mk`.   Relying on a env variable is also a bit brittle in case you have multiple installations on same machine (e.g., one with concept release 1 and another one with master), in which case you might get to confusing situation if your make reads partially stuff with the TOP (i.e., what you expect) and partially stuff via FPC_PATH (which would be the wrong place ...)

common/crypto/types.cpp
```diff
@@ -0,0 +1,85 @@
+/* Copyright 2018 Intel Corporation
```
g2flyer, line 1: Can't we reuse the existing one from pdo (or at least refactor it)?  Or alternatively (and maybe better?)  we already "import" base64 (and hex_string) via pdo, so we could essentially dump the FPC version and just re-use the PDO version?

bvavala: > Can't we reuse the existing one from pdo (or at least refactor it)?
No, because PDO's version uses PDO's base64 package which is different.

> Or alternatively (and maybe better?) we already "import" base64 (and hex_string) via pdo, so we could essentially dump the FPC version and just re-use the PDO version?
We are not importing base64 from PDO!!! -- that's the point.
This `types.cpp` has been changed to use FPC's base64.

Let's clarify the problem. For performance reasons, PDO has/uses a modified version of the original base64 package. FPC still uses the original package. As the PDO crypto library requires `types.cpp` (which is the one using the base64 package), this has to be adapted accordingly for FPC. Most likely, in the future, we will adopt PDO's base64 package and dump the original one. This effort is not considered worthwhile, for what we are doing/need now.

bvavala:
> . Would be nice to also run the crypto tests when building here, not only when building PDO ....

Good idea, I have added a todo

common/crypto/types.cpp
g2flyer: the crypto itself does not require the base64 functions from types.cpp; so one could easily address it by splitting types.cpp into two?  Also if there is such a performance difference that we changed it in PDO, why wouldn't we want it here?

bvavala: > the crypto itself does not require the base64 functions from types.cpp; so one could easily address it by splitting types.cpp into two?

Possibly. But since the file has some functionality that might be useful in FPC, it seems premature to do that.
Also, if FPC adopt PDO's base64 package, we need to neither split nor copy/modify `types.cpp`.

> Also if there is such a performance difference that we changed it in PDO, why wouldn't we want it here?

As I said, we will "most likely" adopt that.
Reasons not to do it now:
* personally, I have not run any related performance test
* adopting that means
    * either keeping 2 base64 packages, which is ugly
    * or dropping FPC's base64 package
        * in the latter case, you have modifications to do all around FPC -- which is reasonable, but in a separate PR.

g2flyer: Hmm, but if you split types.cpp in two, you still can adopt the use in fpc later if we ever want to do so but also have a clean interface?  I have no strong feeling of going either the split-types.cpp or the one-base64-in-fpc route, i just am not excited in the version with the duplicated types.cpp.   btw: marcela probably should also weigh in on this for her graphene use-case?

cmake/Init.cmake
bvavala: I suppose I've sinned too!
The `FPC_PATH` variable is frequently handy, and clearly there is problem with `TOP`.
I would vote to keep the first and fix the second -- of course, through an issue and separate PR.
@mbrandenburger , your opinion?

g2flyer: what is the problem with TOP? But in any case, why not just define FPC_PATH in `config.mk` as suggested?  It fixes the issue by ensuring that the build (a) has the (not really documented) value and (b) it is consistent?

bvavala: The problem with TOP, as you mentioned, is consistency -- it should be the same as FPC_PATH.

I am ok with `config.mk`.

Oh you mean initializing `FPC_PATH` from `TOP`. I was thinking the opposite.
But, again as you said, I think we already need `FPC_PATH` defined somewhere else. So would this work out-of-the-box?

g2flyer: right now build fails if you do not have FPC_PATH defined. With above change to `config.mk` FPC_PATH should be defined properly during build, regardless whether you had it originally in your environment or not.

common/crypto/types.cpp
bvavala: > Hmm, but if you split types.cpp in two, you still can adopt the use in fpc later if we ever want to do so but also have a clean interface? I have no strong feeling of going either the split-types.cpp or the one-base64-in-fpc route,

I see the split really premature, since in the future I believe we will adopt the PDO's base64.

> i just am not excited in the version with the duplicated types.cpp. btw: marcela probably should also weigh in on this for her graphene use-case?

Let me provide more reasons why this is not a problem.
1) Even base64 is duplicated -- we could have it as a submodule rather than shipping it.
2) The PDO submodule will not be "master". It will be tagged with a commit id, so the code will not change.
3) I think nobody has reasons to change the interfaces of crypto, types, FPC's base64 and PDO's base64. Therefore, removing or splitting `types.cpp` can done seamlessly in the future by only updating (c)makefiles without touching any code.

g2flyer: i don't the split would have been premature. Actually to the contrary in hindsight: it would provide a clean interface to pdo crypto (which doesn't need base64 but the rest of types.cpp). But given that Mic now merged -- i guess i shouldn't have initially approved it before more of the questions popped up ..;-) --- it is now kindof a moot question.

I still don't like the types.cpp as-is but i can understand the resistance to switch right away to pdo base64.  What i would think, though, that in this case we have to add a comment in the FPC types.cpp saying that it is "modified clone" of `pdo/common/types.cpp` and why it's there.  And maybe add a TODO to revisit eventually whether we might want to eventually revisit the use of the more efficient PDO base64 variation.

cmake/Init.cmake
bvavala: will fix it -- I thought it was a precondition already!

common/crypto/types.cpp
bvavala: I still disagree on the split: I don't see why splitting `types` based on the base64 dependency would be cleaner. 

* ok for the comment
* ok for the todo to revisit it
* ok for checking the base64 variation

g2flyer: just in case it was not clear: i was talking splitting types in PDO (and change pdo crypto to use only the base64-less version of it), not here. In that case you wouldn't need this types.cpp here anymore and could use the pdo crypto version right away?  Anyway, seems now a moot point ...

.travis.yml
```diff
@@ -52,6 +52,8 @@ env:
     - PROTOC_CMD=${PROTO_DIR}/bin/protoc
 
 before_install:
+  # openssl
+  - sudo apt-get -y install libssl-dev
```
g2flyer, line 56: why don't you add that to addons->apt->packages further up?

bvavala:
> > . Would be nice to also run the crypto tests when building here, not only when building PDO ....
> 
> Good idea, I have added a todo

@g2flyer it turns out that the tests in PDO need some refactoring:
* they rely on common
* they use pdo's base64 package
* they are under `tests` where also `state` tests are called, which need common.
I will open an issue for this (to document that we need crypto tests).
Anyway, for the moment, it would not be much of a gain to run those tests here too.

cmake/ConfigSGX.cmake
```diff
@@ -30,7 +30,7 @@ endif()
 
 set(SGX_SSL "$ENV{SGX_SSL}")
 if("${SGX_SSL} " STREQUAL " ")
-       set(SGX_SSL /opt/intel/sgxssl)
+    message(FATAL_ERROR "SGX_SSL: undefined environment variable")
```
g2flyer, line 33: as discussed oob, having defaults only globally in `config.mk` is the right way, so i agree with this change. However, i think we should apply it uniformaly also for the other defaults here; certainly for the ones which are in config.mk (e.g, SGX_MODE and SGX_BUILD) but probably also for SGX_ARCH and SGX_SDK)

bvavala:
> I really was too liberal in the approval for the PDO PR ;-)

No, you were not.

>I somewhat expected that there might be some smaller issues we might discover via this PR.

It is not an issue:
* the adopted library is already tested as part of PDO
* FPC has currently no (different) tests for the crypto part
* any test _can_ already be implemented separately now

> I agree that it's not worth here to run all other tests (as this almost certainly brings other unwanted sw and config dependencies from PDO). We should keep this in mind though when touching any related part in PDO again (similar to the types.cpp issue)

yes, and yes -- I have a todo to document this.

bvavala:
> Looks in general good, just the suggested consistency of defaults in config.mk and i'm happy :-)

Really in this PR?
I will make you happy as part of the next one - these modifications have nothing to do with the crypto lib integration.

g2flyer:
> > Looks in general good, just the suggested consistency of defaults in config.mk and i'm happy :-)
> 
> Really in this PR?
> I will make you happy as part of the next one - these modifications have nothing to do with the crypto lib integration.

oh, c'mon. You already touched the two files and did half of the job....

cmake/Init.cmake
mbrandenburger: Having FPC_PATH in config.mk makes sense for me! :) I like it

.travis.yml
```diff
@@ -28,7 +28,7 @@ env:
     # SGX SSL
     - OPENSSL_VERSION=1.1.0j
     - SGXSSL_VERSION=v2.4.1
-    - SGXSSL=/opt/intel/sgxssl
+    - SGX_SSL=/opt/intel/sgxssl
```
mbrandenburger, line 31: I like `SGX_SSL` over `SGXSSL` but it introduces a small inconsistency with other *SSL vars here in this travis file.

cmake/Init.cmake
bvavala: * FPC_PATH is defined in config.mk
* Init.cmake checks that it exists
@g2flyer  would you double check again that this is fine?

.travis.yml
bvavala: This was actually misnamed in travis, because other parts of the project use `SGX_SSL`.
And this was the cause that made travis fail, because the pdo crypto lib requires it to compile properly.

bvavala:
> I would like to see a "small" test in ecc_enclave which uses the new crypto lib. Something easy just to have some code usage that shows that the integration is successful.

I have added a code snippet to use the crypto library.
Here some notes:
* the snippet simply creates a keys and dumps the PEM encoding;
* the snippet will be removed
* `common/types.h` has been renamed to `fpc-types.h`

> Also, would it make sense to hide the pdo crypto behind an fpc crypto interface? So we may have the chance to swap out pdo dependency if needed in the future?

This is the exact reason why we created the pdo crypto lib in the first place. It's actually a wrapper of openssl, that's it! So unless you want to create a wrapper-of-a-wrapper, I would suggest taking a quick look at its headers.

> Last, let's pull the commits from #338 here as this makes it easier to review.

This PR is self-contained and does not need 338.

cmake/Init.cmake
g2flyer: That's what we discussed above and is consistent with the change to SGX_SSL and the other half of related changes from #338.   I actually thought, this would be already in an earlier commit; if not, even better than it is now here :-)

bvavala: I had removed it completely but... I still needed to define the cmake-internal FPC_PATH from the one defined in the environment.
Anyway... solved!

ecc_enclave/enclave/enclave.cpp
```diff
@@ -224,6 +228,23 @@ int ecall_cc_invoke(const char* encoded_args,
...
+    // NOTE/TODO: this crypto part will be removed
+    if(!crypto_created)
+    {
+        pdo::crypto::sig::PublicKey verification_key_;
+        pdo::crypto::sig::PrivateKey signature_key_;
+        signature_key_.Generate(); //private key
+        verification_key_ = signature_key_.GetPublicKey();
+        //debug
+        std::string s = verification_key_.Serialize();
+        LOG_DEBUG("enclave verification key: %s", s.c_str());
```
mbrandenburger, line 240: Have you actually ever checked if this serialised key can be used plain go crypto? We want to use the pdo crypto lib to have only a single crypto implementation used everywhere inside the enclave and outside using some wrapper in go. However, we haven't really checked with the fabric community if this is fine OR if we insist of a go implementation where ever possible.

mbrandenburger: In particular, we should double check that we can use this crypto lib in our custom ecc-vscc.

ecc_enclave/sgxcclib/sgxcclib.h
```diff
@@ -8,7 +8,7 @@
 #ifndef _SGXCCLIB_H_
 #define _SGXCCLIB_H_
 
-#include "types.h"
+#include "fpc-types.h"
```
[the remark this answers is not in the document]
bvavala: the problem with `type.h` is that the code includes the file directly -- without the path, because the path is include at link time. Same in PDO.

ok for fixing the define
ok for renaming the cpp -- but the header is still a problem because of the include in pdo

ecc_enclave/enclave/enclave.cpp
bvavala: > Have you actually ever checked if this serialised key can be used plain go crypto? We want to use the pdo crypto lib to have only a single crypto implementation used everywhere inside the enclave and outside using some wrapper in go. However, we haven't really checked with the fabric community if this is fine OR if they insist of a go implementation where ever possible.

I have not tested, but my answer is yes. The serialized key is in PEM format: `-----BEGIN Type-----Headers\nbase64-encoded Bytes-----END Type-----`. So you can use the `encoding/pem` package to encode/decode them, and then use `crypto`.

> In particular, we should double check that we can use this crypto lib in our custom ecc-vscc.

As the plugin will not verify the attestation, we could simply resort to crypto to verify the signature.

PR #338, opened 2020-04-25
Move default values to config.mk

bvavala:
(this PR complements #337 )
move the definition of some default values (directory paths) to config.mk

these default value are set in several places, this is a first pass to centralize these configuration parameters


cmake/ConfigSGX.cmake
```diff
@@ -10,27 +10,27 @@ execute_process(COMMAND getconf LONG_BIT OUTPUT_VARIABLE VAR_LONG_BIT)
...
 if("${SGX_BUILD} " STREQUAL " ")
-       set(SGX_BUILD PRERELEASE)
+    message(FATAL_ERROR "SGX_BUILD: undefined environment variable")
 endif()
 
 set(SGX_SSL "$ENV{SGX_SSL}")
 if("${SGX_SSL} " STREQUAL " ")
-       set(SGX_SSL /opt/intel/sgxssl)
+    message(FATAL_ERROR "SGX_SSL: undefined environment variable")
 endif()
```
MHBauer, line 34: I like these changes. Making it fatal might be something to do in the toplevel file depending on where it's used.

cmake/ConfigSGX.cmake
```diff
@@ -10,27 +10,27 @@ execute_process(COMMAND getconf LONG_BIT OUTPUT_VARIABLE VAR_LONG_BIT)
 # set them to the usual defaults.
```
MHBauer, line 10: needs to change as it's failing instead of defaulting.

cmake/ConfigSGX.cmake
bvavala: I am not following. This file is supposed to be a toplevel one that should be include in cmake-based sgx builds.

cmake/ConfigSGX.cmake
bvavala: missed... fixed!

cmake/ConfigSGX.cmake
```diff
@@ -5,32 +5,31 @@
 # get arch information
 execute_process(COMMAND getconf LONG_BIT OUTPUT_VARIABLE VAR_LONG_BIT)
 
-# Set variables for common SGX variables.
-# if they are defined in the environment, take them from there, otherwise
-# set them to the usual defaults.
+# Check for the existence of common required SGX variables,
+# and fail if these are not defined.
```
g2flyer, line 9:
```suggestion
# and fail if these are not defined. Note global build defaults are defined `../config.mk`.
```

bvavala: yes! -- I suppose the `co-author` line is not recognized as a sign-off. Disappointing!

cmake/ConfigSGX.cmake
MHBauer: is regular make not ever involved in the non-cmake builds?

bvavala: it is. Actually the build starts with the regular make, which includes `config.mk` (where the variables are defined). Then make calls cmake, which inherits these definitions. The first thing  cmake files do (in sgx builds) is to include this `ConfigSGX.cmake`, which checks for the existence of the required definitions.

mbrandenburger:
I also think moving this little commits to PR #337 would it make easier to review.

bvavala:
> @bvavala What is the strategy here? Should we block this PR until #337 is merged? In that case redundant commits should be cleaned out.

correct

bvavala:
commits ported to #337

PR #339, opened 2020-04-28
Concept Release (1.0) with versioned images & able out of the box to run demo

### LINKED ISSUE #314: Docker-dev is missing killall command
**Describe the bug**
The test in demo/chaincode/fpc requires `killall` command to kill the mock server after tests are finished. Our docker dev-image misses that command.

**To Reproduce**
```
cd utiles/docker
make dev; make run
cd demo/chaincode/fpc
make
```

**Expected behavior**
```
Test successful.
root@7984cc9a34dc:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc#
```

**Log-filesScreenshots**
```
Test successful.
./test.sh: line 11: killall: command not found
Makefile:48: recipe for target 'test' failed
make: *** [test] Error 127
root@7984cc9a34dc:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc#
```

MacOS using docker-dev images from master (42b34e4)

**Additional context**

g2flyer:
This PR does make sure that `concept release` (first edition) will build properly from a fresh install and allow running the demo via UI and docker-compose fpc network (note an non-interactive variant of the demo based on fpc network is exercised as part of the `test` makefile target in `demo/client/scripting`!).  It should work for both `SIM` and `HW` SGX mode and both directly on host _and_ inside a `dev` container.

Additionally, it will also version the docker images (via tag) and have a unified docker image naming scheme as well includes a variety of robustness and other smaller improvements related to make build/cleanup. The basic approach is to define `FPC_VERSION` in `config.mk` and then tag all container images based on it.  For concept release 1.0.1, i've used the convention `cr1.0.1`, deviating from fabric's `v2.0.1` but i wanted to distinguish concept releases from future complete versions (which i would then expect to follow the fabric naming).  For non-tagged code i would expect the value to be `master` ...)

Note: 
* this is a PR on a newly created branch concept-release-1.0 which currently is rooted in the `concept-release` tag.  As discussed on call, for future consistency, i added the version info to the branch.  The idea would be to remove the original `concept-release` tag and replace it with a `cr1.0.1` tag on the then-HEAD of branch.  
* Given the importance of the dev container for Fabric integration i've added to travis also an end-to-end build/test within a freshly built dev container. This will the build-time, though again by roughly 50%.  If this is too much of a concern, then one could test only inside dev container and skip the host-based full build/test. This probably should reduce it back to slightly above what it was before.  That said, travis on a PR shouldn't really be your test platform but only a validation of locally performed tests. So the corresponding time shouldn't be that crucial ...

Fixes completely  #314, also fixes (once also applied to master) #345 & #346 ; makes a big step towards #348; provides short-term fix for #269.

The PR is a bit overkill for the concept release but is intended to be also then (in a separate PR) to be applied to master.

Regarding testing it, you should be able with a freshly checked out PR to build and run the demo (non-interactive and interactive via UI, with fpc-network or mock) as documented in the UI.  I have tested on Linux in VM and on HW with sgx support, both on directly on host and inside `dev` container. For demo it should also work when optionally also requesting explorer and couchdb for demo (see section on `USE_EXPLORER` and `USE_COUCHDB` in [utils/docker-compose/README.md](utils/docker-compose/README.md).  However, as i do not have access to a Mac, i couldn't test on a mac.  However, i think the changes should also work on a Mac.

One caveat: currently the dev container runs as root and hence generated files outside the container might or might not be accessible depending on what user id you run outside.  Iff you run a `make clean` and `./utils/docker-compose/scripts/teardown.sh --clean-slate` inside the container before quitting it you should be fine (if not, you might have to do some `sudo`-based cleanup ...).  This might also be an issue only on linux as Mac apparently uses NFS for volume mapping and supposedly does not have that problem. However, not having a mac i couldn't confirm that.   Unfortunately, i couldn't find immediately an easy solution (e.g., just running with `--user` does _not_ work ....) so i left this out for a later PR.  This should also not matter here for the concept-release-1.0 variant, more an issue for the later adoption to master of this PR ....

As discussed on rocket-chat, it also adds a few (more precisly, 8) commits from master to incorporate the commit which makes the avatar load more robust.  So the review should be primarily over the commits from April 27th onwards.

Also as mentioned also below, one of the old commits has a DCO which confuses github.

it shouldn't other than other than that it should work now :-)


g2flyer:
btw: dco fails because of an old commit which already existed on master ...

README.md
```diff
@@ -474,7 +484,7 @@ By running the following command you can generate the documentation.
 ## Getting Help
 
 Found a bug? Need help to fix an issue? You have a great idea for a new feature? Talk to us! You can reach us on
-[RocketChat](https://chat.hyperledger.org/) in #private-data-objects.
```
MHBauer, line 477: yay

MHBauer:
Looks like we shoudl focus on the last four commits, as the others are a roll-up of previous work.

g2flyer:
> Looks like we shoudl focus on the last four commits, as the others are a roll-up of previous work.

I mentioned that in the description but i should have explicitly pointed out which commits are new. Thanks for doing so!

g2flyer:
> Can you please check for spaces! Your PR was the motivation enabling linting again :)

I've accepted your changes. But honestly, i don't think these spaces are that crucial to have consistent that i want to spend much time on it  (if spaces change it's a different story but that was not the case here?)

utils/docker-compose/scripts/lib/common.sh
```diff
@@ -16,6 +16,7 @@ export SGX_MODE=${SGX_MODE:=SIM}
 
 # Variables which we allow the caller override ..
 export FABRIC_VERSION=${FABRIC_VERSION:=1.4.3}
+export FPC_VERSION=${FPC_VERSION:=latest}
```
MHBauer, line 19: probably should be FPC_IMAGE_VERSION

utils/docker/dev/Dockerfile
```diff
@@ -10,7 +10,7 @@ ARG GO_VERSION=go1.13
 ARG GO_TAR=$GO_VERSION.linux-amd64.tar.gz
 ARG NANOPB_VERSION=0.3.9.2
 ARG FABRIC_REPO=https://github.com/hyperledger/fabric.git
-ARG FABRIC_VERSION=v1.4.3
+ARG FABRIC_VERSION=1.4.3
```
MHBauer, line 13: why move the 'v' around?

MHBauer: git branches are tagged with v, and docker images are tagged with no v? This is a problem for what?

demo/Makefile
```diff
@@ -7,9 +7,9 @@ include $(TOP)/build.mk
...
 SUB_DIRS = chaincode client
 
-build test clean clobber:
+build test clean:
 	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
 
-
 clobber:
 	./scripts/teardown.sh --clean-slate
+	$(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
```
MHBauer, line 15: I'm not sure I've used clobber before?

utils/docker/dev/Dockerfile
```diff
@@ -11,15 +11,18 @@ ARG FABRIC_REPO=https://github.com/hyperledger/fabric.git
 ARG FABRIC_VERSION=v1.4.3
 ARG FPC_REPO_URL=https://github.com/hyperledger-labs/fabric-private-chaincode.git
 ARG FPC_REPO_BRANCH=master
+# Note: while above says _BRANCH, it also works with a commit id or tag
```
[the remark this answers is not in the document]
g2flyer: It still works with branches, but i guess i can change to `FPC_REPO_BRANCH_OR_COMMIT` to make it more explicit

ecc/Makefile
```diff
@@ -86,6 +88,7 @@ docker-fpc-app: docker-boilerplate-ecc
 		--build-arg BOILERPLATE_EXTENSION=$(BOILERPLATE_EXTENSION)\
 		--build-arg enclave_so_path=$(DOCKER_ENCLAVE_SO_PATH)\
 		..
+	# tagging here, this is application image ..
```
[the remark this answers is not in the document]
g2flyer: a 'no ' was missing, but i guess i just drop the comment ...

utils/docker-compose/scripts/lib/common.sh
g2flyer: why would you think so?  We also say FABRIC_VERSION?

utils/docker/dev/Dockerfile
g2flyer: FABRIC_VERSION var is used for both cases you mention: much easier to add the 'v' where needed than to remove it where not

demo/Makefile
g2flyer: only you can answer that ;-)   But the goal of this target is to reset the build to a true clean-slate with all build/downloaded artifacts removed.

utils/docker/dev/Dockerfile
```diff
@@ -72,12 +75,16 @@ RUN go get github.com/golang/protobuf/proto \
...
 # Note: could add --single-branch to below to speed-up and keep size smaller. But for now for a dev-image better keep complete repo
 
 # Get FPC
 ENV FPC_PATH=$GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
-RUN git clone --branch $FPC_REPO_BRANCH $FPC_REPO_URL $FPC_PATH
+# We copy context so we can use that to potentially get local .git as repo ...
+COPY .git /tmp/cloned-local-fpc-git-repo
+RUN git clone $FPC_REPO_URL $FPC_PATH \
+  && cd $FPC_PATH \
+  && git checkout $FPC_REPO_BRANCH_TAGo_OR_COMMIT
```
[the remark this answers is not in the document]
g2flyer: arghh, not sure how this got here -- i've tried to globally replace with sed -i ... -- and why my tests didn't catch it :-(  Thanks for catching this!

utils/docker/Makefile
```diff
@@ -76,32 +94,48 @@ DOCKER_BUILD_OPTS ?=
...
 
 dev: base
-	$(DOCKER) build $(DOCKER_BUILD_OPTS) -t $(FPC_DOCKER_DEV_NAME) $(DOCKER_DEV_BUILD_OPTS) dev
+	(cd ${TOP}; \
+         $(DOCKER) build $(DOCKER_BUILD_OPTS) -t $(FPC_DOCKER_DEV_NAME)\
+         $(DOCKER_DEV_BUILD_OPTS)\
+         -f utils/docker/dev/Dockerfile\
+         --build-arg FPC_REPO_URL=file:///tmp/cloned-local-fpc-git-repo\
+         --build-arg FPC_REPO_BRANCH_TAG_OR_COMMIT=$$(git rev-parse HEAD)\
+         --build-arg SGX_MODE=${SGX_MODE}\
```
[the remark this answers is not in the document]
g2flyer: I didn't change anything when dev image docker build is called (although tried to minimize what is actually build rather than re-used from cache).  What fails in your case though shouldn't be the build but the run.  I actually did add at some time could which should have removed  a container with name fpc-development-${FPC_VERSION} iff it is dormant (you clearly don't want to kill a running one ...) but somehow it got lost; will (re-)add ...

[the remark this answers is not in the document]
g2flyer: fyi: corresponding fixup is uploaded

utils/docker/dev/Dockerfile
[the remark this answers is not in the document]
g2flyer: fyi: corresponding fixup is uploaded

Jordy24:
I have tested PR #339 on mac using the docker dev container and the demo is running fine.... I execute all the scripts found in the demo readme inside the container.

# Testing (Starting in the Fpc_path directory):
make clean
cd /utils/docker 
make dev
make run
------- Once inside the dev container 
./utils/docker-compose/scripts/teardown.sh --clean-slate
cd /demo
scripts/startFPCAuctionNetwork.sh --build-cc

g2flyer:
> I have tested PR #339 on mac using the docker dev container and the demo is running fine.... I execute all the scripts found in the demo readme inside the container.

Did you also try to UI part of the demo, i.e., execute interactively a test auction scenario? Curious whether that works properly.  also would be curious you got into permission problems if you exit the container (without cleaning up) and try to restart the container. On linux this gives permission problems but the mac does volume mounts differently, so might not be an issue there.  Of course depends on how you start the container: are you root on the host or run it as a non-root user?

If that all works fine and you are fine with the actual code, then of course i'd welcome also an "approve" of the PR :-)

Jordy24:
so the UI works fine.... and I was not getting any issue after exiting and the restarting the container again

MHBauer:
> I have tested PR #339 on mac using the docker dev container and the demo is running fine.... I execute all the scripts found in the demo readme inside the container.
> # Testing (Starting in the Fpc_path directory):
> 
> make clean
> cd /utils/docker
> make dev
> make run
> ------- Once inside the dev container
> ./utils/docker-compose/scripts/teardown.sh --clean-slate
> cd /demo
> scripts/startFPCAuctionNetwork.sh --build-cc

This worked!

g2flyer:
> overall looks good.
> Two comments:
> 
> 1. the pr adds: 13min more build time due to demo; in travis, it also add 21min to test everything again in the dev container -- so travis needs 50min overall to complete. As suggested, we might remove the original build, and rely on the test performed through the dev container.

I disagree regarding the 13 mins.  This was in already before in concept release (just not present in master as we disabled).  Some of the docker-image restructuring should have improved slightly the build process as more is shared between dev and peer which docker can re-use (practically more importantly, also, the docker reuse should be now faster also for re-builds where you don't check the git history). 

> 2. the build in the dev container crashes in several places _in travis_. This seems only due to docker credentials (same issue is mentioned here [docker/docker-credential-helpers#103](https://github.com/docker/docker-credential-helpers/issues/103)). Curiously it does not make travis fail, and the test still completes as expected.

I see that also locally on many machines and have seen it for a long time (both local and travis), so i guess i just ignore it as functionally everything still works.  From PR it seems as if going to a newer docker-compose directly from docker might remove it, but i don't think it is worth the change?  (Certainly not in this PR :-)

g2flyer:
> I disagree regarding the 13 mins (in terms of _add in the pr._) This was in already before in concept release (just not present in master as we temporarily disabled all of demo as part of concept-release 2.0). Some of the docker-image restructuring should have improved slightly the build process as more is shared between dev and peer which docker can re-use (practically more importantly, also, the docker reuse should be now faster also for re-builds where you don't check the git history).

Seems Travis forced our hands now, though, as it hit the 50 min timeout.  (Strangely, it did _not_ killed the build in prior builds which also took (slightly) more than 50 mins. Anyway, added now a commit which does only the build/test via dev container (which hopefully should be a super-set of the build/test verification compared to do it directly on the host ...)

MHBauer:
> One caveat: currently the dev container runs as root and hence generated files outside the container might or might not be accessible depending on what user id you run outside. Iff you run a `make clean` and `./utils/docker-compose/scripts/teardown.sh --clean-slate` inside the container before quitting it you should be fine (if not, you might have to do some `sudo`-based cleanup ...). This might also be an issue only on linux as Mac apparently uses NFS for volume mapping and supposedly does not have that problem. However, not having a mac i couldn't confirm that. Unfortunately, i couldn't find immediately an easy solution (e.g., just running with `--user` does _not_ work ....) so i left this out for a later PR. This should also not matter here for the concept-release-1.0 variant, more an issue for the later adoption to master of this PR ....

This permissions thing might be solvable with judicious use of something like `  --user "$(id -u):$(id -g)" `, but I am not sure how compose interacts with it.

g2flyer:
> This permissions thing might be solvable with judicious use of something like ` --user "$(id -u):$(id -g)"`, but I am not sure how compose interacts with it.

I've tried that but as mentioned that does not work by itself. It will give you then permission problems inside the container (e.g., fabric patching, .gocache in HOME ...)

PR #342, opened 2020-04-28
Enable linter again

mbrandenburger:

This PR brings back go and cpp linting. Also some minor changes to make linting happy again with the current code base.

No issue created for this.

Just run make checks.

Developers need to be careful when submitting PRs and make sure linting checks are passing.


Makefile
```diff
@@ -26,11 +26,11 @@ license:
 
 linter: gotools build
 	# TODO enable linter again
-	#@echo "LINT: Running code checks for Go files..."
-	#@cd $$(/bin/pwd) && ./scripts/golinter.sh
+	@echo "LINT: Running code checks for Go files..."
+	@cd $$(/bin/pwd) && ./scripts/golinter.sh
```
MHBauer, line 29: not sure I like that the implementation links to something in fabric directory. Can we add the linter to our own repo?

Makefile
```diff
@@ -34,7 +34,7 @@ linter: gotools build
 
 gotools:
 	# install goimports if not present
-	$(GO) install golang.org/x/tools/cmd/goimports
+	cd tools && $(GO) install golang.org/x/tools/cmd/goimports
```
MHBauer, line 36: wat?

there isn't a tools directory.

tools/tools.go
```diff
@@ -1,11 +1,11 @@
+// +build tools
+
```
MHBauer, line 2: oh man, did mess this up? Is it required to be at the top?

MHBauer: from https://golang.org/pkg/go/build/#hdr-Build_Constraints
"near the top of the file, preceded only by blank lines and other line comments. "

bleh, oops, my bad.... bloock comments are not line comments, I suppose.

What was the effect? Is this breaking a build?

tlcc/tlcc_test.go
```diff
@@ -9,49 +9,72 @@ package tlcc
 import (
```
MHBauer, line 9: I feel like these changes should be separate unless they fail linting.

Makefile
g2flyer: but then we have to maintain the consistency with the rest of fabric? At least right now before a everything-is-go-modularized, the symlink works rather safely?

.travis.yml
```diff
@@ -80,4 +80,4 @@ script:
   # travis doesn't check out the branch itself but for our make to work we
   # need a branch or alike. So check create a local one ...
   - git checkout -b we-need-a-branch-for-below-to-work
-  - make
+  - make checks build test clobber
```
g2flyer: why this change?  Is it to test 'clobber'?  If so, `make all clobber` might be better as it guaranteed that it does all of the default builds?  Certainly no reason to change the order here over the standard build?

tlcc/tlcc_test.go
g2flyer: (Referring to Morgan's comment which i kindof agree) I guess that also applies to above change to `tlcc/Makefile` ..

MHBauer:
looks like some stuff needs to be excluded from the linter
```
In file included from ecc/enclave/enclave_stub.go:38:0:

ecc/enclave/../../common/sgxcclib/common-sgxcclib.h:11:10: fatal error: types.h: No such file or directory

 #include "types.h"

          ^~~~~~~~~

compilation terminated.

# github.com/hyperledger-labs/fabric-private-chaincode/ecc/enclave

In file included from ecc/enclave/enclave_stub.go:38:0:

ecc/enclave/../../common/sgxcclib/common-sgxcclib.h:11:10: fatal error: types.h: No such file or directory

 #include "types.h"

          ^~~~~~~~~

compilation terminated.

# github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave

In file included from tlcc/enclave/enclave_stub.go:19:0:

tlcc/enclave/../../common/sgxcclib/common-sgxcclib.h:11:10: fatal error: types.h: No such file or directory

 #include "types.h"

          ^~~~~~~~~

compilation terminated.

# github.com/hyperledger-labs/fabric-private-chaincode/tlcc/enclave

In file included from tlcc/enclave/enclave_stub.go:19:0:

tlcc/enclave/../../common/sgxcclib/common-sgxcclib.h:11:10: fatal error: types.h: No such file or directory

 #include "types.h"

          ^~~~~~~~~

compilation terminated.

# github.com/hyperledger-labs/fabric-private-chaincode/plugins/ercc-vscc

vet: plugins/ercc-vscc/ercc_validation_logic.go:37:19: GetVerifier not declared by package attestation

Makefile:28: recipe for target 'linter' failed

make: *** [linter] Error 2
```

Makefile
mbrandenburger: The main reason why we symlink fabric linters and reuse them is that we dont need to develop our own script nor maintain a copy of the fabric script. This way we can also say that our code passes *all* fabric test and meets the code standard :D

Makefile
mbrandenburger: uhhh this one is tricky ... I moved tools.go to tools as the fabric linter does not work when go files are in the root dir.

tools/tools.go
mbrandenburger: without moving this to the top `go vet` complained about bad import.

Makefile
g2flyer: Below it is added as tools.go is moved. But do we really need a new directory `tools`, if we don't want it on the top-level (for which i see some argument), why not move it to `utils`?

tlcc/tlcc_test.go
mbrandenburger: well ... all the changes I made in this PR are to fix linting

g2flyer: hmm, lint usually shouldn't force you to change the behaviour ;-?  At least the test seems to be changed beyond what i would have expected to just make lint work ...  Now i'm not religious of not piggibacking some small change on a PR, but would be good if you say something about the changed/added behaviour?

.travis.yml
mbrandenburger: already removed!

Makefile
mbrandenburger: I dont have a strong opinion if we move this to tools/, utils/, or even utils/tools. I guess if we move it to a dedicated tool folder we can fix tool dependencies easily. 

BTW fabric also has it in tools/, see here https://github.com/hyperledger/fabric/tree/master/tools

tlcc/tlcc_test.go
mbrandenburger: I agree that tlcc test fixing is not small anymore. Will create a separate PR for this once linter is enabled again.

Makefile
g2flyer: what's the difference between utils and tools?  all stuff in utils are tools?  So i don't see the reason to have an additional directory....

mbrandenburger: moved to utils

mbrandenburger:
Please have a look at this PR. Note that this PR contains already some fixups and revert of the enable-tlcc-testing commit. Will be removed after review when rebasing and squashing.

go.mod
```diff
@@ -28,3 +28,5 @@ require (
 	google.golang.org/grpc v1.28.0 // indirect
 	gopkg.in/ini.v1 v1.55.0 // indirect
 )
+
+replace github.com/hyperledger-labs/fabric-private-chaincode/ercc => ./ercc
```
g2flyer, line 32: What's the reason to do this?

Makefile
```diff
@@ -24,17 +24,16 @@ license:
...
 	@scripts/check_license.sh
 
-linter: gotools build
-	# TODO enable linter again
-	#@echo "LINT: Running code checks for Go files..."
-	#@cd $$(/bin/pwd) && ./scripts/golinter.sh
-	#@echo "LINT: Running code checks for Cpp/header files..."
-	#@cd $$(/bin/pwd) && ./scripts/cpplinter.sh
-	#@echo "LINT completed."
+linter: gotools
```
g2flyer, line 27: hmm, why did you remove the dependency on build?  Now a `make linter` will fail unless you have done a build ...

go.mod
mbrandenburger: attestation.GetVerifier in the ercc/attestation package is only available when we build the provider.go. This leads to the problem that go vet complains that getVerifer is not defined in the package as it checks aginst the go-mod dependency. When pointing to ercc directly all checks pass when provider.go exists.

Makefile
mbrandenburger: Removed build dep for linter target as build is invoked by travis anyway and when using linter locally forcing a complete build just for running linting I feel is not good. I would argue when a developer wants to run linting, they should be able to run it with the current code without build components.

g2flyer: This would be fine if target would run successfully on non-build code. However it fails!

go.mod
g2flyer: hmm, but essentially means our module is kindof broken as it should build self-contained?  But then maybe this is only temporary and the generation of provider.go is eventually going away? (btw: i forgot, what was actually the reason that we used templates rather than go build tags which would side step the problem as all files would exist out-of-the-box?)

mbrandenburger: We used the templates to control wether we want to use mock attestation or the real one. ERCC used to be built by fabric and therefore we had no chance to add extra build commands regarding what attestation we want to use, therefore we generated the provider.go before and supplied it when installing/building the chaincode.

I think with external launchers we can overcome this issue as we can control how ercc is built. So we can use build tags.

Makefile
```diff
@@ -24,17 +24,16 @@ license:
...
+	@echo "LINT: Running code checks for Go files..."
+	@cd $$(/bin/pwd) && ./scripts/golinter.sh
+	@echo "LINT: Running code checks for Cpp/header files..."
+	@cd $$(/bin/pwd) && ./scripts/cpplinter.sh
+	@echo "LINT completed."
 
 gotools:
 	# install goimports if not present
-	$(GO) install golang.org/x/tools/cmd/goimports
+	cd utils && $(GO) install golang.org/x/tools/cmd/goimports
```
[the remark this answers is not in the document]
mbrandenburger: aha! even better so removed `cd` and I suggest to keep gotools where it is right now and refactor later if necessary.

Makefile
mbrandenburger: added build dependency back. Still I think it is not useful and it's ok that it fails when a developer does not run make build before.

go.mod
g2flyer: We definitely could do that using external builder and might have to do that for even more fundamental reasons iff ercc relies on C-based crypto/attestation libraries.   Will probably require though to add a new chaincode-lang `fpc-ercc` or alike so the external builder can distinguish it from other go code and, correspondingly, also would need to be added to the peer-cli wrapper as a second package special case (with `fpc-c` being the first one ...)

PR #343, opened 2020-04-30
(Hopefully) improve usability of external builder state & log-file

g2flyer:

* Clean-up them up on the next ledger init
* add also (parts of) chaincode id to directory name

No issue but should address the complaint from some users ...

After you build, the easiest way to test/check the behaviour is to run some integration tests and monitor `/tmp/` for presence (or absence) of `fpc-...` files ...

Nope

### DIFF
--- a/fabric/bin/lib/common_ledger.sh
+++ b/fabric/bin/lib/common_ledger.sh
@@ -108,9 +108,21 @@ ledger_init() {
     pushd ${FABRIC_CFG_PATH}
 
     # 1. clean up any prior state
+    #     - ledger state and log files from peer & orderer
     [ ! -z "${FABRIC_STATE_DIR}" ] || die "FABRIC_STATE_DIR not defined" # just in case ..
     try mkdir -p ${FABRIC_STATE_DIR}
     try rm -rf ${FABRIC_STATE_DIR}/*
+    #     - external builder artifacts
+    #       Note: FABRIC_STATE_DIR which will be ledger specific and allows for
+    #         multiple concurrent ledgers (assuming their core.yaml have 
+    #         different fileSsystemPaths) these artifacts are also shared 
+    #         _across_ ledger definitions (core.yaml).
+    #         Contrary to that, external builder are ledger instance agnostic, hence we also
+    #         provide an opt-out for the wiping to allow running current ledgers by
+    #         defining an environment variable FPC_SKIP_EXTBUILDER_CLEANUP ..
+    if [ -z ${FPC_SKIP_EXTBUILDER_CLEANUP+x} ]; then
+    	try rm -rf /tmp/fpc-extbuilder.*
+    fi
 
     try ${FABRIC_BIN_DIR}/cryptogen generate --config=./crypto-config.yaml --output="${FABRIC_STATE_DIR}/organizations"
     # 2. start orderer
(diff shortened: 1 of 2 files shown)


fabric/bin/lib/common_ledger.sh
```diff
@@ -108,9 +108,20 @@ ledger_init() {
...
     try mkdir -p ${FABRIC_STATE_DIR}
     try rm -rf ${FABRIC_STATE_DIR}/*
+    #     - external builder artifacts
+    #       Note: FABRIC_STATE_DIR which will be ledger specific and allows for
+    #         multiple concurrent ledgers (assuming their core.yaml have 
+    #         different fileSsystemPaths) these artifacts are also shared 
+    #         _across_ ledger definitions (core.yaml).
+    #         Contrary to that, external builder are ledger instance agnostic, hence we also
+    #         provide an opt-out for the wiping to allow running current ledgers ..  
+    if [ -z ${FPC_SKIP_EXTBUILDER_CLEANUP+x} ]; then
```
bvavala, line 122: so with this setting, it cleans up the tmp folder at every execution of the integration test. right?

fabric/externalBuilder/bin/lib/common.sh
```diff
@@ -68,23 +67,24 @@ cc_build_for_docker() {
...
     [ -f "${CC_BUILD_DIR}/${METADATA_FILE}" ] || die "no metadata file '${METADATA_FILE}'"
     SGX_MODE="$(jq -r .sgx_mode "${CC_BUILD_DIR}/metadata.json")"
     [ ! -z "${SGX_MODE}" ]                       || die "SGX mode not specified in metadata file"
 
     [ -f "${CC_RT_METADATA_DIR}/chaincode.json" ] || die "chaincode.jsaon does not exist"
     export CORE_CHAINCODE_ID_NAME="$(jq -r .chaincode_id "${CC_RT_METADATA_DIR}/chaincode.json")" || die "could not extract chaincode-id"
+
+	
+    SHORT_CHAINCODE_ID_NAME=$(echo ${CORE_CHAINCODE_ID_NAME} | sed 's/\(.*:.\{5\}\).*/\1/')
+    RUN_STATE_DIR="/tmp/fpc-extbuilder.${SHORT_CHAINCODE_ID_NAME}.$$"
```
bvavala, line 81: if the cleanup variable is not set, this does not help -- the problem is the pid!
As we discussed, the timestamp should solve this:
```suggestion
    RUN_STATE_DIR="/tmp/fpc-extbuilder.${SHORT_CHAINCODE_ID_NAME}.$(date +%s%N)"
```
(I have not tested the suggestion)

fabric/bin/lib/common_ledger.sh
g2flyer: unless you set the corresponding `FPC_SKIP_EXTBUILDER_CLEANUP` env variable, it will clean-up

fabric/externalBuilder/bin/lib/common.sh
g2flyer: actually, you already _do_ have a timestamp with the directory entry (e.g., `ls -ltra /tmp/fpc*` will give you the most recently used one.   I thought you would be fine with that.   This gives though slightly different info than above (least recently touched vs creation time) so both could be useful.  For date format i would suggest though `date '+%Y-%m-%d_%H:%M:%s'` which gives more meaning output.   Also, for sort-order, you want `<name, date>` not `<date,name>` ?

bvavala: I had misunderstood, but yes, I am fine.

FYI: <name, date> is also fine to determine the last one per instance.

fabric/bin/lib/common_ledger.sh
bvavala: sounds good

PR #344, opened 2020-05-01
exclude pdo from linter and fix indentation

bvavala:


scripts/cpplinter.sh
```diff
@@ -24,6 +24,7 @@ EXCLUDED_PATTERNS="\
     /_build \
     .pb. \
     /node_modules \
+    common/crypto/pdo
```
g2flyer, line 27:
```suggestion
    common/crypto/pdo \
```
i guess in this particular case it does work without but for general robustness i think the \ would be good to have ...

PR #353, opened 2020-05-11
fpc lifecycle diagram

bvavala:
It contains the FPC lifecycle diagrams, including all of the operations performed by the parties (admin, peers, etc.) to package, install, approve and commit an FPC chaincode.


docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+    - the approveformyorg and commit ops are org-specific
+    So the approve stores data for an FPC chaincode installed locally,
+    and the commit checks for data of a locally-approved FPC chaincode.    
+end note
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : retrieve cc_language from package
+note over FPC_Peer1_Wrapper
+    Queryinstalled is redudant, yet consistent with implementation.    
+end note
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode queryinstalled
+return packageId
```
g2flyer, line 67: Let's drop this. No need to make diagrams more complicated than necessary?

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+
+group install
+Admin -> FPC_Peer1_Wrapper ++: lifecycle chaincode install
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode install
+return packageId
+note over FPC_Peer1_Wrapper
+    Bookeeping operations to take into account that:
+    - FPC's initial version will run a chaincode on a designated peer
+    - the install lifecyle op is peer-specific
+    - the approveformyorg and commit ops are org-specific
```
g2flyer, line 58: I think you are mixing up two things:
- we do bookkeeping as the cli doesn't have enough information from the `commit` parameters to figure out whether it is a FPC chaincode and which packages corresponds to it. That's why have to keep track of the mapping <code, package-id> in `install` and <package-id,id,version> in `approveformyorg`.

The other issue is that install is that not all action have to happen everywhere.  I think this is less an issue on bookkeeping but on requiring that approving and committing peers have the chaincode installed (which i'd argue is the normal expectation) and that we implicitly could define the committing peer as the designated-endorser (and hence piggy-back the corresponding enclave registration et al in commit)

g2flyer, line 58:
```suggestion
    Bookeeping operations to take into account that:
    - FPC's initial version will run a chaincode on a designated peer
    - the install lifecyle op is peer-specific
    - the `approveformyorg` ops is org-specific and executed at most on one peer per org
    - the `commit` op is executed only on a single peer
```

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+ref over Admin, FPC_Peer2_Wrapper, Peer2
+    install chaincode on Peer2
+end /' ref '/
+
+group /' approve '/
+Admin -> FPC_Peer1_Wrapper ++: lifecycle chaincode approveformyorg
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode approveformyorg
+note right
+    **Chaincode definition**
+    Name: cc_name, Version: mrenclave,
```
g2flyer, line 81: btw: currently the peer cli takes the version passed by the humans as-is and doesn't do anything with mrenclave.  Given that the package defines the MRenclave, i think we probably should just skip the version parameter from `approveformyorg` and automatically set it from the package or at least cross-check?  (if we keep track of sequence numbers in the bookkeeping, we could also drop it from `commit` ...)

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+    Now, check if this fpc cc was previously approved.
+    If so, perform some invocations.
+end note
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if cc_name, cc_version not stored then return
+FPC_Peer1_Wrapper -> Peer1 ++: invoke chaincode __setup
+Peer1 -> Orderer ++: invoke transaction
+return transaction committed
+return /' FPC_Peer1_Wrapper -> Peer1 '/
+FPC_Peer1_Wrapper -> Peer1 ++: query chaincode __getEnclavePk
+return /' FPC_Peer1_Wrapper -> Peer1 '/
```
mbrandenburger, line 135: This does not align well with the new enclave registration protocol I just reviewed.

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+return packageId
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if cc_language==fpc then store packageId
+return packageId /' Admin -> FPC_Peer1_Wrapper '/
+end /' group install '/
+
+ref over Admin, FPC_Peer2_Wrapper, Peer2
+    install chaincode on Peer2
+end /' ref '/
+
+group /' approve '/
```
mbrandenburger, line 76: the group name is not correctly shown ...

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if packageId was stored then store cc_namie, cc_version
+return /' Admin -> FPC_Peer1_Wrapper '/
+
+Channel -> Channel : approvals from other parties
+note over Channel
+    Other participants must approve the chaincode definition,
+    and commit the corresponding transaction,
+    according to the LifecycleEndorsement policy.
+    So the party here has to loop to check for the approvals.
+end note
```
mbrandenburger, line 101: I think this important step is not prominent enough .... I propose to remove the channel abstraction and add another ref box between approve and loop that refers to `approve` ... something similar like this
```
ref over Admin, FPC_Peer2_Wrapper, Peer2
    approvals form other orgs
end
```

and also including your note

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+group /' approve '/
+Admin -> FPC_Peer1_Wrapper ++: lifecycle chaincode approveformyorg
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode approveformyorg
+note right
+    **Chaincode definition**
+    Name: cc_name, Version: mrenclave,
+    Sequence: sequence_number, Endorsement Policy: (2of3),
+    Plugins: <custom endorsement/validation plugins>
+end note
+Peer1 -> Orderer ++: approve transaction
```
mbrandenburger, line 85: peer cannot invoke a transaction ... only client can

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+end
+
+group commit
+Admin -> FPC_Peer1_Wrapper ++:lifecycle chaincode commit
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode commit
+note right
+    **Chaincode definition**
+    (implemented to match the one that is ready to be committed)
+end note
+Peer1 -> Orderer ++: commit transaction
```
mbrandenburger, line 121: peer cannot invoke a transaction ... only client can

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+    Plugins: <custom endorsement/validation plugins>
+end note
+Peer1 -> Orderer ++: approve transaction
+return transaction committed /' Peer1 -> Orderer '/
+return /' FPC_Peer1_Wrapper -> Peer1 '/
+note over FPC_Peer1_Wrapper
+    Bookkeping operations.
+    Note: packageId is an input parameter of approveformyorg
+end note
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if packageId was stored then store cc_namie, cc_version
```
mbrandenburger, line 92:
```suggestion
FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if packageId was stored then store cc_name, cc_version
```

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+title Foot Box removed
+
+title Fabric Private Chaincode - Lifecycle v2
+
+box "Org1"
+actor	Admin order 10
+entity "FPC_Peer1_Wrapper" as FPC_Peer1_Wrapper order 15
+participant Peer1 order 20
+entity "FPC_Peer2_Wrapper" as FPC_Peer2_Wrapper order 30
+participant Peer2 order 40
```
mbrandenburger, line 22: our current implementation calls this a peer wrapper, however, when we talk about the new design should the e_manager be involved in the action that the wrapper performs here now?

mbrandenburger: Or should it be just a FPC_PEER? the wrapper is an actual implementation detail from my perspective

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
```diff
@@ -0,0 +1,139 @@
...
+    So the approve stores data for an FPC chaincode installed locally,
+    and the commit checks for data of a locally-approved FPC chaincode.    
+end note
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : retrieve cc_language from package
+note over FPC_Peer1_Wrapper
+    Queryinstalled is redudant, yet consistent with implementation.    
+end note
+FPC_Peer1_Wrapper -> Peer1 ++: lifecycle chaincode queryinstalled
+return packageId
+FPC_Peer1_Wrapper -> FPC_Peer1_Wrapper : if cc_language==fpc then store packageId
```
mbrandenburger, line 68: where is the packageId actually stored?

mbrandenburger: on the peer locally? just thinking about implementation .... if someone uses the peer cli to perform `peer chaincode install` remotely ... can this actually work?

mbrandenburger: why do we store packageId? I cant see in this diagram how we use it later.

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
g2flyer: the peer wrapper is really the peer _cli_ wrapper and from what we discussed on the other diagram maybe we should just call it `cli`?

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
g2flyer: @mbrandenburger : `packageid` has only peer-local semantics. The mapping package-id and the related packages is stored inside the peer. However, for our wrapper to work we also have to keep some mapping which we currently do via files on the filesystem (see also above comments of mine related bookkeeping.

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
+    install chaincode on Peer2
+end /' ref '/
+
+group local approve
+Admin -> FPC_Peer1_CLI_Wrapper ++: lifecycle chaincode approveformyorg
+FPC_Peer1_CLI_Wrapper -> Peer1_CLI ++: lifecycle chaincode approveformyorg
+note right
+    **Chaincode definition**
+    Name: cc_name, Version: mrenclave,
+    Sequence: sequence_number, Endorsement Policy: (2of3),
```
g2flyer, line 81: Iff we want protocols to reflect MVP, then endorsement policy cannot be 2of3 but would have to be Org1.peer? But hopefully this would be a policy which we could eventually support so i don't have a fundamental reason to keep it, but maybe worth adding a note that in MVP this would be on a designated peer endorsement policy

BTW: how the designated endorsement policy is represented in the fabric EP language is implicitly part of #358 and we had a corresponding issue in the gdoc mvp workplan, but not yet reflected in github issues. For now i just added a corresponding reference to #358 ...

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
+    Plugins: <custom endorsement/validation plugins>
```
g2flyer, line 82: Actually, we do not really support (user-specified) custom plugins. So maybe better to explicitly specify the ecc validation plugin (which will stay for MVP and presumably even for any later MSP/X509-style attestation handling)  and a note that no custom endorsement and validation plugins ?


Proposal for change:
```suggestion
    Plugins: <fpc ecc validation plugins>  
    [Note: FPC does not support user-defined endorsement or validation plugins]
```

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
g2flyer: Depends a bit on the workflow and ground-truth you expect the users to use.  I was assuming the code would be the reference for the user as easier but it is true that even though more cumbersome for the user (as well as for the scripting!) the ultimate ground truth is MRENCLAVE.

Given that wrong versions are quite likely (e.g., typo, copy/paste errros ...) the error happens only delayed (and probably also not very intuitively), cross-checking at approve time would seem a good idea.
  
BTW: I think it wouldn't be TLCC but rather ERCC would find the problem? (as discussed elsewhere ERCC would handle attestation and TLCC defers to it by just verifying proper ERCC transactions have happened. In that case I would expect TLCC only consult ercc state to see whether the enclave endorsement is done by a properly registered key?)

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
+FPC_Peer1_CLI_Wrapper -> FPC_Peer1_CLI_Wrapper : if cc_name, cc_version not stored then return
+FPC_Peer1_CLI_Wrapper -> Peer1_CLI ++: invoke chaincode __setup
+Peer1_CLI -> Orderer ++: invoke transaction
+return transaction committed
+return /' FPC_Peer1_CLI_Wrapper -> Peer1_CLI '/
+FPC_Peer1_CLI_Wrapper -> Peer1_CLI ++: query chaincode __getEnclavePk
+return /' FPC_Peer1_CLI_Wrapper -> Peer1_CLI '/
+return /' Admin -> FPC_Peer1_CLI_Wrapper '/
+end /' group commit '/
+
```
g2flyer, line 139: The `__setup` is not consistent with name in the registration protocol flow (this was also an observation by Marcus in the original review). It is also not clear whether we really want to hide all of that behind `commit`? It works for designated peer but definitely not beyond for 
 single peer or multi-peer endorsement? So for future-proofing seems better to have a separate admin command?.  This really also depends from what we decided to on first defining the peer cli management interface as part of #358 in a separate markdown rather than defining this interface in the UML as initially thought. (see, though, why below proposal elegantly side-steps that for this diagram :-) Won't work for the registration diagram though ...)  

Also `__getEnclavePK` is superfluous (even though i realize that it exists i think still in the actual cli wrapper).

Proposal would be to drop completely and just reference here that registration and key-managements follow with just a reference to these diagrams. Below a suggestion (but note i haven't tested whether it looks nice in PUML!)

```suggestion
end /' group commit '/
note left
  To complete setting up an FPC chaincode, enclaves have to be registered and provisioned with keys.  See the separate UML diagrams `fpc-registration` and `fpc-key-dist` for the corresponding admin commands and protocol flows.
end note
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
+ref over Admin, FPC_Peer2_CLI_Wrapper, Peer2_CLI
+    install chaincode on Peer2
+end /' ref '/
+
+group local approve
+Admin -> FPC_Peer1_CLI_Wrapper ++: lifecycle chaincode approveformyorg
+FPC_Peer1_CLI_Wrapper -> Peer1_CLI ++: lifecycle chaincode approveformyorg
+note right
+    **Chaincode definition**
```
g2flyer, line 79: We should probably make explicit this applies only iff the chaincode is an fpc chaincode. 
```suggestion
    **Chaincode definition (in case chaincode is fpc, ie.e., we stored packageid in an earlier install)**
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
 end note
-
-Admin_Org2 -> Orderer : Approve
-note left
-	**Chaincode definition**
-	Name: mycc, Version: 1
-	End Policy: (2of3), Seq 1,
-	Package ID: mycc_1:hash, <-???
-	//<color #00FF00>ECC_MR: mrenc_1</color>//
+FPC_Peer1_CLI_Wrapper -> FPC_Peer1_CLI_Wrapper : if packageId was stored then store cc_name, cc_version
```
g2flyer, line 91: actually, we have to store a mapping package-id to <cc_name, cc_version>
```suggestion
FPC_Peer1_CLI_Wrapper -> FPC_Peer1_CLI_Wrapper : if packageId was stored (and hence it is an FPC chaincode) then store mapping packageid to <cc_name, cc_version>
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,126 @@ title Foot Box removed
...
+    The following bookking operations help remember (across lifecycle calls)
+    that a package (id) corresponds to a FPC chaincode. Similarly, the
+    approve will later store data for an FPC chaincode installed locally,
+    and the commit checks for data of a locally-approved FPC chaincode.
+    Also, it should be noted that
+    - the install lifecycle op is peer-specific, while
+    - the approveformyorg is org-specific, to be run on at most one org's peer
+    - the and commit op is expected to run on a single peer
+    Hence, the current bookeeping ops expect the approve and commit ops to
+    run on peers where the FPC chaincode has been previously installed.
```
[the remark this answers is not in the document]
bvavala: 650f888635af9a01b34085d1b2b587308b866faf

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
bvavala: f1fb20ff656811c72403b872681a17b94303cd39

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
bvavala: 03d533192d7c9c2bfcc6c523700d12d9fead3cda

docs/design/fabric-v2+/fpc-lifecycle-v2-new.puml
bvavala: 07b4ce2a989f777ea4f0b16722ca2a6274ce6a15

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
bvavala: 07b4ce2a989f777ea4f0b16722ca2a6274ce6a15

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
bvavala: 1719c419520883924a432a9281ec80736240a000

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
bvavala: I have two concerns here going forward:
1) we have a diagram that does not match the code
2) most importantly, given the confusion, we should consider
- dropping the wrapper (for the short-term)
- implementing the commands along side Fabric's ones (just as Jason suggested)
- once FPC is mature, then consolidate FPC and Fabric commands (possibly in a wrapper)

g2flyer: Regarding 1., well, we haven't changed anything in the code yet, so how can the new design match the code?  Inevitably, the design is always ahead of the code?!  Regarding the wrapper, i don't see how that is contrary to what Jason said: we are exposing additional user actions as Jason suggested but we do it in a way which is future proof.  In any case, i thought we decided already in the registration diagram discussion to handle user (admin) actions in the cli component (which essentially is the cli+wrapper)?

mbrandenburger: Regarding 2), in the lifecycle diagram I would drop the differentiation between cli and cli wrapper and just call it `peer cli` as used by the admin to talk to a peer. As mentioned earlier the wrapper is just our way to add additional functionality to the fabric peer cli, an alternative would be to indeed add our functionality to the cli code and recompile. 

As this diagram should illustrate the conceptional changes we need for FPC to work. 
Let me ask the question what are the major changes here? 

For lifecycle ...

- do we add any additional management commands needed for FPC? - No
- what are the crucial steps?

1. package and install: FPC chaincode comes as external builder package

1. approveformyorg : Here FPC changes are important as we need to set MRENCLAVE and ecc-vscc, this is the way all orgs agree on MRENCLAVE

Regarding the details on storing packageID inside the CLI (wrapper), I understand this is a nice helper to abstract away FPC complexity but I see this as an implementation specific decision to let our CLI implementation autofill the chaincode version with MRENCLAVE and validation plugin with ecc-vscc. I am questioning if this is relevant for the purpose of this document.

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,125 @@ title Foot Box removed
...
-participant Peer1_Org1
-participant Peer2_Org1
-end box
-
-box "Org2"
-actor	Admin_Org2
-participant Peer1_Org2
-participant Peer2_Org2
+actor	Admin order 10
+entity "FPC_Peer1_CLI_Wrapper" as FPC_Peer1_CLI_Wrapper order 15
```
mbrandenburger, line 19:
```suggestion
entity "FPC_Peer1_CLI_Wrapper" as FPC_Peer_CLI order 15
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,125 @@ title Foot Box removed
...
+participant Peer1_CLI order 20
+entity "FPC_Peer2_CLI_Wrapper" as FPC_Peer2_CLI_Wrapper order 30
```
mbrandenburger, line 21:
```suggestion
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,125 @@ title Foot Box removed
...
-participant Peer2_Org1
-end box
-
-box "Org2"
-actor	Admin_Org2
-participant Peer1_Org2
-participant Peer2_Org2
+actor	Admin order 10
+entity "FPC_Peer1_CLI_Wrapper" as FPC_Peer1_CLI_Wrapper order 15
+participant Peer1_CLI order 20
```
mbrandenburger, line 20:
```suggestion
participant Peer1 order 20
```

I think this should be just a peer

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -18,119 +15,125 @@ title Foot Box removed
...
+entity "FPC_Peer2_CLI_Wrapper" as FPC_Peer2_CLI_Wrapper order 30
+participant Peer2_CLI order 40
```
mbrandenburger, line 22:
```suggestion
participant Peer2 order 40
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
g2flyer: @mbrandenburger  Ah, now, i see that you actually did add an explanation to your change below :-) [and i think it is a good suggestion: the cli component is essentially our management API, many of the commands as-is from fabric, some with slightly refined semantics (approveformyorg/commit require version to be mrenclave) and some new (enclave creation & registration etc)

PR #359, opened 2020-05-14
Replace explicit list of maintainers but group (team)

g2flyer:
As discussed today, the PR which changes from explict CODEOWNERS to the github group, presumably then also replacing the group as choosen reviewers for PRs ....


MHBauer:
teams, nice 👌

PR #362, opened 2020-05-19
Make peer & dev builds leaner by excluding unnecessary PDO dependencies

g2flyer:

Subject says it all (mostly: the excluded dependencies are micro-wasm and, in particular, CCF ....)

```

### DIFF
--- a/utils/docker/dev/Dockerfile
+++ b/utils/docker/dev/Dockerfile
@@ -86,9 +86,13 @@ RUN git clone --branch v$FABRIC_VERSION $FABRIC_REPO $FABRIC_PATH
 ENV FPC_PATH=$GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 # We copy context so we can use that to potentially get local .git as repo ...
 COPY .git /tmp/cloned-local-fpc-git-repo
-RUN git clone --recurse $FPC_REPO_URL $FPC_PATH \
+RUN git \
+       -c submodule.interpreters/wasm-micro-runtime.update=none -c submodule.ccf_transaction_processor/CCF.update=none \
+	clone --recurse $FPC_REPO_URL $FPC_PATH \
   && cd $FPC_PATH \
   && git checkout --recurse $FPC_REPO_BRANCH_TAG_OR_COMMIT
+# Notes:
+# - the -c submodule's are to prevent dragging in large but unneeded sub-sub-modules of pdo ...
 
 # Note: below are differences to peer image. Eventually this should be consolidated via some refactoring
 # but for now we just make sure we have a common prefix so docker caching will reuse above for the second
--- a/utils/docker/peer/Dockerfile
+++ b/utils/docker/peer/Dockerfile
@@ -86,10 +86,14 @@ RUN git clone --branch v$FABRIC_VERSION $FABRIC_REPO $FABRIC_PATH
 ENV FPC_PATH=$GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 # We copy context so we can use that to potentially get local .git as repo ...
 COPY .git /tmp/cloned-local-fpc-git-repo
-RUN git clone --recurse $FPC_REPO_URL $FPC_PATH \
+RUN git \
+       -c submodule.interpreters/wasm-micro-runtime.update=none -c submodule.ccf_transaction_processor/CCF.update=none \
+	clone --recurse $FPC_REPO_URL $FPC_PATH \
   && cd $FPC_PATH \
   && git checkout --recurse $FPC_REPO_BRANCH_TAG_OR_COMMIT
-# Note: could add --single-branch to below to speed-up and keep size smaller. But for now for a dev-image better keep complete repo
+# Notes:
+# - the -c submodule's are to prevent dragging in large but unneeded sub-sub-modules of pdo ...
+# - could add --single-branch to below to speed-up and keep size smaller. But for now for a dev-image better keep complete repo
 
 # Build FPC peer including plugins and tlcc support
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -313,6 +313,10 @@ Clone the code and make sure it is on your `$GOPATH`. (Important: we assume in t
 
     $ git clone --recursive https://github.com/hyperledger-labs/fabric-private-chaincode.git $GOPATH/src/github.com/hyperledger-labs/fabric-private-chaincode
 
+Note: above will pull in some large but not needed dependencies from
+PDO. You can omit the corresponding pull by prefixing `clone` with
+`-c submodule.interpreters/wasm-micro-runtime.update=none -c submodule.ccf_transaction_processor/CCF.update=none `.
+
```
MHBauer, line 319: put this before the clone, so people read it before they do it.

utils/docker/dev/Dockerfile
```diff
@@ -86,9 +86,13 @@ RUN git clone --branch v$FABRIC_VERSION $FABRIC_REPO $FABRIC_PATH
...
 # We copy context so we can use that to potentially get local .git as repo ...
 COPY .git /tmp/cloned-local-fpc-git-repo
-RUN git clone --recurse $FPC_REPO_URL $FPC_PATH \
+RUN git \
+       -c submodule.interpreters/wasm-micro-runtime.update=none -c submodule.ccf_transaction_processor/CCF.update=none \
+	clone --recurse $FPC_REPO_URL $FPC_PATH \
   && cd $FPC_PATH \
   && git checkout --recurse $FPC_REPO_BRANCH_TAG_OR_COMMIT
+# Notes:
+# - the -c submodule's are to prevent dragging in large but unneeded sub-sub-modules of pdo ...
```
MHBauer, line 95: same, comment before doing things.

utils/docker/peer/Dockerfile
```diff
@@ -86,10 +86,14 @@ RUN git clone --branch v$FABRIC_VERSION $FABRIC_REPO $FABRIC_PATH
...
-RUN git clone --recurse $FPC_REPO_URL $FPC_PATH \
+RUN git \
+       -c submodule.interpreters/wasm-micro-runtime.update=none -c submodule.ccf_transaction_processor/CCF.update=none \
+	clone --recurse $FPC_REPO_URL $FPC_PATH \
   && cd $FPC_PATH \
   && git checkout --recurse $FPC_REPO_BRANCH_TAG_OR_COMMIT
-# Note: could add --single-branch to below to speed-up and keep size smaller. But for now for a dev-image better keep complete repo
+# Notes:
+# - the -c submodule's are to prevent dragging in large but unneeded sub-sub-modules of pdo ...
+# - could add --single-branch to below to speed-up and keep size smaller. But for now for a dev-image better keep complete repo
```
MHBauer, line 96: same

utils/docker/dev/Dockerfile
g2flyer: Right now in dockerfiles (and also in scripts afaik) such "Notes:" comments are always after the command.  So for consistency reason i would leave it (unless we think it's worth the effort to consistently change it to opposite order)?

> same, comment before doing things.

Right now in dockerfiles (and also in scripts afaik) such "Notes:" comments are always after the command.  So for consistency reason i would leave it (unless we think it's worth the effort to consistently change it to opposite order)?

utils/docker/peer/Dockerfile
g2flyer: > same

See above

README.md
g2flyer: > put this before the clone, so people read it before they do it.

Personally I think adding that before is a bit confusing.  If you think most people really would like to use the these options, then we might just add the options to the command right away? (I just didn't do it as i thought not everybody cared and the command would get a bit long/clumsy if all spelled out.  But if the majority of folks would prefer the leaner variant, then it's probably best to just add it to the command right away (and just drop the note altogether?)

mbrandenburger: ufff .... I would not expose this to users here as it makes it too complicated. Isn't there a way to clone with shallow-submodules? 
https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---no-shallow-submodules

Our goal should be to make the fpc `git clone` command as natural/easy as possible. 

If we exclude these submodules in the docker images, that is fine, as long as we dont ask the user to exclude them as well!

g2flyer:
> Thanks Michael! I think it's a good idea to strip out unneeded (sub)submodules from PDO. What I don't like so much is that we expose this to the user in the readme as most people will don't understand it and eventually just copy and paste the `git clone` without excluding them. I think for the users, the git clone is an one time action whereas in our docker setup, which we will also use for CI, this checkout is performed more often, so it makes sense to make it more lightweight.

Keeping the command simple and just give the additional space-saving info for whoever cares afterwards was my original approach to deal with that.  But as also Morgan didn't really like it, i'll just remove the note paragraph from the README.  I guess the smart user who cares might figure it out also from the Dockerfile which he might anyway consult to figure out how to install dependencies :-)

> Maybe there is an easy way to just prevent all subsubmodules from checkout?

I couldn't find anything clean & easy. This was the best i found.  The shallow-submodule will still check out all submodules but the amount checked out should be indeed leaner.  From a peer image perspective you could even do a shallow check out of fpc proper (see comment in dockerfile) but that doesn't seem what you want from the dev container and for symmetry perspective i kept them the same ...

> Within docker I dont see an issue excluding the subsubmodulse as you propose.

mbrandenburger:
@g2flyer I missed the fixup and just wanted to update the branch and merge it. sorry. Can you please do the squash/rebase and I will merge.

PR #367, opened 2020-06-08
port pdo crypto tests

bvavala:
(subject is self-explanatory)

### DIFF
--- a/common/crypto/CMakeLists.txt
+++ b/common/crypto/CMakeLists.txt
@@ -10,35 +10,58 @@ SET(U_CRYPTO_ADAPT_LIB_NAME updo-crypto-adapt)
 SET(T_CRYPTO_ADAPT_LIB_NAME tpdo-crypto-adapt)
 
 SET(PDO_CRYPTO_DIR "pdo/common/crypto")
+SET(PDO_CRYPTO_TEST_DIR "${PDO_CRYPTO_DIR}/../tests/crypto")
 
 FILE(GLOB PROJECT_HEADERS "${PDO_CRYPTO_DIR}/../*.h")
-FILE(GLOB PROJECT_SOURCES 
+FILE(GLOB PROJECT_SOURCES
     "${PDO_CRYPTO_DIR}/../c11_support.cpp"
     "${PDO_CRYPTO_DIR}/../hex_string.cpp"
     "${PDO_CRYPTO_DIR}/../log.cpp"
     "pdo-types.cpp"
+    "$ENV{FPC_PATH}/common/base64/base64.cpp"
     )
 
 ###################################################################################################
 # Untrusted crypto adapt library
 ###################################################################################################
-#PROJECT(${U_CRYPTO_ADAPT_LIB_NAME} CXX)
-#ADD_LIBRARY(${U_CRYPTO_ADAPT_LIB_NAME} STATIC ${PROJECT_HEADERS} ${PROJECT_SOURCES})
-#TARGET_INCLUDE_DIRECTORIES(${U_CRYPTO_ADAPT_LIB_NAME} PRIVATE "${PDO_CRYPTO_DIR}/..")
-#TARGET_INCLUDE_DIRECTORIES(${U_CRYPTO_ADAPT_LIB_NAME} PRIVATE "$ENV{FPC_PATH}/common/base64")
-#TARGET_INCLUDE_DIRECTORIES(${U_CRYPTO_ADAPT_LIB_NAME} PUBLIC $ENV{SGX_SDK}/include)
+PROJECT(${U_CRYPTO_ADAPT_LIB_NAME} CXX)
+ADD_LIBRARY(${U_CRYPTO_ADAPT_LIB_NAME} STATIC ${PROJECT_HEADERS} ${PROJECT_SOURCES})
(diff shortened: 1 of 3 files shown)


common/crypto/Makefile
```diff
@@ -7,15 +7,22 @@ include $(TOP)/build.mk
...
+
 $(BUILD_DIR):
 	@if [ ! -d $(BUILD_DIR) ]; then \
 		mkdir -p $(BUILD_DIR) && \
-		cd $(BUILD_DIR) && \
-		cmake ./..; \
-	fi
+        cd $(BUILD_DIR) && \
+		openssl genrsa -3 -out $(PDO_ENCLAVE_CODE_SIGN_PEM) 3072 && \
+        export PDO_ENCLAVE_CODE_SIGN_PEM=`readlink -f $(PDO_ENCLAVE_CODE_SIGN_PEM)` && \
```
g2flyer, line 17: We already have an enclave signer for FPC, can't we just use that and point PDO_ENCLAVCE_CODE_SIGN_PEM and ultimately we might want only a single signer key which could be configured? (admittedly, not really important for test programs but still a bit confusing)

PS: Looking at FPC code i realize that we do that kindof ugly in ../../ecc_enclave/enclave/CMakeLists-common-app-enclave.txt and ../../tlcc_enclave/enclave/CMakeLists.txt where we unconditionally create separate new keys.  So this seems more like a separate issue to be opened ...

bvavala: PDO_ENCLAVCE_CODE_SIGN_PEM is required in PDO.
Anyway:
1) we don't need configurability for the signer key
2) we should not have dangling keys in the repo.
It is for (1) (2) that "I" removed those keys in FPC and coded their creation just and only when we need them at runtime. 
For the same reasons, I was actually thinking of removing the key in PDO too -- additionally getting rid of unnecessary complexity.

g2flyer: I've created now issue #368 for that ...

g2flyer: I realize that we need a defined PDO_ENCLAVE_CODE_SIGN_PEM variable pointing to a key.  If we want a secure version of FPC, we _do need_ though a configurable version of that key (see #368).  But certainly hasn't to be done now and even less so in this PR.  (I just first we already had a centralized version in which case we should have reused the same one instead of creating a new one ...)

bvavala:
> I've tested by manually setting the sub-project to PDO PR 285 and the pdo crypto tests build and run nicely. Overall build fails, though (see below PS) and one suggestion on how we probably can prevent further duplication. But getting the standard (and via PR 285 also improved) pdo crypto tests invoked in our build is definitely nice!
> 
> PS: license check fails, seems we haven't excluded pdo from all our linter & license check scripts?
> 
> ```
> The following files are missing SPDX-License-Identifier headers:
> common/crypto/pdo
> 
> Please replace the Apache license header comment text with:
> SPDX-License-Identifier: Apache-2.0
> 
> Checking committed files for traditional Apache License headers ...
> grep: common/crypto/pdo: Is a directory
> The following files are missing traditional Apache 2.0 headers:
> common/crypto/pdo
> Fatal Error - All files must have a license header
> Makefile:24: recipe for target 'license' failed
> make: *** [license] Error 1
> ```

Weird, it does not fail on master. I will fix this later if necessary.

common/crypto/CMakeLists.txt
```diff
@@ -10,35 +10,58 @@ SET(U_CRYPTO_ADAPT_LIB_NAME updo-crypto-adapt)
...
+# lib names needed at link time
+SET(TRUSTED_LIB_NAME ${T_CRYPTO_ADAPT_LIB_NAME})
+SET(UNTRUSTED_LIB_NAME ${U_CRYPTO_ADAPT_LIB_NAME})
+SET(U_CRYPTO_LIB_NAME updo-crypto)
+SET(T_CRYPTO_LIB_NAME tpdo-crypto)
+# pkg_check_modules needed to set OPENSSL_LDFLAGS
+find_package(PkgConfig REQUIRED)
+pkg_check_modules (OPENSSL REQUIRED openssl>=1.1.0g)
+# add pdo crypto tests
+ADD_SUBDIRECTORY (${PDO_CRYPTO_TEST_DIR})
```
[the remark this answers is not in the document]
bvavala: Not sure what you mean.
In `common/crypto/CMakeLists.txt`, the `adapt` lib mirrors `pdo_common`, which contains everything else (besides crypto) to compile tests. The "PDO Crypto tests" block mainly (lines 54-65) sets the names for `adapt` and `crypto` libraries which are expected to compile the tests (line 67).
Instead, `pdo/tests/crypto/CMakeLists.txt` only says how to compile the trusted and the untrusted test executables.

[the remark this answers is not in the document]
g2flyer: sorry, got confused (or more precisely my visual pattern matching was too buggy ..) and i've overlooked the last line with the ADD_SUBDIRECTORY ...

mbrandenburger:
@g2flyer @bvavala We we merge this one?

bvavala:
> @g2flyer @bvavala We we merge this one?

yes, after we close the others. This has low priority, and it's not worth addressing the conflicts now.

Issue #134, opened 2019-08-07
New ERCC model


mbrandenburger:
The current ERCC model comes with some drawbacks:
  - Single-endorser policy only + Custom validation at __every__ peer
  - ERCC must contact IAS (Internet connection and IAS credentials needed)
  - Current impl: No way to securely pass credentials to ERCC only
  - for EPID attestation SPID is needed when quote is created (ECC must provide this)

The purpose if this task is to design a new ERCC model that overcomes these limitations.

Last, implement this.

mbrandenburger:
#116 proposes a new design.

mbrandenburger:
#116 merges an intermediated version into _new_design_ branch.

mbrandenburger:
@bvavala Do you want to take over this issue?

mbrandenburger:
PR #371 had a good discussion regarding ERCC validation. This may help when implementing new ERCC scheme and moving away from ERCC custom validation.

> Why is this actually not the only/main case?  Do we currently allow updates? Otherwise, there shouldn't be anything in the state anyway? If so, how are we doing updates securely?  I realize this plugin will go away but the validation logic should transfered to some extent to ercc itself, so this code fragment is still somewhat relevant? maybe some comments could be helpful for the future?

_Originally posted by @g2flyer in https://github.com/hyperledger-labs/fabric-private-chaincode/pull/371_

mbrandenburger:
> I guess the current code would allow wrap an ecc _setup-tx by (via cc2cc) by another tx and "hijack" the attestation and register it with a different ecc_namespace?  Per-se i don't see immediately how this leads to bad stuff but it's a bit ugly.  Ideally, i think the attestation should explicitly bind the the ecc namespace?  (Not necessarily here, but might be good to do so in the revised version? Actually, iff we implement the new registration protocol, we should do so ... :-)

_Originally posted by @g2flyer in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0NDMyNDQ1MjA1/pull_requests/unread_timeline_

mbrandenburger:
The new ERCC model is defined in https://github.com/hyperledger-labs/fabric-private-chaincode/blob/cr2.0.0/docs/design/fabric-v2%2B/fpc-registration.puml

mbrandenburger:
@bvavala Can we close this one as #404 already focuses on the implementation of this new design?

bvavala:
[...]

PR #371, opened 2020-06-10
Re-enable ecc-vscc and ercc-vscc

### LINKED ISSUE #250: Make custom-vscc for ECC and ERCC Fabric 2.0 compatible

mbrandenburger:
This PR fixes #250 and re-enables ecc and ercc custom validation. 

This PR fixes #250 and re-enables ecc and ercc custom validation. 

- Change core.yaml to point to plugins/*-vscc.so
- Set ercc-vscc in ercc chaincode definition in our peer wrapper
- Set ecc-vscc for fpc chaincodes in integration tests
- add workaround in ercc-vscc to fix missing "ecc_" namespace prefix
  for fpc chaincodes; and fix mrenclave retrieval
- remove utils depency from ecc-vscc as it caused errors when peer
  loads the plugin
- improved debug logging in ecc-vscc
- add workaround in ecc-vscc for fix missing "ecc_" namespace prefix

Fixes #250

```

### CODE REVIEW

File: plugins/ecc-vscc/ecc_validation_logic.go [hunk 1]
```diff
@@ -233,12 +246,25 @@ func (vscc *VSCCECC) checkEnclaveEndorsement(cis *peer.ChaincodeInvocationSpec,
...
 }
 
+func isFPCNameSpace(nameSpace string) bool {
+
+	if nameSpace == "_lifecycle" {
+		// do nothing for lifecycle namespace
+		return false
+	}
+
+	// TODO do some real check if namespace is a FPC namespce.
```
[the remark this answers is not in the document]
g2flyer: Can't we solve it right now?  Couldn't we use presence of the "ecc-vscc" validation plugins?  Otherwise, let's revert to the ecc_ (or better, fpc_) prefixing of chaincode name we did earlier? (it's really ugly, but short of changing chaincode definition which we are unlikely going to do for MVP, i guess there are only more or less ugly solutions?  (Personally, the validation-plugin route seems nicer than the prefix one ...)

BTW: this issue exists also in TLCC. So if we do not address it here then we definitely need an issue and work-plan update for it!

integration/deployment_test.sh
```diff
@@ -26,9 +26,9 @@ run_test() {
     try ${PEER_CMD} lifecycle chaincode package --lang fpc-c --label auction_test --path ${FPC_TOP_DIR}/examples/auction/_build/lib/ ${PKG}
     try ${PEER_CMD} lifecycle chaincode install ${PKG}
     PKG_ID=$(${PEER_CMD} lifecycle chaincode queryinstalled | awk "/Package ID: auction_test/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
-    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name auction_test --version ${CC_VERS}
-    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name auction_test --version ${CC_VERS}
-    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name auction_test --version ${CC_VERS}
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name auction_test --version ${CC_VERS} -V ecc-vscc
```
g2flyer, line 29: Wouldn't it be better to add the validation plugin inside the peer cli?  Mid/long-term some of these plugins will change for FPC, so it would be cleaner/easier from a functional interface point of view as the interface is simpler and would be the same even if we would move to MSP/X509 style validation.  As we anyway do not allow any user-provided plugins, the peer cli of course should also barf if any plugins provided (which it currently doesn't, e.g., if you provide an endorsement plugin, which by itself is also a bit of a problem?)

mbrandenburger: > Wouldn't it be better to add the validation plugin inside the peer cli? Mid/long-term some of these plugins will change for FPC, so it would be cleaner/easier from a functional interface point of view as the interface is simpler and would be the same even if we would move to MSP/X509 style validation. As we anyway do not allow any user-provided plugins, the peer cli of course should also barf if any plugins provided (which it currently doesn't, e.g., if you provide an endorsement plugin, which by itself is also a bit of a problem?)

Created a new issue for that #372

plugins/ecc-vscc/ecc_validation_logic.go
```diff
@@ -27,6 +26,19 @@ import (
...
+// Response contains the response data and signature produced by the enclave
+type Response struct {
+	ResponseData []byte `json:"ResponseData"`
+	Signature    []byte `json:"Signature"`
+	PublicKey    []byte `json:"PublicKey"`
+}
+
+const SEP = "."
+
+func TransformToSGX(comp, sep string) string {
```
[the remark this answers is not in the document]
mbrandenburger: naaaah not really ... at least we are always using it with this const SEP.

I just copied this function from [utils/utils.go](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/master/utils/utils.go#L49) to break the dependency issue.

plugins/ercc-vscc/ercc_validation_logic.go
```diff
@@ -232,13 +236,15 @@ func policyErr(err error) *commonerrors.VSCCEndorsementPolicyError {
 }
 
 func getEccNamespace(txRWSet *rwsetutil.TxRwSet) (string, error) {
-	// Readsets must contain only one ecc_*.MRENCLAVE
+	// Readsets must contain only one *some_fpc_chaincode_name*.MRENCLAVE
 	eccNamespace := ""
 	for _, ns := range txRWSet.NsRwSets {
 		for _, read := range ns.KvRwSet.Reads {
-			if read.Key == MrEnclaveStateKey && strings.HasPrefix(ns.NameSpace, "ecc_") {
+			// Note that we used to use ecc_ as prefix for every fpc chaincode; however, this has been removed
```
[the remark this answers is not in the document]
mbrandenburger: Ok good idea! Let's remove comment here and add to commit msg.

mbrandenburger:
> Looks nice to me. Does CI Run this check?

Yes! Our custom vsccs are now used in integration tests.

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
g2flyer: Argh, we really have to duplicate the function and constants?! This is really ugly.  Is this really necessary? At the very least we should add comments at both places cross-referencing each other as presumably they should be kept consistent.

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
g2flyer: From discussion on today's call i guess conclusion is that (a) presence of escc-validation is proper way of identification of FPC chaincode, (b) until ERCC is refactored, we need though a workaround for the current erccc cc2cc (presumably based on the "ercc" namespace?) and (c) we need a new (MVP) issue to implement that strategy in TLCC and remove the workaround as soon as ercc is refactored?

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
g2flyer: btw: i agree with morgan. Given the name of the function, the SEP as argument is rather confusing and would be better handled inside the function ....

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
mbrandenburger: let me look into that if we can solve it right now or not. Since we are not supporting cc2cc calls with FPC there is a chance that non-fpc chaincodes wont show up in the read/writeset. Only _lifecycle and maybe ercc? but we can exclude them. Will look into that.

plugins/ercc-vscc/ercc_validation_logic.go
```diff
@@ -187,10 +186,15 @@ func (t *VSCCERCC) checkAttestation(respPayload *peer.ChaincodeAction) error {
...
 		}
+		// if mrenclave not already on the ledger
 		if mrenclave == nil {
-			return errors.New("mrenclave is empty")
+			// it must be in the writeset
+			mrenclave = getMrEnclaveWriteSet(eccNamespace, txRWSet)
+			if mrenclave == nil {
+				return fmt.Errorf("no MRENCLAVE on the ledger nor in write set for %s", eccNamespace)
+			}
 		}
```
g2flyer, line 198: Why is this actually not the only/main case?  Do we currently allow updates? Otherwise, there shouldn't be anything in the state anyway? If so, how are we doing updates securely?  I realize this plugin will go away but the validation logic should transfered to some extent to ercc itself, so this code fragment is still somewhat relevant? maybe some comments could be helpful for the future?

mbrandenburger: Well, the code also considers the case where setup is already called for another enclave , in that case mrenclave may already exists on the ledger. However, I guess this peace of logic needs bigger rework as we will consume MRENCLAVE from chaincode definition and not from ledger state. Thus, I don't think we need to touch it here or add any comment.

mbrandenburger: I added a pointer to this comment in #134

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
mbrandenburger: ok, I refactored utils/utils.go and removed the go-shim dependency, so we can use utils.go again within ecc-vscc and ecc_enclave. This solves code duplication concerns here. 

Also removed `sep` input from util functions, even though this should be a seperate work item and does not help to solve the actual task of this PR. But if we are already touching it here - it's OK!

plugins/ecc-vscc/ecc_validation_logic.go
```diff
@@ -121,70 +120,70 @@ func (vscc *VSCCECC) checkEnclaveEndorsement(cis *peer.ChaincodeInvocationSpec,
...
+	logger.Debugf("function: %s\n", string(function))
 
-		function := cis.ChaincodeSpec.Input.Args[0]
-		logger.Debugf("function: %s\n", string(function))
-
-		if string(function) == "__setup" {
-			continue
-		}
+	if string(function) == "__setup" {
+		// do nothing for setup transaction as we handle this with ercc-vscc
```
g2flyer, line 127: Just for clarifications -- overall, a few more comments for the whole plugin validations couldn't have hurt ... -- fabric calls validation plugins always on the (complete) transactions as as soon as there is a top-level or sub (cc2cc) transacation for a chaincode associated with a custom validation plugin?  I.e., for setup transaction both validation plugins are called -- and hence above works ... -- , any other (top-level) ercc transaction involves only ercc-vscc and any other (top-level) ecc tx involves only ecc-vscc? (regarding any other top-level ercc tx, i guess all would be rejected by the plugin with non-intuitve errors but would be safe, in any case, this plugin would be obsolete soon ...)

plugins/ecc-vscc/ecc_validation_logic.go
```diff
@@ -121,70 +120,70 @@ func (vscc *VSCCECC) checkEnclaveEndorsement(cis *peer.ChaincodeInvocationSpec,
 		return err
 	}
 
-	for _, ns := range txRWSet.NsRwSets {
-		logger.Debugf("Namespace %s", ns.NameSpace)
-
-		// TODO make this more flexible
-		if !strings.HasPrefix(ns.NameSpace, "ecc") {
-			continue
-		}
```
g2flyer, line 130: I guess this we dropped as, as discussed, by definition the ecc plugin should be configured exactly (and only) for ecc transactions?

plugins/ecc-vscc/ecc_validation_logic.go
```diff
@@ -121,70 +120,70 @@ func (vscc *VSCCECC) checkEnclaveEndorsement(cis *peer.ChaincodeInvocationSpec,
...
-		// Next, reproduce sorted read/writeset
-		var readset, writeset [][]byte
+	var readset, writeset [][]byte
+	// go through the read/writeset as produced within the enclave
+	for _, ns := range txRWSet.NsRwSets {
+		if ns.NameSpace != cis.ChaincodeSpec.ChaincodeId.Name {
+			// note that the read/write set may contain other namespaces than our fpc chaincode such as lifecycle
+			// we only care about FPC chaincode namespace and therefore filter
+			continue
+		}
```
g2flyer, line 181: I would have expected that ercc would be the only other namespace?  Is a custom validation plugin also called on lifecycle events?  That would be kindof surprising (e.g., for chaincode spec updates updating plugins, which plugins are called?!) but more importantly, how would we make sure the corresponding transactions are properly verified!  By doing continue what would validate the corresponding path?   In fact, i doubt that lifecycle tx would come here as they usually would not have ecc attestations and hence below verification fails?

So ideally we would test that it is ercc or fail otherwise?  In any case, i think the current comment is wrong and should be updated

plugins/ercc-vscc/ercc_validation_logic.go
```diff
@@ -232,13 +236,14 @@ func policyErr(err error) *commonerrors.VSCCEndorsementPolicyError {
...
 	for _, ns := range txRWSet.NsRwSets {
 		for _, read := range ns.KvRwSet.Reads {
-			if read.Key == MrEnclaveStateKey && strings.HasPrefix(ns.NameSpace, "ecc_") {
+			if read.Key == MrEnclaveStateKey {
+				logger.Debugf("found MRENCLAVE within namespace: %s", ns.NameSpace)
 				if eccNamespace != "" {
-					return "", fmt.Errorf("mutiple namespaces for ecc found")
+					return "", fmt.Errorf("mutiple namespaces with MRENCLAVE key found")
 				}
 				eccNamespace = ns.NameSpace
```
g2flyer, line 248: I guess the current code would allow wrap an ecc _setup-tx by (via cc2cc) by another tx and "hijack" the attestation and register it with a different ecc_namespace?  Per-se i don't see immediately how this leads to bad stuff but it's a bit ugly.  Ideally, i think the attestation should explicitly bind the the ecc namespace?  (Not necessarily here, but might be good to do so in the revised version? Actually, iff we implement the new registration protocol, we should do so ... :-)

plugins/ecc-vscc/ecc_validation_logic.go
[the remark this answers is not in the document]
g2flyer: You mean "non-fpc chaincodes" _except_ ercc?!  In any case, from your latest version i think you consider it resolved for the ecc (and ercc) validation plugins?    I would tend to agree (although see some comments in my review). However, what about correspondingly also fix tlcc: do we have a corresponding issue for that?

plugins/ecc-vscc/ecc_validation_logic.go
mbrandenburger: here we iterate over the read/writeset of a transaction associated with ecc-vscc. As an ecc-enclave endorses only read/writes that are part of the fpc chaincode namespace, any other entry showing up here are not signed by the enclave, thus, must be filtered out in order to verify the enclave signature.

What are those other entries? If you look at the logs we can see that every chaincode transaction contains a readset entry to `_lifecycle`.`$chaincode_name$`. In particular, TLCC logs show it pretty well and I assume that explorer would show it as well.
It seems that this is a check that all endorsements use the same "chaincode version". Anyway, as this appears in the readset, the entry is covered by the peer endorsement which we have on top of ecc-vscc right now. So no need for additional checks here.

plugins/ecc-vscc/ecc_validation_logic.go
mbrandenburger: yes

plugins/ecc-vscc/ecc_validation_logic.go
mbrandenburger: I believe the `__setup` transaction will go away with ercc refactoring and the corresponding comments on how fabric deals with cc2cc invocations are then obsolete here.

I updated the comment but I refrain from adding details on how cc2cc are handled in fabric.

plugins/ercc-vscc/ercc_validation_logic.go
mbrandenburger: I linked this conversation to #134

mbrandenburger:
> Current version works fine in HW mode and i think it resolved the validation questions raised in the discussion. However, i think at least one comment is misleading/wrong and we should make sure the corresponding changes are also in TLCC (probably via a separate issue?). In general, for such security-sensitive code i think a bit more comments which connects the dots could be useful (although i agree that much (but not all) of the code will change with ercc refactoring ...)

I created #373 to track corresponding todos on TLCC and also addressed your concerns regarding the comments.

plugins/ecc-vscc/ecc_validation_logic.go
g2flyer: oh, i see, never saw these __lifecyce values. So definitely was wrong on that.

PR #374, opened 2020-06-17
Revisit lifecycle

mbrandenburger:

This PR addresses some open comments in issue #353 and more.

* make it printer friendly
* rename entities (use admin, peer_cli, and peer)
* remove channel and added comment in checkreadiness step
* add chaincode inspection

Fixes #

```

### CODE REVIEW

File: docs/design/fabric-v2+/fpc-lifecycle-v2.puml [hunk 1]
```diff
@@ -15,124 +15,149 @@ title Foot Box removed
...
+    end note
+
+    note right Peer_CLI
+    Other participants must approve the chaincode definition
+    by committing the corresponding transaction according to
+    the LifecycleEndorsement policy.
+
+    Note that other orgs follow the same lifecycle steps as
+    illustrated here, in particular, they inspect the FPC
+    chaincode and generate **mrenclave** accordingly.
```
g2flyer, line 140: If you really want to emphasize the mrenclave also here i would suggest we emphasize also that it must be the _same_ one (and hence we do not support the n-programming possible in Fabric since v2)
```suggestion
    Note that other orgs follow the same lifecycle steps as
    illustrated here, in particular, they inspect the FPC
    chaincode and must approve the same **mrenclave** (and hence identical code).  Requiring identical code is a deparature of the new capability added to Fabric v2 in allowing different peers to have different implementations. Unfortunately, due to confidentiality requirements, such a feature is not possible for FPC.
```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -15,124 +15,149 @@ title Foot Box removed
...
-note over Channel
-    Other participants must approve the chaincode definition,
-    and commit the corresponding transaction,
-    according to the LifecycleEndorsement policy.
-    So the party here has to loop to check for the approvals.
-end note
-end group 'other orgs approve'
+    end note
+
+    Peer_CLI -> Orderer++ : approve transaction
```
g2flyer, line 114: I think the peer is actually also (first) directly contacted by the cli.  However, i think for simplicity removing that flow here (and commit where i think the same applies) seems fine ...

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -15,124 +15,149 @@ title Foot Box removed
...
-    enalve multiple endorsers, please check out the diagram describing
-    how secret/private keys are expected to be distributed among endorsing chaincode enclaves.
+    Sequence: sequence_number,
+    Endorsement Policy: (2of3),
+    Plugins: ecc-vscc (FPC validation plugin)
+
+    NOTE: the admin approves the chaincode definition
+    including the **mrenclave** that was computed during
+    code inspection.
+
```
g2flyer, line 100: Not immediately clear what the point of this is. (Invocation of) approvefororg is by definition the approval of the chaincode, so that this also covers mrenclave should be clear from above. At least as-is i think the text is more confusing rather than further illuminating. So i would drop this.
```suggestion

```

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -15,124 +15,149 @@ title Foot Box removed
...
+    return packageId
+
+    note right Peer_CLI
+    The following bookkeeping operations help remember
+    (across lifecycle calls) that a package (id)
+    corresponds to a FPC chaincode. Similarly, the
+    approve will later store data for an FPC chaincode
+    installed locally, and the commit checks for data
+    of a locally-approved FPC chaincode.
+
```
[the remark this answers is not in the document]
mbrandenburger: I agree! The reason why this was reformatted, just to fit better in the diagram and make it more printer friendly. I wish there where a better way to define the width of the note boxes.

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -85,7 +85,7 @@ ref over Admin, Peer_CLI, Peer2
 end
 
 group approveformyorg
-    Admin -> Peer_CLI++ : lifecycle\nchaincode approveformyorg\n...**--version mrenclave**...
+    Admin -> Peer_CLI++ : lifecycle\nchaincode approveformyorg\n(version=**mrenclave**)
```
g2flyer, line 88: Not sure why you changed my original proposal (which matched the parameters of the cli) to the current version which seems to me a bit less clear but i have no strong feeling

PR #376, opened 2020-06-18
FPC Management markdown

bvavala:


docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,45 @@
...
+```peer lifecycle chaincode registereenclave <base64-encoded Credentials structure>```
+
+A successful command returns `0`.
+
+### Generate FPC Chaincode Keys
+
+This command makes an (already created) enclave generate chaincode-specific cryptographic keys,
+and registers them on the Enclave Registry (chaincode).
+
+```peer lifecycle chaincode generatekeys -n <chaincode id>```
```
g2flyer, line 41: If we expose generate keys explicitly, shouldn't we then explicit split the generation of

docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,80 @@
...
+Such definition builds on the design diagrams describing the enclave creation, 
+the enclave registration and the chaincode key generation.
+
+[//]: # (## Enclave and Chaincode APIs)
+
+[//]: # (### Create Chaincode Enclave)
+
+[//]: # (### Register Chaincode Enclave)
+
+[//]: # (### Generate FPC Chaincode Keys)
```
mbrandenburger, line 13: What is this? Can not see anything rendered from this.

docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,80 @@
+# FPC Management
```
mbrandenburger, line 1: Do we also need something like a enclave_status command? to see if there is already an enclave up and running? or some command to helps to check if we need to call create or not?

docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,80 @@
...
+Any other commands, which does not appear in this list, remains unchanged in FPC.
+
+#### Package
+
+This commands has a new option `--lang fpc-c` to specify the FPC chaincode type.
+
+#### Approveformyorg
+
+This command has the following requirements *when it is used for FPC chaincodes*:
+* `--version <MRENCLAVE>`, the version of the FPC chaincode must contain a string that represent the identity of the enclave. This is the same identity which the trusted hardware computes and attests to.
```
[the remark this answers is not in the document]
bvavala: * added the encoding
* added the file (though with no paths, as it's too early to fix it)

docs/design/fabric-v2+/fpc-management.md
bvavala: removed -- those were comments

docs/design/fabric-v2+/fpc-management.md
bvavala: maybe, but I don't know what we gain at this point.
perhaps we should think about meaningful error codes for these situations.

docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,72 @@
...
+* `--validation-plugin string`, this flag is not supported in FPC.
+
+#### Checkcommitreadiness
+
+This command has the following requirements *when it is used for FPC chaincodes*:
+* `--version <base64-encoded MRENCLAVE>`, the version of the FPC chaincode must contain a string that represent the identity of the enclave. This is the same identity which the trusted hardware computes and attests to. The string can be conveniently found in the `mrenclave` output file of the `generate_mrenclave.sh` script.
+* `--signature-policy string`, the endorsement policy must be a valid FPC endorsement policy.
+FPC currently supports only a single enclave as endorser, running at a designated peer.
+* `--endorsement-plugin string`, this flag is not supported in FPC.
+* `--validation-plugin string`, this flag is not supported in FPC.
```
g2flyer, line 36: From a maintainance perspective easier 
Maybe just to refer to approveformyorg section rather than duplicating it? Will make our life easier when updating but also i think makes it explicit that exactly the same reasoning applies in all three commands
```suggestion
This command has the same requirements *when it is used for FPC chaincodes* as described above for `approveformyorg`.
```

docs/design/fabric-v2+/fpc-management.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Commit
+
+This command has the following requirements *when it is used for FPC chaincodes*:
+* `--version <base64-encoded MRENCLAVE>`, the version of the FPC chaincode must contain a string that represent the identity of the enclave. This is the same identity which the trusted hardware computes and attests to. The string can be conveniently found in the `mrenclave` output file of the `generate_mrenclave.sh` script.
+* `--signature-policy string`, the endorsement policy must be a valid FPC endorsement policy.
+FPC currently supports only a single enclave as endorser, running at a designated peer.
+* `--endorsement-plugin string`, this flag is not supported in FPC.
+* `--validation-plugin string`, this flag is not supported in FPC.
```
g2flyer, line 45: Symmetric comment to above
```suggestion
This command has the same requirements *when it is used for FPC chaincodes* as described above for `approveformyorg`.
```

docs/design/fabric-v2+/fpc-management.md
g2flyer: I guess from a fabric symmetry perspective, a `querycreatedenclave` could make sense. Implementation should be probably quite straightforward but it will (as all thing do) require a non-trivial amount of time. As it doesn't seem essential nd we always can added it on demand, i would skip it for now.

g2flyer:
BTW: given that i have approved, there was no assignee, ive add @mbrandenburger as assignee.  As there are no fixups (and hence auto-squashing), I guess either of Marcus or Bruno can decide on my suggestions and marcus can then squash-merge

docs/design/fabric-v2+/fpc-management.md
g2flyer: > perhaps we should think about meaningful error codes for these situations.

I don't recall the peer cli have any well-defined (well-documented) exit codes other than being non-zero on errors.  So maybe just having somewhat meaningful error-messages should do.  (Not that exit codes wouldn't be a good idea, but i guess for our scripting we would do this immediatly after commit, so double registration shouldn't be any issue and we probably would have to bail anyway, no matter what ...?)

bvavala:
> BTW: given that i have approved, there was no assignee, ive add @mbrandenburger as assignee. As there are no fixups (and hence auto-squashing), I guess either of Marcus or Bruno can decide on my suggestions and marcus can then squash-merge

@mbrandenburger I am fine with Michael's suggestions. If you want me to apply the commits and merge, let me know it.

PR #378, opened 2020-06-19
Revist FPC validation diagram

mbrandenburger:

This PR completes the entire FPC flow with the last missing step - the validation.

* replace MSP validtion with custom vscc plugin

Fixes #

```

### CODE REVIEW

File: docs/design/fabric-v2+/fpc-validation.puml [hunk 1]
```diff
@@ -33,33 +33,11 @@ ref over User1, Orderer
...
 
-note right ECC3 #FFaa00
-    How can we express that we need an enclave here?
-    In particular, an enclave that has ECC_MR and also has a valid attestation?!
-end note
-
 ref over User1, Orderer
     register enclaves and key-dist
     see fpc-registration.puml and fpc-key-dist.puml
 end ref
```
g2flyer, line 39: I cannot add a suggestion as it is not in your diff, but we should mention designated peer endorsement policy as 2 of 3 actually doesn't work with the plugin approach (or we also need deterministic signatures and alike ...)

Actually, might be worth mention this limitation, this is only for MVP and we have architecture via MSP to go beyond?

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -97,22 +75,32 @@ end group
...
+
+            group default validation
+                ECC1 -> ECC1 : perform default validation
+                ECC1 -> ECC1 : validate tx (client) signature
+                ECC1 -> ECC1 : MVCC check
+                ECC1 -> ECC1 : validate endorsements
+            end
+
+            group fpc-vscc
+                ECC1 -> ECC1 : enclave_id <- extract from transaction
```
g2flyer, line 92: Don't we need here also first a parsing of the payload into actual fpc-and fpc metadata (signature, enclave id, ..)?  Extraction of id is only one part, you would need extracted signature (so it's not confused with normal endorsement signature).  Ideally, we use the same decomposition/terms as in fpc-cc-invocation ..

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -97,22 +75,32 @@ end group
...
+                create participant ERCC order 30
+                ECC1 -> ERCC++ : get_attestation(enclave_id)
+                return attestation
+                ECC1 -> ECC1 : check attestation matches enclave identity
```
g2flyer, line 96: Why do we need the attestation?  we really just care here that enclave_id was registered for that chaincode?
```suggestion
                ECC1 -> ERCC++ : check_registration(cc_name, enclave_id)
                return attestation
                ECC1 -> ECC1 : true/false
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -97,22 +75,32 @@ end group
...
+                ECC1 -> ECC1 : verify enclave signature
```
g2flyer, line 97: Worth mentioning what the signature is over? (i.e., symmetric to the generation in fpc-cc-invocation; obviously depends also on how you handled my comment above on parsing)

docs/design/fabric-v2+/fpc-validation.puml
mbrandenburger: Ok added a comment

docs/design/fabric-v2+/fpc-validation.puml
mbrandenburger: Updated the flow giving more details and sync with terms used in fpc-cc-invocation

docs/design/fabric-v2+/fpc-validation.puml
mbrandenburger: Oh good catch! I suggest the following. Let's introduce queryECCInfo and queryECCInfos. (this is similar what Lifecycle exposes for chaincode definitions). 
I update invocation diagram accordingly

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -0,0 +1,163 @@
...
+
+            group default validation
+                ECC1 -> ECC1 : perform default validation
+                ECC1 -> ECC1 : validate tx (client) signature
+                ECC1 -> ECC1 : MVCC check
+                ECC1 -> ECC1 : validate endorsements
+            end
+
+            group fpc-vscc
+                ECC1 -> ECC1 : <args, read/writeset, result> <- extract from transaction
```
g2flyer, line 96: Folks might not know what fpc-vscc is
```suggestion
            group fpc-vscc
                note over ECC1
                     fpc-vscc is a validation plugin associated with each fpc chaincode.
                end note
                ECC1 -> ECC1 : <args, read/writeset, result> <- extract from transaction
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -0,0 +1,166 @@
...
+        see fpc-cc-invocation.puml
+    end ref
+    return proposal response
+
+    User1 -> ECC2 ++: send proposal
+    ref over ECC2
+        invoke chaincode enclave
+        see fpc-cc-invocation.puml
+    end ref
+    return
```
bvavala, line 60: Given that this is a "designated peer" scenario, it is misleading to have multiple endorsers.

```suggestion
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -0,0 +1,166 @@
...
+            validation & commit
+        end ref
+        note over TLCC1
+            Ledger enclave performs the same
+            validation steps as the peer does.
+            This is necessary to let the ledger
+            enclave establish its own trusted
+            view of the ledger.
+        end note
+        |||
```
bvavala, line 144: Missing integrity metadata...
```suggestion
        end note
        ECC1 -> ECC1 : compute and store integrity-metadata for the invoked FPC chaincode's writeset
        |||
```

g2flyer:
> two small clarification suggestions

Yep, good finds! I've accepted them as an independent approval, so you or i can squash-merge.  I realize now, though, that i should have added a `Signed-off-by: ..`  as marcus has done ..  and it would have removed the need to overrule the DCO check ...

Issue #355, opened 2020-05-14
Protocol & Architecture Specifications


g2flyer:
**Description**

Complete FPC protocol and high-level overall architecture specification for MVP / RFC
- [x] Determine what diagrams are necessary (see also issue #356)
- [ ] Finish registration flow (see, e.g., issue #303 and PR #333)
- [x] Finish lifecycle flow (PR #353 & PR #374 )
- [x] cleanup/validate key-dist flows (see, e.g., issue #303 and PR #375 & #379)
- [x] cleanup/validate execution flows (see, e.g., issue #303 & #361 and PR #377 & #378)

Protocol must cover everything from [MVP milestone](https://github.com/hyperledger-labs/fabric-private-chaincode/milestone/38) but should also cover what we already have discussed as limitations and way to go forward at least as comments so prior discussions/insights is not lost)

See for more long-term work-items [Protocol & Architecture Specifications Milestone](https://github.com/hyperledger-labs/fabric-private-chaincode/milestone/9)

mbrandenburger:
Regarding the step `Determine what diagrams are necessary`

Jeb's recent update on the RFC added diagrams to the RFC.

In contains a component diagram and a series of diagrams to the describe the transaction flow. I think we should keep the component diagram and reduce the number of flow diagrams where we can.

See https://github.com/mbrandenburger/fabric-rfcs/blob/31e9afa647271dd31fdc216c4a9340fb37759629/text/0000-fabric-private-chaincode.md

Additionally, we should add a simplified version of the creation/registration and how FPC chaincode interacts with the Ledger enclave to validate input (similar to Figure 4 in the paper https://arxiv.org/pdf/1805.08541.pdf).

WDYT? @bvavala @g2flyer @jrlinton

g2flyer:
I think there are two separate questions:
1. what do we need for the high-level description for the RFC
1. what do we need for the detail design/protocol (i.e. UML diagrams)

For 2. i guess it is essentially what we already have 
- fpc-lifecycle-v2.puml
- fpc-registration.puml
- fpc-key-dist.puml
- fpc-cc-invocation.puml
- fpc-cc-execution.puml   
- fpc-validation.puml
- fpc-components.puml
(and potentially fpc-integration-touchpoints.puml as overview, maybe with cross references to the other diagrams?)

[...]

Issue #303, opened 2020-03-20
Update of UMLs following HLGF'20 discussion


g2flyer:
See corresponding [meeting minutes](https://docs.google.com/document/d/1ztdJykj4I-e9G-YGGnn2ExDVCQQHmbuhgdaA5HWkd9M/edit#heading=h.72b0xl56at64)

mbrandenburger:
PR #377 updates invocation and execution diagrams based on HLFG2020 meeting notes.

Issue #313, opened 2020-03-24
FPC Chaincode: Non-upgradability


g2flyer:
**Is your feature request related to a problem? Please describe.**

This is a non-feature, i.e., not supporting chaincode upgrading.  The reason for it is that it might help side-step the issue that ERCC might not be able to read directly a chaincode definition of a FPC chaincode from the ledger and hence doesn't know what the valid FPC version=MRENCLAVE is. But not allowing upgrades, admins can to verify the (single) enclave and the correctness of its MRENCLAVE (as visible in the attestation report). They would only engage in that chaincode if satisfied. Without upgrades, this correctness check would be an invariant holding also in the future. 

**Additional context**

See issue #312 for sibling feature

bvavala:
For the record and for future discussions, here is a possible reasonable solution to the problem.

1. implement a chaincode approval mechanism in ERCC, so that organizations pre-approve the new chaincode in the enclave registry, before installing and approving it on the channel. This does not require changes to Fabric, but it duplicates the approval mechanism in ERCC.
2. implement and run a key transfer mechanism from the old (though still installed) chaincode enclave to the new one, based on the info in ERCC. The key transfer mechanism should be similar to the key distribution mechanism described in the FPC design documents. As usual, ERCC data is validated through the trusted ledger, and the key transfer collateral is validated and stored by ERCC.
3. perform the chaincode upgrade in Fabric, with the approval of all organizations. This makes the new chaincode ready to receive invocation in the channel.
4. complete the key transfer to the new chaincode (enclave) -- which otherwise would not be able to serve requests.

g2flyer:
Wouldn't you have to at swap (at least partially) the order of 2. and 3.? Seems to me you have to first install the new chaincode and start with the lifecycle commands before you can create the enclave and hence start a key-transfer? So there would have to be some interleaving of the normal approve and the ercc approves before you then could do the ercc and cc commits, respectively?   Depends of course a bit on what atomicity guarantees you want in terms of switch-over.

bvavala:
> Wouldn't you have to at swap (at least partially) the order of 2. and 3.? Seems to me you have to first install the new chaincode and start with the lifecycle commands before you can create the enclave and hence start a key-transfer? So there would have to be some interleaving of the normal approve and the ercc approves before you then could do the ercc and cc commits, respectively? Depends of course a bit on what atomicity guarantees you want in terms of switch-over.

[...]

PR #380, opened 2020-06-23
Update registration diagram

mbrandenburger:

* Consistency with other diagrams
* rename components
* integrate new mgtn API
* replace tlcc with lifecycle for chaincode definition access
* change note color to default
* enhanced grouping for better readability
* Address comments from HLGF meeting

This PR completes #355, #134, and #303 

```

### DIFF
--- a/docs/design/fabric-v2+/fpc-registration.puml
+++ b/docs/design/fabric-v2+/fpc-registration.puml
@@ -39,10 +39,11 @@ Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id)
 
 group enclave creation
   note right Peer_CLI
-    This operation can only be performed only after the chaincode definition is
-    committed on the channel as it defined the root of trust (MRENCLAVE) for
-    the FPC Chaincode. In particular, this must be set in order to complete the
-    enclave registration step as described next.
+    This operation can only be performed after the chaincode definition --- 
+    which defines the root of trust (MRENCLAVE) for the FPC Chaincode --- 
+    is committed on the channel. More specifically, this commitment must have 
+    completed before below 'registerEnclave' starts  in order to complete the
+    enclave registration step.
   end note
 
   Peer_CLI -> ECC1 ++: createEnclave (CC_Id)
--- a/docs/design/fabric-v2+/fpc-registration.puml
+++ b/docs/design/fabric-v2+/fpc-registration.puml
@@ -196,7 +196,7 @@ end
     end note
 
     create Lifecycle order 55
-    ERCC1 -> Lifecycle +: queryChaincodeDefinition
+    ERCC1 -> Lifecycle +: queryChaincodeDefinition (Credentials.Attested_Data.CC_Parameters.CC_Id)
     return chaincode definition
 
     ERCC1 -> ERCC1    : check Credentials against chaincode definition
--- a/docs/design/fabric-v2+/fpc-registration.puml
+++ b/docs/design/fabric-v2+/fpc-registration.puml
@@ -7,23 +7,26 @@
 
 @startuml
 
+!pragma teoz true
+
 title FPC Enclave Registration
 
-actor "Admin Org1" as Admin order 10
-participant "E_Manager" as E_Manager order 15
-participant "ECC Peer1_Org1" as ECC1 order 20
-participant "ERCC Peer1_Org1" as ERCC1 order 40
-participant "ERCC Peer1_Org2" as ERCC2 order 50
+actor "Admin" as Admin order 10
+participant "Peer_CLI" as Peer_CLI order 15
+participant "FPC_stub" as ECC1 order 20
+participant "Ledger Enclave" as TLCC1 order 36 #99FF99
+participant "ERCC" as ERCC1 order 40
 
-entity IAS order 1
+entity IAS order 100
+participant Orderer as Orderer order 100
 
-note over IAS, ERCC2
+note right Admin
 Types are shown in the fpc-components diagram.
 end note
 
-note over ERCC1, ERCC2 #FFaa00
+note right Admin
   It is assumed that:
-    -channel has already been created;
+    -Channel has already been created;
     -ERCC has been installed -- approved by the participating organizations, see lifecycle;
     -ERCC contains hard-coded the TLCC_MRenclave and the IAS Root Certificate;
     -ERCC has an Endorsement Policy that requires signatures from a majority of orgs,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/design/fabric-v2+/fpc-registration.puml [hunk 1]
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
-    These operations are assumed to be performed only after the chaincode definition is
-    committed on the channel for two reasons:
-    1) chaincode invocation may not be available before the definition is committed,
-    2) enclave registration should be feasible only for committed chaincodes
-
-    PS: in a future tighter integration with fabric we could remove the first constraint
-    and could potentially piggyback the steps before the ercc registration (enclave and
-    evidence generation) in the `approveformyorg` and the registration as part of
-    `commit` without having to change the protocol flows.
-  end note
```
bvavala, line 48: This (removed) note should remain because it might be useful when, in the future, FPC will support chaincode upgrades.
The reason is that upgrading an enclave requires to change the chaincode definition. But if we commit the new definition before the key distribution, we may lose access to the old enclave. That's why something should likely happen between approve and commit.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
+  Enclave1 -> Enclave1 : quote <- create quote using SPID over Hash(Attested_Data)
+  note right Enclave1
+      There are differnt options to implmement getSPID:
+      - FPC_stub could explicitly send the SPID to FPC_stub_enclave
+      - SPID could be an optional argument within an Aux_Data field.
+      - FPC_stub_enclave performs a ocall to fetch the SPID from FPC_stub
   end note
 
-  note over Enclave1 #FFaa00
+  Enclave1 -> IAS ++: send quote for verification
```
bvavala, line 99: (just a comment -- though we had talked and agreed about this)
On a second thought, this message to IAS looks more like an implementation detail than an abstraction/generalization. Also, since it originates from the enclave/chaincode, who's familiar with Fabric will most certainly be confused. Anyway, let's look for feedback.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
+        key changes, the ERCC must be updated and upgraded through the normal
+        chaincode lifecycle upgrade operation.
+      end note
+      ERCC1 -> ERCC1 : check Credentials.Attested_Data match CC_Parameters known to ERCC
+      note right ERCC1
+        In particular Channel_Id, MSPID of sender, TLCC_MRENCLAVE is verified
+      end note
+   end group
+
+  ERCC1 -> ERCC1 : check Org-Enclave binding certificate
```
g2flyer, line 186: This is also a post-MVP item, so might have to be correspondingly marked? (But see also above)

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: > since it originates from the enclave/chaincode,

well, it's the chaincode _stub_, not the (fpc_)chaincode itself. So it wouldn't really be a contradiction?  (Note since externalbuilder, we do not really have a golang CC anymore where the golang chaincode is the stub for fpc)

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: Even with designated peer, to support upgrade we would need key-distribution which we say right now is not MVP? (but then non-upgradability (#313) is right now also MVP). 

In any case, your observation that upgrades could result in a atomicity issue which might require interleaving key-mgmt with commits is a good one. If we leave a corresponding comment (which seems a good idea), we should make it though a bit more explicit that we might need a kindof two-phase commit between the chaincode definition and ercc updates?

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
-    evidence generation) in the `approveformyorg` and the registration as part of
-    `commit` without having to change the protocol flows.
-  end note
+Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id)
 
-  note over Admin, E_Manager #FFaa00
-    The E_Manager component hides some operations that do not necessarily need to be
-    performed by the Admin -- they could be scripted or embedded in the peer process;
-    depending on the actual implementation it may require admin credential on the peer.
+group enclave creation
```
g2flyer, line 40: Right now we do external builders only in invocation and handle there the run.   Effectively, though, the run will be triggered here!  Note we refer later to external builder for storage, so the diagram is somewhat external builder aware.

Maybe just add a forward reference to the invocation part?  Or, just for our own benefit, add this observation as a puml comment, so we remember?

We also don't handle the detect and build external builder commands which would have to be in the lifecycle diagram. But that i guess seems fine, the crucial part of the external builder is really interaction of run, it's scratchspace in the filesystem and invocations ....

docs/design/fabric-v2+/fpc-registration.puml
bvavala: But that's still chaincode-side logic requesting data from a remote service. And if you implement it through an invoke, it must be a "single query".

Regarding the external builder/golang cc, I guess you mean "we do have it and use it; we could avoid it, but it requires a major effort to provide similar/same support."

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
+        The evidence check ensures that the evidence is correct (e.g. in
+        case of EPID-based attestation the evidence must be signed by IAS)
+        and covers the attested data.
+
+        Note that the IAS verification key is hardcoded in ERCC. This
+        verification key does not change frequently. In the case the verification
+        key changes, the ERCC must be updated and upgraded through the normal
+        chaincode lifecycle upgrade operation.
+      end note
+      ERCC1 -> ERCC1 : check Credentials.Attested_Data match CC_Parameters known to ERCC
```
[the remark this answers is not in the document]
bvavala: it is in the note below: channel id, tlcc mrenclave, mspid of sender

docs/design/fabric-v2+/fpc-registration.puml
bvavala: Ok for the forward reference -- it adds clarity.
Not sure documenting the external builder adds anything; perhaps, as you say, the run part, particularly for handling sealed storage.

bvavala:
just put few neutral comments; I am fine merging and then do any follow up if necessary.

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: > But that's still chaincode-side logic requesting data from a remote service. And if you implement it through an invoke, it must be a "single query".

we don't mention `invoke`, here, though, so the fact that we might initially implement that way shouldn't be confusing? 
Regarding invoke vs query: with the new flows we anyway want `query` rather than `invoke` not only for the "single" aspect you mention but also as we don't really need/want a transaction on the create --- that's what the register is for -- and it side-steps any issues like the one we just run into this morning (see #381 ) where the invoke for `getSPID` caused an (unnecessary) error. The invoke, though, would implicitly ensure that only a a proper designated peer could create an enclave, so there would be some plus: i.e., the easiest way to implement consistency with designated peer policy ....  :-)
 
> Regarding the external builder/golang cc, I guess you mean "we do have it and use it; we could avoid it, but it requires a major effort to provide similar/same support."

Well, we use golang stub and have functions which look like a (peculiar) golang chaincode, but that's all essentially invisible to the fabric core (as it is for the viewer of this diagram).  Which actually seems exactly the right way?  I.e., we do not show the particular implementation, which also is not really pertinent to understand the flows, and it leaves the freedom to do whatever is easiest now -- and i definitely was not arguing for moving away the current fpc-stub = go-stub + fpc-loader/bridge-code-ooking-like-go-chaincode! -- but leaves all freedom to change it if it ever would be useful.

docs/design/fabric-v2+/fpc-registration.puml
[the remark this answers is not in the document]
g2flyer: Not really? The note says what _match_ means. What i was pointing out that it is well defined what the match is _against_.  CC_Params can obviously not refer to what is inside Credentials.Attestat_Data, so what does it point to?  I think the test is essential as soon as you have a second enclave. 
PS: Actually, thinking now a bit more. Isn't this exactly the test what we have to do _further down_ as part of "check credentials against chaincode definition"?

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: yep, the run part is useful because of the sealing and is done already in fpc-invocation.  For the other part i also think it's not essential (otherwise i would already have commented in the fpc-invocation PR :-)

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -32,197 +35,203 @@ note over ERCC1, ERCC2 #FFaa00
...
-  note over Enclave1 #FFaa00
+  Enclave1 -> IAS ++: send quote for verification
+  note right Enclave1
     Ideally the enclave itself is able to retrieve the verifiable evidence
     (i.e., the verification report, using EPID) and return it.
-    This additionally requires some ocalls and helper code in untrusted environment.
+    This additionally requires some ocalls and helper code in FPC_stub.
     For convenience, the enclave now returns the attestation, and the evidence
     (i.e., the verification report) is retrieved from untrusted environment.
   end note
```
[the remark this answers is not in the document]
g2flyer: handled in separate PR ...

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: will (hopefully :-) be addressed in upcoming separate PR ...

Issue #372, opened 2020-06-11
Hide ecc validation plugin with peer wrapper


mbrandenburger:
**Description**

In order to use a FPC chaincode, our custom validation plugin must be used by setting it in the chaincode definition with `-V ecc-vscc` . For example:

```
${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name auction_test --version ${CC_VERS} -V ecc-vscc
```

As this may change in the future we would like to hide the use of ecc-vscc by embedding it into the peer wrapper script.

See discussion:

> Wouldn't it be better to add the validation plugin inside the peer cli?  Mid/long-term some of these plugins will change for FPC, so it would be cleaner/easier from a functional interface point of view as the interface is simpler and would be the same even if we would move to MSP/X509 style validation.  As we anyway do not allow any user-provided plugins, the peer cli of course should also barf if any plugins provided (which it currently doesn't, e.g., if you provide an endorsement plugin, which by itself is also a bit of a problem?)

_Originally posted by @g2flyer in https://github.com/hyperledger-labs/fabric-private-chaincode/pull/371_

**Link to dependent issues**
#371 re-enables custom validation plugins for ecc and ercc
#358 Management API definition  (where we should also make explicit that neither endorsement nor validation plugins are supported for FPC)

g2flyer:
Shouldn't this be in Tech Preview milestone, not MVP? It defines the functional interface and so would be related to #358 !

PR #382, opened 2020-06-24
hide validation and endorsement plugins

bvavala:
addresses #372

### DIFF
--- a/integration/auction_test.sh
+++ b/integration/auction_test.sh
@@ -30,9 +30,17 @@ auction_test() {
 
     PKG_ID=$(${PEER_CMD} lifecycle chaincode queryinstalled | awk "/Package ID: ${CC_ID}/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
 
-    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
-    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
-    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
+    # first call negated as it fails due to specification of validation plugin
+    try_rev ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER}
+
+    # first call negated as it fails due to specification of validation plugin
+    try_rev ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER}
+
+    # first call negated as it fails due to specification of validation plugin
+    try_rev ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER}
 
     try ${PEER_CMD} lifecycle chaincode querycommitted -C ${CHAN_ID}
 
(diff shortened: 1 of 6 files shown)


g2flyer:
quick comment (will do review tomorrow): demo also has some fpc chaincode installs (primarily demo/scripts/install* but maybe others). Could you change them also in this PR?  Obviously you cannot test them yet -- i still have to finalize the PR -- but from a git log perspective (and probably also to capture everything properly) it would be great if you could have it already in your commit/pr ..

bvavala:
> quick comment (will do review tomorrow): demo also has some fpc chaincode installs (primarily demo/scripts/install* but maybe others). Could you change them also in this PR? Obviously you cannot test them yet -- i still have to finalize the PR -- but from a git log perspective (and probably also to capture everything properly) it would be great if you could have it already in your commit/pr ..

I don't see anything to update there:
* still based on 1.4
* no validation plugin

mbrandenburger:
> quick comment (will do review tomorrow): demo also has some fpc chaincode installs (primarily demo/scripts/install* but maybe others). Could you change them also in this PR? Obviously you cannot test them yet -- i still have to finalize the PR -- but from a git log perspective (and probably also to capture everything properly) it would be great if you could have it already in your commit/pr ..

I don't think that the `try_rev` invocations are needed in the demo script. And since the demo script should not use -V ecc-vscc at all, nothing has to be touched in this PR I guess?!?!

g2flyer:
> I don't see anything to update there:
> 
> * still based on 1.4
> * no validation plugin

Good point. As Marcus's PR also touches lifecycle commands but in his case they actually have to be handled differently for demo (as we do have proper roles there), i have to do it anyway in my PR

g2flyer:
> By seeing this PR I think we should also block the use of custom endorsement plugins as well.
> Maybe a separate PR but I would also be fine integrating it here as a second commit.
> 
> ```
> -E, --endorsement-plugin string      The name of the endorsement plugin to be used for this chaincode
> ```

Let's do it in this PR. Note issue #372 actually _does_ discuss also the endorsement plugins, it's just not reflected in the title ....

integration/auction_test.sh
```diff
@@ -30,9 +30,17 @@ auction_test() {
...
+    try_rev ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER} -E mock-escc -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode approveformyorg -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER}
+
+    # first call negated as it fails due to specification of validation plugin
+    try_rev ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -E mock-escc -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode checkcommitreadiness -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER}
+
+    # first call negated as it fails due to specification of validation plugin
+    try_rev ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER} -E mock-escc -V ecc-vscc
+    try ${PEER_CMD} lifecycle chaincode commit -C ${CHAN_ID} --name ${CC_ID} --version ${CC_VER}
```
mbrandenburger, line 43: when I see this test for custom vscc and escc here in all integration tests I think this is a bit redundant.
Ideally, we should have a `validation_test.sh` where we just check the correct use of `-E` and `-V` for the related chaincode operations. However, we can also do that separately.

bvavala: I agree: redundant; need separate test; should be done separately.

g2flyer:
> * Please also handle the endorsement plugins in same PR, a separate PR would seem to add additional work and as mentioned in the separate comment it actually was mentioned in same issue #372 just the issue title is misleading.

Just noticed that you actually have done that already between me looking at the source and pressing the review button ...

> * a more intuitive name for try_rev (and potentially alignment with try_r) would be great

... and this also since.  Changes look good.  If you could also remove the -V from the README i'm happy to merge. I'm fine also if you do already autosquash ..

bvavala:
> Overall looks good. Three things, though:
> 
> * We also have to update the docu.  I think it's only `utils/docker-compose/README.md` -- Curiously, in examples/README we never added the -V option which actually adds another reason for why this PR is important: besides making the admin's task easier, it actually also makes it fail-safe to automatically add the plugin. If you forget, it still works but is horribly insecure ..... -- but you might cross-check that we don't refer it elsewhere (i did though search with egrep for -V & --validation-plugin, so hopefully captured all further references.

done... though kind of useless since (1) it's about `instantiate`, (2) it's for 1.4 and it will have to change anyway.

> * Please also handle the endorsement plugins in same PR, a separate PR would seem to add additional work and as mentioned in the separate comment it actually was mentioned in same issue #372 just the issue title is misleading.

done

> * a more intuitive name for try_rev (and potentially alignment with try_r) would be great

done, `try_fail` seems good.

About the testing for specific error codes, that's an actual issue but
1) we need the error codes
2) we need the tests
Good for future collaborators.

PR #384, opened 2020-06-24
Adding high-level architecture diagrams

jrlinton:
Adding the Powerpoint file which includes the high-level architecture diagrams linked in the RFC

No issue created for this

```


g2flyer:
> * we should a reference in the RFC (e.g., as markdown comments in the .md file or as a separate README) pointing to this file as the source.

I'll add it as comment to the RFC in the pass i'm just doing ....

g2flyer:
> * i would move the file into the `design` or `design/fabric-v2+` folder. (i've seen the slack discussion but thought we were talking about RFC repo, here design folder seems somewhat more appropriate. But no strong feeling about it).

BTW: the other thing regarding naming is that we wouldn't really need the v1.3 in the name as the document is now version controlled.  And names without spaces would make live on unix systems easier ;-)

Issue #309, opened 2020-03-23
Make demo work on Fabric v2


g2flyer:
**Description**
Ensure that demo also works in Fabric v2 in following scenarios (with decreasing priority)
- [X] mock-server with fpc cc
- [ ] fabric-gateway with fpc cc
- [x] ~~mock-server with golang cc~~
Depending on complexity, above might have to be broken down into further smaller work-items ..

**Link to dependent issues**
Issues #165, #254, #256

mbrandenburger:
Just tested mock-server with fpc cc and it works with current master (42b34e4)

mbrandenburger:
Just tested mock-server with go cc at least compiles but invoking create auction returned the following response:
```
No response payload received (status=500/message=json error: json: cannot unmarshal string into Go struct field Auction.owner of type golang.Principal)
```

I believe that is issue is not related to fabric v2 changes.

mbrandenburger:
Today we decided to remove golang mock chaincode, see PR #316.

mbrandenburger:
Once #316 is merged we will move this issue to MVP milestone.

Issue #98, opened 2019-06-22
Multi-org/peer example/test


g2flyer:
To make sure all our design really work and also to illustrate to users how a multi-org (and hence, multi-peer) would work we should have a test-case with at least 2 orgs and peers ...

mbrandenburger:
This is a good idea! For solid multi-peer testing the use of the client SDK will make our life much easier.

In such a test we would execute (endorse) a transaction on peer A and check if peer B successfully commits this tx.

We could use the peer CLI for that (https://hyperledger-fabric.readthedocs.io/en/release-1.4/commands/peerchannel.html#peer-channel-fetch) but this requires polling blocks and parsing them in order to check. Not nice.

mbrandenburger:
Using the fabric client SDK (https://github.com/hyperledger/fabric-sdk-go) one can easily register for events. Eg.
```
client.RegisterChaincodeEvent(ccID, eventID)
```

mbrandenburger:
seems the peer CLI indeed provides an event listener, which we can use here (and also remove the sleep command)

```
--waitForEvent                   Whether to wait for the event from each peer's deliver filtered service signifying that the 'invoke' transaction has been committed successfully
--waitForEventTimeout duration   Time to wait for the event from each peer's deliver filtered service signifying that the 'invoke' transaction has been committed successfully (default 30s)
```


See here https://hyperledger-fabric.readthedocs.io/en/release-1.4/commands/peerchaincode.html#peer-chaincode-invoke

g2flyer:
relevant pointers of setting up multi-peer/multi-org:
https://hyperledger-fabric.readthedocs.io/en/latest/channel_update_tutorial.html

g2flyer:
Moving to multi-peer we also will have to adapt the handling of __setup in the peer wrapper by persisting info on whether setup has happened and check before invoke and query whether it has happened and if not, call it there.  See issue #141 and PR #144

g2flyer:
Note: while in old lifecycle, our peer.sh wrapper logic should have worked with multiple peers and the ercc install&instantiate and the ecc onDemand setup, with the new lifecyle this is _not_ true anymore, e.g., the ercc configuration in channel join would require tighter coordination between multiple peers.  So this might require now more explicit additional management commands but also has to be looked at from the perspective of the additional steps necessary for th  new key-managment protocols

g2flyer:
depends on resolution of issue #12

g2flyer:
[...]

Issue #373, opened 2020-06-16
TLCC revise FPC transaction validation


mbrandenburger:
**Description**

TLCC FPC transaction validation logic needs to be aligned with recent changes in ecc-vscc as introduced with PR #371. 

Another issue is that the current tlcc code requires a `ecc_` prefix which was removed with PR #320. Quick and dirty fix applied but needs to resolved. More detailed discussion on PR 323

bvavala:
Additional info in:
* comment https://github.com/hyperledger-labs/fabric-private-chaincode/pull/388/files#r447985598 (including suggestion to look for the validation plugin)

PR #388, opened 2020-06-26
Upgrade Docker-compose network to 2.x and re-enable demo

### LINKED ISSUE #165: Change our examples and demo to fabric 2.0 lifecycle commands

### LINKED ISSUE #256: Upgrade docker-compose ledger setup scripts & config to v2.0
move `utils/docker-compose/*` to fabric v2.   Ideally, should be as far as possible devoid of legacy (e.g., use raft rather deprecated solo).

Might also consider looking at new test-network in fabric-samples which should supercede byfn (on which current scripts are based) and, potentially, minifabric (https://github.com/litong01/minifabric/blob/master/docs/README.md).

Related to (and probably best done after) #254

g2flyer:

Subject says most.  Noteworthy changes:
- roles and related policy checks are enabled
- yaml have now clear provenance
- scripts/docker-compose honor FABRIC_LOGGING_SPEC env var

With these changes also the scripted demo works again.
_However_, there is a bug in `demo/client/backend/fabric_gateway`  where `getRegisteredUsers` does returns an incorrect (empty) list which renders UI unusable. Hence, this PR does _not_ yet address #309.  If i can fix it before the review is over, i might add it to this PR. Otherwise it will be a follow-up PR.

Another remaining issue is that we should enable also to run the cli from the host and not do first a `docker exec` into the peer to run it.  I've done already some experiments but there seem to be some non-trivial (or at least non-obvious) hurdles to cross first. So this probably will have to wait for the PR addressing #98 (multi-org). 

Closes #256 and mostly closes #165.  The mostly is due that it doesn't move from old fabric-client to new fabric-network node API but i guess that is something we can do on the back burner for now given that we don't touch node client SDK for MVP.

As some docker images changed, you might want to do a `make clobber` first before running tests.
Also, in case you still have an old (1.4) `./util/docker-compose/bin` directory, delete it!

No


plugins/Makefile
```diff
@@ -35,16 +41,16 @@ godeps:
 	$(GO) get github.com/hyperledger/fabric-chaincode-go/shimtest
 
 .PHONY: ecc-vscc
-ecc-vscc: $(ECC_VSCC_SRC)
-	$(GO) build -o $(ECC_VSCC_OUT) -buildmode=plugin $^
+ecc-vscc: $(ECC_VSCC_SRC) $(ATTESTATION_PROVIDER)
+	$(GO) build -o $(ECC_VSCC_OUT) -buildmode=plugin $(ECC_VSCC_SRC)
```
mbrandenburger, line 45: Why making attestation provider a dependency here? Looking into `ecc_validation_logic.go` there is no import for `ercc/attestation`?!

utils/docker-compose/network-config/core-fpc.yaml
```diff
@@ -389,25 +405,26 @@ peer:
...
             name: DefaultValidation
             library:
           # FPC Addition:2
           ercc-vscc:
-            name: ercc-vscc
-            library: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/ercc/ercc-vscc.so
-          # FPC Addition:3 
+             name: ercc-vscc
+             library: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/plugins/ercc-vscc.so
+          # # FPC Addition:3
```
mbrandenburger, line 424:
```suggestion
          # FPC Addition:3 
```

utils/docker-compose/scripts/bootstrap.sh
```diff
@@ -10,22 +10,33 @@
...
+
 printHelp() {
   echo "Usage: bootstrap.sh [<version>] [<ca_version>] [<thirdparty_version>][-d -b]"
   echo
   echo "-d - bypass docker image download"
   echo "-b - bypass download of platform-specific binaries"
   echo
-  echo "e.g. bootstrap.sh 1.4.1 1.4.1 0.4.15"
+  echo "e.g. bootstrap.sh 2.1 1.4 0.4.15"
   echo "would download docker images and binaries for version 1.4.0 (fabric) 1.4.0 (fabric-ca) 0.4.15 (thirdparty)"
```
mbrandenburger, line 40:
```suggestion
  echo "e.g. bootstrap.sh 2.1 1.4 0.4.15"
  echo "would download docker images and binaries for version 2.1 (fabric) 1.4 (fabric-ca) 0.4.15 (thirdparty)"
```

utils/docker/dev/Dockerfile
```diff
@@ -34,11 +34,14 @@ WORKDIR /tmp
...
 
-# we need docker commands (even though we use docker daemon from "outside")
+# Notes:
+# - we need docker commands (even though we use docker daemon from "outside")
+# - jq we need for e.g., external-builder
 RUN apt-get update \
  && apt-get install -y \
     docker.io \
     clang-format \
+    jq\
```
mbrandenburger, line 44:
```suggestion
    jq \
```

utils/docker/peer/Dockerfile
```diff
@@ -34,11 +34,14 @@ WORKDIR /tmp
...
 
-# we need docker commands (even though we use docker daemon from "outside")
+# Notes:
+# - we need docker commands (even though we use docker daemon from "outside")
+# - jq we need for e.g., external-builder
 RUN apt-get update \
  && apt-get install -y \
     docker.io \
     clang-format \
+    jq\
```
mbrandenburger, line 44:
```suggestion
    jq \
```

plugins/Makefile
g2flyer: Can't remember what failed but something definitely failed ...

Makefile
```diff
@@ -6,10 +6,11 @@
...
 
-# TODO bring back demo
-# SUB_DIRS = utils ercc ecc_enclave ecc tlcc_enclave tlcc examples integration demo # docs
-SUB_DIRS = common utils ercc ecc_enclave ecc tlcc_enclave tlcc plugins fabric examples integration
-FPC_SDK = utils/fabric ecc_enclave ecc
+SUB_DIRS = common utils ercc ecc_enclave ecc tlcc_enclave tlcc plugins fabric examples integration demo # docs
+
+FPC_SDK = utils/fabric common ecc_enclave ecc
+FPC_PEER = plugins tlcc_enclave tlcc fabric ecc_enclave ecc # ecc & ecc_enclave we need if we run on host in external builder, not via docker
+FPC_PEER_CLI = utils/fabric
```
bvavala, line 13: I was confused by this, until I realized that that is not the cli, nor there is anything to build about that.
Suggestion, for both this is the target: `FPC_UTILS` or similar.

utils/docker-compose/network-config/core-fpc.yaml
```diff
@@ -269,7 +287,7 @@ peer:
     # Path on the file system where peer will store data (eg ledger). This
     # location must be access control protected to prevent unintended
     # modification that might corrupt the peer operations.
-    fileSystemPath: /tmp/hyperledger/test
+    fileSystemPath: /var/hyperledger/production
```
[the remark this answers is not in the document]
g2flyer: The /var location is the default location from fabric.  As this file is used inside docker where everything runs as root, permissions is a non-issue ..

Makefile
g2flyer: it the list of build dependency directories to run peer cli.   It matches the pattern we had for FPC_SDK and FPC_PEER.   Arguably the pattern doesn't capture that it is a list of directories and that they form dependencies.    It would be more verbose but maybe more clear to change all three to `..._DEPEND_DIRS` .   Thoughts? A better name?

utils/docker-compose/network-config/core-fpc.yaml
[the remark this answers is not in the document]
bvavala: good, this makes me happy

Makefile
bvavala: > it the list of build dependency directories to run peer cli.

so, as I said, it's not the CLI.

> It matches the pattern we had for FPC_SDK and FPC_PEER. 

though FPC_PEER contains fabric, which will eventually build the peer.

> Arguably the pattern doesn't capture that it is a list of directories and that they form dependencies. It would be more verbose but maybe more clear to change all three to `..._DEPEND_DIRS` . Thoughts? A better name?

Whatever is less misleading, `..._DEPEND_DIRS`, `..._DEPS`, ...

utils/docker-compose/network-config/core-fpc.yaml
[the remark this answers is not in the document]
g2flyer: I'm glad to be of service;-)

Makefile
g2flyer: > > It matches the pattern we had for FPC_SDK and FPC_PEER.
> 
> though FPC_PEER contains fabric, which will eventually build the peer.

I guess you are correct that our fpc peer (wrapper) also depends on peer and so strictly speaking FPC_PEER_CLI should also include $FPC_PEER.  That said, from a cli perspective we actually do not really need FPC-modifed peer binary and in some docker script scenario it actually might currently run on the downloaded rather than self-built binaries ...

plugins/Makefile
mbrandenburger: I guess it failed for `ercc-vscc`.

utils/docker-compose/network-config/core-fpc.yaml
```diff
@@ -389,25 +405,26 @@ peer:
...
-          # FPC Addition:1
+            # FPC Addition:1
             name: ERCCDecorator
-            library: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/ercc/ercc-decorator.so
+            library: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/plugins/ercc-decorator.so
         endorsers:
           escc:
             name: DefaultEndorsement
             library:
+        #TODO: Enable below FPC Additions once validation plugins are re-intergrated
```
mbrandenburger, line 415: this todo looks obsolete?

mbrandenburger:
When running the demo script from docker dev it looks a bit better:
```
# '${DOCKER_COMPOSE} ps' to get status of docker compose network or
# '${DOCKER_COMPOSE} logs peer0.org1.example.com' for peer logs, e.g., to investigate start errors
export \
 COMPOSE_PROJECT_NAME="fabricfpc"\
 FABRIC_VERSION="2.1.1"\
 PEER_CMD="/project/src/github.com/hyperledger-labs/fabric-private-chaincode/fabric/bin/peer.sh"\
 FPC_CONFIG="-fpc"\
 CHANNEL_NAME="mychannel"\
 DOCKER_COMPOSE="docker-compose -f /project/src/github.com/hyperledger-labs/fabric-private-chaincode/utils/docker-compose/network-config/docker-compose.yml -f /project/src/github.com/hyperledger-labs/fabric-private-chaincode/utils/docker-compose/network-config/docker-compose.yml"

Removing network fabricfpc_basic
WARNING: Network fabricfpc_basic not found.
Creating network "fabricfpc_basic" with the default driver
Creating ca.example.com ...
Creating orderer.example.com ...
Creating ca.example.com
Creating orderer.example.com ... done
Creating ca.example.com ... done
Creating peer0.org1.example.com ... done
         Name                       Command                State             Ports
-------------------------------------------------------------------------------------------
ca.example.com           sh -c fabric-ca-server sta ...   Up         0.0.0.0:7054->7054/tcp
orderer.example.com      orderer                          Up         0.0.0.0:7050->7050/tcp
peer0.org1.example.com   /project/src/github.com/hy ...   Exit 111
Error response from daemon: Container fbede6152857a5e9cb5f7b9c550c673c7d41990742b5412526f5fcb835a5dd35 is not running
```
But another error and the peer crashes?!

plugins/Makefile
g2flyer: ok, seems to work fine when having dependency only on ercc-vscc.

g2flyer:
I moved it back to draft as i will have to do some additional adoption after #390 is merged (which certainly should be merged first)

g2flyer:
> Note that I am trying to run this directly from Mac (not using docker dev here) and it seems that it looks for cmake on my mac? But I thought that stuff will run inside the docker container including cmake?

The scripts check whether you have all docker images and if not, try to build them but obviously do that locally. I didn't change anything in that behavior, that was always like that.

Regarding the docker-compose error, you have to provide me the peer logs, i.e., the output of `$DOCKER_COMPOSE logs peer0.org1.example.com` (after copy/pasting the `export ...` statement from the stdout (maybe `docker logs  peer0.org1.example.com` might also work ..)

BTW: just to make sure, did you do a `make clobber` before any rebuilding?  Oh, and another thing, do you have by chance still an old (1.4) `./util/docker-compose/bin` directory: if so, delete it!

g2flyer:
> > Note that I am trying to run this directly from Mac (not using docker dev here) and it seems that it looks for cmake on my mac? But I thought that stuff will run inside the docker container including cmake?
> 
> The scripts check whether you have all docker images and if not, try to build them but do that local to where you call the script. I didn't change anything in that behavior, that was always like that, so i think that never did work on a Mac. (I think it they probably do work though from outside after you run a `make` inside the dev container ...)
> 
> Regarding the docker-compose error, you have to provide me the peer logs, i.e., the output of `$DOCKER_COMPOSE logs peer0.org1.example.com` (after copy/pasting the `export ...` statement from the stdout (maybe `docker logs peer0.org1.example.com` might also work ..)
> 
> BTW: just to make sure, did you do a `make clobber` before any rebuilding? 
Oh, and another thing, do you have by chance still an old (1.4) `./util/docker-compose/bin` directory: if so, delete it!  I was contemplating on removing it on `teardown --clear-slate` but so far we don't clean any fabric artifacts (besides this `bin` e.g., also fabric docker images) and at least at home the download also take quite some time (although in the context of something like it takes for a `make clobber all`.  If you think we should remove, then i could add that ...

mbrandenburger:
I did a `make clean / clobber / deleted` `docker-compose/bin`; `make` in `utils/docker`; started a dev container running `make run`;

Inside the container I ran `make` ... it took ages and my batterie went from 100% to 55% ... this entire docker setup is not OK! At this point I propose to drop everything which is related to start docker inside docker and build fpc again inside another docker. Also the our way to build a container with the build tags to get the "right" branch takes ages .... most of the time waiting and see this message `Sending build context to Docker daemon  xxxxGB` ticking slowly up. ridicules.  

Regarding the peer error above, the logs show me the following:
```
scripts/startFPCAuctionNetwork.sh 

...

Removing network fabricfpc_basic
WARNING: Network fabricfpc_basic not found.
Creating network "fabricfpc_basic" with the default driver
Creating ca.example.com ...
Creating orderer.example.com ...
Creating ca.example.com
Creating orderer.example.com ... done
Creating peer0.org1.example.com ...
Creating peer0.org1.example.com ... done
         Name                       Command                State             Ports
-------------------------------------------------------------------------------------------
ca.example.com           sh -c fabric-ca-server sta ...   Up         0.0.0.0:7054->7054/tcp
orderer.example.com      orderer                          Up         0.0.0.0:7050->7050/tcp
peer0.org1.example.com   /project/src/github.com/hy ...   Exit 111

...

root@docker-desktop:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo# docker logs peer0.org1.example.com
peer.sh: FABRIC_BIN_DIR not properly defined as '/project/src/github.com/hyperledger/fabric/build/bin'
```

Also, it seems that the cmake/die related issues as reported above are gone by running it from dev container.

Also, when running `scripts/startFPCAuctionNetwork.sh --build-cc`, I got:
```
/libcxx  -m64 -O2 -nostdinc -fno-builtin -fvisibility=hidden -fpie -fstack-protector -std=c++11 -nostdinc++ -fPIC   -o CMakeFiles/enclave.dir/bid.cpp.o -c /project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc/bid.cpp
[100%] Linking CXX shared library enclave.so
/usr/bin/cmake -E cmake_link_script CMakeFiles/enclave.dir/link.txt --verbose=1
/usr/bin/c++ -fPIC  -m64 -O2 -nostdinc -fno-builtin -fvisibility=hidden -fpie -fstack-protector -std=c++11 -nostdinc++  -shared -Wl,-soname,enclave.so -o enclave.so CMakeFiles/enclave.dir/dispatcher.cpp.o CMakeFiles/enclave.dir/entry.cpp.o CMakeFiles/enclave.dir/error-codes.cpp.o CMakeFiles/enclave.dir/spectrum-auction.cpp.o CMakeFiles/enclave.dir/spectrum-auction-message.cpp.o CMakeFiles/enclave.dir/storage.cpp.o CMakeFiles/enclave.dir/principal.cpp.o CMakeFiles/enclave.dir/territory.cpp.o CMakeFiles/enclave.dir/eligibility.cpp.o CMakeFiles/enclave.dir/bidder.cpp.o CMakeFiles/enclave.dir/auction-state.cpp.o CMakeFiles/enclave.dir/utils.cpp.o CMakeFiles/enclave.dir/evaluate.cpp.o CMakeFiles/enclave.dir/bid.cpp.o -m64 -O2 -Wl,--no-undefined -nostdlib -nodefaultlibs -nostartfiles -L/project/src/github.com/hyperledger-labs/fabric-private-chaincode/ecc_enclave/enclave/../../common/crypto/_build -L/project/src/github.com/hyperledger-labs/fabric-private-chaincode/ecc_enclave/enclave/../../ecc_enclave/_build/lib -Wl,--start-group -ltpdo-crypto -lshim -Wl,--end-group -L/opt/intel/sgxssl/lib64 -Wl,--whole-archive -lsgx_tsgxssl -Wl,--no-whole-archive -lsgx_tsgxssl_crypto -L/opt/intel/sgxsdk/lib64 -Wl,--whole-archive -lsgx_trts_sim -Wl,--no-whole-archive -Wl,--start-group -lsgx_tstdc -lsgx_tcxx -lsgx_tcrypto -lsgx_tservice_sim -Wl,--end-group -Wl,-Bstatic -Wl,-Bsymbolic -Wl,--no-undefined -Wl,-pie,-eenclave_entry -Wl,--export-dynamic -Wl,--defsym,__ImageBase=0 -Wl,--version-script=/project/src/github.com/hyperledger-labs/fabric-private-chaincode/ecc_enclave/enclave/../../ecc_enclave/enclave/enclave.lds
/usr/bin/ld: cannot find -ltpdo-crypto
/usr/bin/ld: cannot find -lshim
collect2: error: ld returned 1 exit status
make[3]: *** [enclave.so] Error 1
CMakeFiles/enclave.dir/build.make:435: recipe for target 'enclave.so' failed
make[3]: Leaving directory '/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc/_build'
CMakeFiles/Makefile2:70: recipe for target 'CMakeFiles/enclave.dir/all' failed
make[2]: Leaving directory '/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc/_build'
make[2]: *** [CMakeFiles/enclave.dir/all] Error 2
Makefile:86: recipe for target 'all' failed
make[1]: Leaving directory '/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo/chaincode/fpc/_build'
make[1]: *** [all] Error 2
Makefile:21: recipe for target 'build' failed
make: *** [build] Error 2
Makefile:33: recipe for target 'docker-build' failed
make: *** [docker-build] Error 2
```

Running `make` in `demo/chaincode/fpc`, however, runs through inside the dev container.

I think I agree with you that most of the issue I experience here were here before and not introduced by this PR.

g2flyer:
> I did a `make clean / clobber / deleted` `docker-compose/bin`; `make` in `utils/docker`; started a dev container running `make run`;
> 
> Inside the container I ran `make` ... it took ages and my batterie went from 100% to 55% ...

Maybe a power supply would help ;-?

>  this entire docker setup is not OK! At this point I propose to drop everything which is related to start docker inside docker and build fpc again inside another docker. 

I know it's saturday and i guess you thought you can do a quick test which turned out a not-so-quick test but let's take first a step back before jumping to to many conclusion.

- travis reports 29 min for a full build, which is i think faster than doing a full build of fabric, so not outlandish.
  Just building the base images cost a sizable amount of that as do building the demo images (and it's thousands of npm packages!!), in particular if you are not on a super fast network. And this seems somewhat unavoidable, no matter what.

- clobber forces a rebuild of everything.  It is far from always needed, just easiest way to make sure everything should be build freshly and cleanly from scratch, e.g., after a major change.  (You can always selectively force clean builds of individual containers by defining FORCE_REBUILD when running the corresponding make.)

- to make docker run on clean and well-defined states, they are also run from a particular .git state rather than volume mounted.  This causes some overhead of the COPY operation (see below) but imho is definitely the right trade-off.

That said, there is potential room for improvement (though afaict no low-hanging one):

- we currently do not have dependencies on docker images and they are unconditionally (re-)build. Given some separate cross-links even several times.
  Docker caching in general makes rebuild fast when nothing has changed, so for many layer nothing is really re-executed. The exception is for COPY as used for accessing local .git in 'dev' and 'peer' builds.  In these cases, there might be some redundant recomputation (although less than before the tagging changes, see below). Although i don't think it's a large part of the overall build time.

  So causing docker builds only dependencies has changed would in principle help but in practice i don't really know how to specify such dependencies.
  What might be more tractable (but still quite some work) is to redirect the dependency graph among makefiles and scripts that the make targets in `utils\docker` are not hit from various separate places during builds and/or script runs.

  In any case, i think we have more pressing issues than speeding up initial builds.

- what might be nice for casual end-users would be some script/make target which builds the dev container and runs just all targets necessary to build all containers so the scripts would work from outside as you initially tried.  However, mid-term we would want to distributed pre-built docker images as fabric has done which makes this a bit a moot exercise and so also not worth the effort.


> Also the our way to build a container with the build tags to get the "right" branch takes ages .... most of the time waiting and see this message `Sending build context to Docker daemon xxxxGB` ticking slowly up. ridicules.

For which docker image does this happen?

This message afaicr is shown when stuff is COPYed or ADDed from the local space to daemon.  We do that in peer and dev container. However, it should only copy $FPC_PATH/.git which is even for the case where i have 6 different `remote', is only 37M. Originally, we COPYed all of FPC_PATH which potentially could be _much_ larger but i've changed that to .git when i added the image labeling. So it really shouldn't copy GBs for the main images.

That said, we do COPY also directories in demo/client/backend/fabric-gateway, so if you have lots of stuff laying round there, maybe that causes extra copies?


> Regarding the peer error above, the logs show me the following:
> 
> ```
> scripts/startFPCAuctionNetwork.sh 
> 
> ...
> 
> Removing network fabricfpc_basic
> WARNING: Network fabricfpc_basic not found.
> Creating network "fabricfpc_basic" with the default driver
> Creating ca.example.com ...
> Creating orderer.example.com ...
> Creating ca.example.com
> Creating orderer.example.com ... done
> Creating peer0.org1.example.com ...
> Creating peer0.org1.example.com ... done
>          Name                       Command                State             Ports
> -------------------------------------------------------------------------------------------
> ca.example.com           sh -c fabric-ca-server sta ...   Up         0.0.0.0:7054->7054/tcp
> orderer.example.com      orderer                          Up         0.0.0.0:7050->7050/tcp
> peer0.org1.example.com   /project/src/github.com/hy ...   Exit 111
> 
> ...
> 
> root@docker-desktop:/project/src/github.com/hyperledger-labs/fabric-private-chaincode/demo# docker logs peer0.org1.example.com
> peer.sh: FABRIC_BIN_DIR not properly defined as '/project/src/github.com/hyperledger/fabric/build/bin'
> ```

I've seen similar to above and below in the past.  One issue it reminds me could be if you still still had some concept-release 1.0 images laying around: the scripts use 'latest' as default FPC_VERSION. So you might want to cross-check with `docker images | grep private` that no latest points to an old version. (or if so, remove the latest tag from that image.)
That said, when i try it on one of my system, a `make clobber all` doesn't cause me a related failure even though i have cr1.0.1 images originally tagged as latest.


> Also, when running `scripts/startFPCAuctionNetwork.sh --build-cc`, I got:
> 
> ```

g2flyer:
btw: i think there is also a small back in `utils/docker-compose/README.md`: where it says `CORE_PEER_MSPCONFIGPATH=/etc/hyperledger/msp/users/Admin@org1.example.com/msp` when installing helloworld, it should read `export CORE_PEER_MSPCONFIGPATH=/etc/hyperledger/msp/users/Admin@org1.example.com/msp`  (as otherwise the actual peer still uses peer rather than admin credentials when installing).  Corresponding commit will come later ..

g2flyer:
> I've finally managed to run `make all` on this PR successfully. In particular, the scripted demo runs through. Also demo network starts manually and the frontend shows the login window as expected.

good to know

> However, I also see the issue @g2flyer has reported. I suggest to fix this in a separate PR.

Sounds like a good idea, in particular, as i end up moving now also fabric node modules from 1.4 to 2.1 (and related to that, also node 8 to node 12) as part of fixing demo.

BTW: i did now also include the changes necessary for the MRENCLAVE versioning and as already rebased them, i ended up also just squashing them together. I hope it's still easy enough to see that i've addressed your past comments.

demo/scripts/installCC.sh
```diff
@@ -10,11 +10,38 @@ set -ev
...
-docker exec peer0.org1.example.com ${PEER_CMD} chaincode install -n auctioncc -v ${version} --path demo/chaincode/fpc/_build/lib -l fpc-c
-sleep ${WAIT_TIME}
+# Command to execute peer cli inside peer container 
+# Note: by default the peer runs with peer credentials, so we have to override CORE_PEER_MSPCONFIGPATH env-var that the peer in cli-mode doesn't use peer but admin credentials
+REMOTE_PEER_CMD="docker exec -e CORE_PEER_LOCALMSPID=Org1MSP -e CORE_PEER_MSPCONFIGPATH=/etc/hyperledger/msp/users/Admin@org1.example.com/msp peer0.org1.example.com env TERM=${TERM} ${PEER_CMD}"
+REMOTE_CAT="docker exec peer0.org1.example.com cat"
 
-docker exec peer0.org1.example.com ${PEER_CMD} chaincode instantiate -n auctioncc -v ${version} --channelID mychannel -c '{"Args":[]}'
-sleep ${WAIT_TIME}
+# Commands to execute cli _outside_ peer container
```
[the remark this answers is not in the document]
g2flyer: Eventually we should have cli running from client host rather than docker execing into peer.  Doesn't 100% work yet, so no default but unless you have strong objection i would prefer to keep these comments there ...

demo/README.md
```diff
@@ -25,7 +25,7 @@ scripts/startFPCAuctionNetwork.sh --build-cc
 
 The channel `mychannel` will be created and used to install and instantiate
 all chaincodes. The [FPC auction chaincode](chaincode/fpc) will be
-instantiated as `ecc_auctioncc`.  If you do not need to build
+instantiated as `auctioncc`.  If you do not need to build
```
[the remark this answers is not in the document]
g2flyer: See commits fadc9f13208bca058d314bedce64eda211e285d5 and c49e394e3e8f0e09cb8f2f96687977d2c00c4eed for details. In short, right now it's based on a state key with well-known name (MRENCLAVE). When refactoring, the use in ercc_validation becomes obsolete and for tlcc we can take presence of ecc_validation plugin as the indicator. I think there is some issue somewhere with some discussion ...

g2flyer:
> I just would like:
> 
> * a comment/issue/documentation on the `ecc_` prefix (see notes)

I assume a comment here on github (which i've done), not in the code?  (There is no prefix anywhere in public interface, so talking of (absence of) prefix in readme would be strange ...

> Also, good to have in this PR:
> 
> * better cleanup in `demo/Makefile` (below it's my diff)
> 
> ```
> -build test clean:
> +build test:
>         $(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
> 
> -clobber:
> +clean:
> +       ./scripts/teardown.sh
> +       $(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
> +
> +clobber: clean
> ```

Clean doing teardown makes sense but clobber should still do clean-slate or you might have other problems ..

> * better cleanup of `utils/docker-compose/Makefile` to remove bin folder (below it's my diff)
> 
> ```
> -clobber:
> +clean:
> +       rm -rf bin
> +
> +clobber: clean
> ```

I don't think we should remove bin on clean (we are not removing any artifacts here for clean) but i'm fine doing it on clobber.

Pushed now corresponding fixup

demo/scripts/installCC.sh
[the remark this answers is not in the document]
bvavala: sounds good

demo/README.md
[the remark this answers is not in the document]
bvavala: that's the comment I was looking for.
I could not see an issue for that, so I'll create one.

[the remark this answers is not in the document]
g2flyer: > I could not see an issue for that, so I'll create one.

i think issue #373 should cover it ...

bvavala:
> > I just would like:
> > 
> > * a comment/issue/documentation on the `ecc_` prefix (see notes)
> 
> I assume a comment here on github (which i've done), not in the code? (There is no prefix anywhere in public interface, so talking of (absence of) prefix in readme would be strange ...
> 
> > Also, good to have in this PR:
> > 
> > * better cleanup in `demo/Makefile` (below it's my diff)
> > 
> > ```
> > -build test clean:
> > +build test:
> >         $(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
> > 
> > -clobber:
> > +clean:
> > +       ./scripts/teardown.sh
> > +       $(foreach DIR, $(SUB_DIRS), $(MAKE) -C $(DIR) $@ || exit;)
> > +
> > +clobber: clean
> > ```
> 
> Clean doing teardown makes sense but clobber should still do clean-slate or you might have other problems ..
> 
> > * better cleanup of `utils/docker-compose/Makefile` to remove bin folder (below it's my diff)
> > 
> > ```
> > -clobber:
> > +clean:
> > +       rm -rf bin
> > +
> > +clobber: clean
> > ```
> 
> I don't think we should remove bin on clean (we are not removing any artifacts here for clean) but i'm fine doing it on clobber.
> 
> Pushed now corresponding fixup

These address my issues.

PR #398, opened 2020-07-08
Integrate fpc-validation as compiled validator

mbrandenburger:

This PR integrates FPC custom validation into the peer without using go plugins. 

- Add FPC Validators (including ERCC and ERCC-decorator) to handler library using FPC build tag
- move from plugin folder to validation folder
- Rename ecc-vscc to fpc-vscc
- Deprecate ercc-vscc and ercc decorator
- Align FPCValdation plugin interface with DefaultValdator
- use protoutils to unmarshal all protos
- refactor checkEnclaveEndorsement for better readability and
maintainability

Fixes #

```

### CODE REVIEW

File: fabric/Makefile [hunk 1]
```diff
@@ -75,24 +76,28 @@ clean-patch:
 $(FABRIC_TLCC):
 	ln -sfn $$(/bin/pwd)/../tlcc $(FABRIC_TLCC)
 
+$(FABRIC_VALIDATION):
+	ln -sfn $$(/bin/pwd)/../validation/fpc $(FABRIC_VALIDATION)
+
```
mbrandenburger, line 81: I am wondering if we really need this.

g2flyer: I hope not. that would kindof defeat a bit the purpose of build tags?

bvavala: I did the same... and for a reason.
The validator has to access the Go protobufs, which "we" have in the go path, while "Fabric" has then in the vendor folder.
So, without this link, go build complains that it wants the protobufs in Fabric's vendor folder (following a similar path: `github.com/hyperledger/...`). With this link, instead, go build grabs the protobufs from vendor and it's happy. In this case, we might think of dropping our FPC dependency on protobufs (though TLCC might still require them).

g2flyer:
This still relies on the plugin framework and so doesn't really help getting rid of plugins and will also require changes once fabric does get rid of plugins.  So imho technically moderately meaningful. But if that puts us Gary off the back, so be it.

From looking at `evaluateModeAndLoad` in `.../library/registry.go`, this should work also for all four plugin types, , so we should do conversion of our other two plugins at the same time.  We should definitely get rid of the symlinks (see separate discussion above).

fabric/Makefile
g2flyer: i doubt that the symlinks will be acceptable: what would happen if we also vendor protos in fpc?  That said, wouldn't vendoring become obsolete once you move to go modules. Actually, fabric-protos-go _is_ modularized and fabric does include fabric-protos-go in go.mod. I think the version in vendor is (should be) obsolete!

validation/fpc/fpc_validation.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
```
g2flyer, line 2: Moving stuff out of plugins makes sense if we do not use plugins anymore. But a `git mv` would have kept the changes smaller and kept provenance?

g2flyer: btw: you also will have to update core-fpc.yaml in utils/docker-compose ...

fabric/Makefile
mbrandenburger: as discussed with @g2flyer we keep the symlinks until we bump to fabric v2.2

validation/fpc/fpc_validation.go
mbrandenburger: updated

validation/fpc/fpc_validation_logic.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// validate the payload type
+	if common.HeaderType(chdr.Type) != common.HeaderType_ENDORSER_TRANSACTION {
+		logger.Errorf("Only Endorser Transactions are supported, provided type %d", chdr.Type)
+		err = fmt.Errorf("Only Endorser Transactions are supported, provided type %d", chdr.Type)
+		return nil, err
+	}
```
g2flyer, line 97: This is something we haven't done before. From `vendor/github.com/hyperledger/fabric-protos-go/common/common.pb.go` no other type seems relevant (and that's why we probably always received only endorser transactions) but definitely robust programming to explicitly check ...

Makefile
```diff
@@ -6,10 +6,10 @@
 TOP = .
 include $(TOP)/build.mk
 
-SUB_DIRS = common utils ercc ecc_enclave ecc tlcc_enclave tlcc plugins fabric examples integration demo # docs
+SUB_DIRS = common utils ercc ecc_enclave ecc tlcc_enclave tlcc fabric examples integration demo # docs
 
 FPC_SDK_DEP_DIRS = utils/fabric common ecc_enclave ecc
-FPC_PEER_DEP_DIRS = plugins tlcc_enclave tlcc fabric ecc_enclave ecc # ecc & ecc_enclave we need if we run on host in external builder, not via docker
+FPC_PEER_DEP_DIRS = ercc tlcc_enclave tlcc fabric ecc_enclave ecc # ecc & ecc_enclave we need if we run on host in external builder, not via docker
```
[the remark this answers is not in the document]
mbrandenburger: the validation does not have any build artefacts anymore as the code is used when build in the peer in `fabric/`. In contrast, in `plugins/` we compiled the actual plugins, so I would say from regarding build structure they are now equivalent. 

However, since the ercc-vscc needs the `provider.go` in `ercc/attestation/`, we need to compile this first, before running `make` in `fabric/` to build the peer.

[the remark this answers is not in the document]
mbrandenburger: With ercc refactoring and removing ercc-vscc this build dep will go away.

README.md
```diff
@@ -376,9 +376,6 @@ enclave and [tlcc/](tlcc) for the ledger enclave.
 In order to run and deploy a chaincode enclave we need to build the enclave
 registry. See [ercc/](ercc).
 
-Both, ecc and ercc use custom validation plugins to verify transactions. In [plugins/](plugins) you can find
```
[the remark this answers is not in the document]
mbrandenburger: well, this is the building instructions, as we don't have a build target for the plugins anymore, I thought we could remove it. 

I brought this section back and linked to fabric folder

validation/fpc/fpc_validation_logic.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// ...and the transaction...
+	tx, err := protoutil.UnmarshalTransaction(payl.Data)
+	if err != nil {
+		logger.Errorf("VSCC error: GetTransaction failed, err %s", err)
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
mbrandenburger: I see that the comment read nicely before. Still, I would to keep the order like that as it follows the the implementation of the default validator. https://github.com/hyperledger/fabric/blob/master/core/handlers/validation/builtin/v20/validation_logic.go#L109

validation/fpc/fpc_validation_logic.go
mbrandenburger: this check we inherited from also from the dafaultValidator https://github.com/hyperledger/fabric/blob/master/core/handlers/validation/builtin/v20/validation_logic.go#L109

mbrandenburger:
@g2flyer Thanks for the comments, I added back the comment in `README.md`. For other feedback see inline review comments. I squashed and ready to merge (once travis made it through).

Makefile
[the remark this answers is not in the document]
g2flyer: key phrase, though, "wll", i.e., "did not yet" ;-)

validation/fpc/fpc_validation_logic.go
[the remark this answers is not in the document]
g2flyer: Actually, given that you kept it closely (identical) to that code, would have been good to add a comment with the provenance of your code.  I think this is good practice and we definitely will have to do that for tlcc ... (i'm trying to do that for all the config files and docker-compose stuff where applicable: it really helps when doing the upgrades ...)

PR #399, opened 2020-07-08
integrate ecc-vscc in fabric peer

bvavala:
This PR is concurrent with #398 and has the same objective and fewer changes.
The expectation is that only one of these (possibly integrating changes of the other one) is eventually merged.

### DIFF
--- a/fabric/0001-Integrate-FPC-Validation-Plugin.patch
+++ b/fabric/0001-Integrate-FPC-Validation-Plugin.patch
@@ -1,0 +1,33 @@
+From a4b40dd866b83b0ff9489bde057a1a68a69ed840 Mon Sep 17 00:00:00 2001
+From: Bruno Vavala <bruno.vavala@intel.com>
+Date: Tue, 7 Jul 2020 23:14:10 -0700
+Subject: [PATCH] Integrate FPC Validation Plugin
+
+Signed-off-by: Bruno Vavala <bruno.vavala@intel.com>
+---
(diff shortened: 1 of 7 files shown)


g2flyer:
> This PR is concurrent with #398 and has the same objective and fewer changes.

I think the other PR is primarily larger as Marcus didn't do a git mv when relocating the plugins to a separate validation folder and also made it go build-tag'able.   The latter i think we need and refactoring the code out of plugins would also make sense, but ideally with a `git mv` so change is smaller and history is still there.   You did, though, catch that there are two core.yamls to fix, not only one.  Did i miss any other more substantial difference?

bvavala:
> > This PR is concurrent with #398 and has the same objective and fewer changes.
> 
> I think the other PR is primarily larger as Marcus didn't do a git mv when relocating the plugins to a separate validation folder and also made it go build-tag'able. The latter i think we need and refactoring the code out of plugins would also make sense, but ideally with a `git mv` so change is smaller and history is still there. You did, though, catch that there are two core.yamls to fix, not only one. Did i miss any other more substantial difference?

No, nothing substantial.
A curious thing is that #398 still compiles `ecc-vscc` as a Go plugin and also uses it in docker compose. The `fpc-core.yaml` does the trick of mapping the name to the externally (rather than internally) built one.

mbrandenburger:
Thanks Bruno, I close this one and we continue with #398. My new revision integrates the parts that I missed but you discovered.

Issue #133, opened 2019-08-07
Add custom data (e.g. ECC_MRENCLAVE) to chaincode definition


mbrandenburger:
With new lifecycle v2 each organization must approve a so-called chaincode definition before a chaincode is "instantiated".

As this approval mechanism is great to also agree on the enclave code, in particular ECC_MRENCLAVE, we should leverage this here. Currently, we use a hard-coded ECC_MRENCLAVE and write it to the ledger using during chaincode instantiate (just for simplicity).

This task builds the foundation for the new ERCC model.

The purpose of this task is to identify ways how to extend or add custom data to a chaincode definition. Maybe existing fields can be used ("hijacked").

Last, implement this.

mbrandenburger:
The latest design of FPC uses the concept of the chaincode definition to let the parties to agree on the MRENCLAVE of a FPC chaincode. See https://github.com/hyperledger-labs/fabric-private-chaincode/blob/cr2.0.0/docs/design/fabric-v2%2B/fpc-lifecycle-v2.puml

mbrandenburger:
The chaincode definition can be accessed through lifecycle. See https://github.com/hyperledger-labs/fabric-private-chaincode/issues/361#issuecomment-646610010

g2flyer:
resolved in new design as encoding in version field (MRENCLAVE) and baking in into ERCC (e.g., IAS keys, TLCC MRENCLAVE ..)

PR #413, opened 2020-07-21
Add ChaincodeDefinition utility functions

mbrandenburger:
This PR adds a new set of utility functions to retrieve mrenclave from the chaincode definition.
Additional utility functions to retrieve lifecycle policies and to verify that a given msp identity satisfies this policy.

This PR builds the foundation for  #133


ercc/ercc.go
```diff
@@ -156,6 +157,22 @@ func (ercc *EnclaveRegistryCC) registerEnclave(stub shim.ChaincodeStubInterface,
 	// set enclave public key in attestation report
 	attestationReport.EnclavePk = enclavePkAsBytes
 
+	// second, check that attestation report contains expectedMrEnclave as defined in chaincode definition
+	// TODO set chaincode name
+	expectedMrEnclave, err := utils.ExtractMrEnclaveFromChaincodeDefinition("chaincode  name", stub)
```
mbrandenburger, line 162: I expect this to fail. however, on my local tests it doesn't. I hope travis catches this.

fabric/patches/0001-Enable-mod-based-FPC-enabled-peer-builds.patch
```diff
@@ -7,14 +7,14 @@ Subject: [PATCH] Enable mod-based FPC-enabled peer builds
 
 Signed-off-by: michael steiner <michael.steiner@intel.com>
 ---
- go.mod   | 10 ++++++++++
```
g2flyer, line 10: go.sum didn't change?

internal/utils/chaincode.go
```diff
@@ -0,0 +1,43 @@
...
+	resp := stub.InvokeChaincode("_lifecycle", [][]byte{[]byte(function), argsBytes}, stub.GetChannelID())
+
+	df := &lifecycle.QueryApprovedChaincodeDefinitionResult{}
+	if err := proto.Unmarshal(resp.Payload, df); err != nil {
+		return nil, err
+	}
+	return df, nil
+}
+
+func ExtractMrEnclaveFromChaincodeDefinition(chaincodeId string, stub shim.ChaincodeStubInterface) (string, error) {
```
g2flyer, line 35: Probably not for this PR, but i'm wondering whether we do not also need the sequence number. Or do we for now make sure ercc does exactly a single read to lscc for a particular chaincode?  If we do multiple calls, we might get strange (and potentially insecure) behaviour when the chaincode changes? admittedly, the at least lifecycle policy orgs have to sign off a change but probably still worth keeping that in mind? Or do we want to explicitly _not_ support changing chaincode definitions for now (in which case we maybe want to make this explicit in MVP issues?)

g2flyer, line 35: " ExtractMrEnclaveFromChaincodeDefinition" seems a rather verbose name hidding a bit the key part. what about a simple "GetMrEnclave" or "GetRegisteredMrEnclave"?

internal/utils/chaincode.go
```diff
@@ -0,0 +1,43 @@
+/*
```
g2flyer, line 1: More high-level questions: It's not clear to me what distinguishes `internal` from `common`.  What is the (clear) semantic difference between the two?  Otherwise i think we might eventually pick one or the other (as otherwise you have to now always look at two place to figure out whether some exists or not ...)

mbrandenburger: `common` currently only contains c/cpp code, where `internal` only has go code. Also, `internal` is a special folder that prevents external projects from importing any packages from them.

I think we should maintain this separation of c and go code. Maybe find better names?

internal/utils/chaincode.go
mbrandenburger: well, this should not happen. When the ercc transaction is processed, it will work on a ledger snapshot. I hope that this snapshot is also shared with `_lifecycle` and thus concurrent changes will not impact the execution.

I am wondering how we are supplying the sequence number?! Is this an input of ercc? Should ercc always use the most recent sequence number?

internal/utils/chaincode.go
g2flyer: Not immediately clear to me why keeping directory based on language separate, in particular on the top-level and with a name which completely obfuscates that. What do you see as an advantage to keep it separate?  The only thing which could come in mind is issues i encountered with vendoring and cgo (e.g., `go mod vendor` does only copy subsets of code into the vendor directories where right now only c-code residing in same directory of (referenced) go-code and eventually even that might go from various discussion ...) but for that one probably would have to restricture it globally with all go and non-go code in seperate sub-dirs if not even repos. (e.g., c-code would be in separate library and would have to be built _and installed_ before building go and cgo references.

internal/utils/chaincode.go
g2flyer: > well, this should not happen. When the ercc transaction is processed, it will work on a ledger snapshot.

My concern was consistency _across_ transactions, e.g., enclave1 has a transaction referencing lscc, chaincode def changes, enclave2 has a transaction references lssc (and doing so without realizing that cc def has changed ...)

> I hope that this snapshot is also shared with `_lifecycle` and thus concurrent changes will not impact the execution.
> 
> I am wondering how we are supplying the sequence number?! Is this an input of ercc? Should ercc always use the most recent sequence number?

as you register a chaincode, ercc could retrieve (current) sequence number from chaincode def and ensure during any ercc update tx  that it is still same sequence number?

ercc/ercc.go
g2flyer: well, travis also succeeded. So i guess something must be wrong ...

internal/utils/chaincode.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	df := &lifecycle.QueryApprovedChaincodeDefinitionResult{}
+	if err := proto.Unmarshal(resp.Payload, df); err != nil {
+		return nil, err
+	}
+	return df, nil
+}
+
+func ExtractMrEnclaveFromChaincodeDefinition(ccDef *lifecycle.QueryApprovedChaincodeDefinitionResult) (string, error) {
```
g2flyer, line 57: does this really have to be a public function given that we have GetMrEnclave?

ercc/ercc.go
mbrandenburger: The tests did not fail because `ercc.ra.CheckMrEnclave(expectedMrEnclave, attestationReport)` always returns true in simulation mode. :) so whatever is read from the chaincode definition (or not) is just ignored by the actual check. Anyway, I commented out all the useful checks in ercc but keep it there to help Bruno to integrate this during ercc refactoring.

internal/utils/chaincode.go
mbrandenburger: yes, my concern is mostly related to cgo. This is also the reason why I initially introduced `ecc` and `ecc_enclave`.

integration/config/core.yaml
```diff
@@ -570,7 +570,7 @@ chaincode:
         # FPC Addition 0: external builder for fpc-c chaincode
         - path: ../../fabric/externalBuilder
           name: fpc-c
-          propagateEnvironmentp:
+          propagateEnvironment:
```
mbrandenburger, line 573: this typo was really nasty.

mbrandenburger: it took me an hour to find the reason why ercc does log anything. :/

mbrandenburger: I blame myself for being a bad reviewer :)

g2flyer: aargh, sorry for the typo. It worked in the past but i had to rename it due to fabric 2.2 having renamed the vars in a more PC manner and i must have accidentally added a p (annoyingly, it was only in the actually used one but not in commented out version, so most have been added post the global copy/replace ...). At least i did it consistently across the two core.yamls ;-)

internal/utils/endorsement.go
```diff
@@ -0,0 +1,42 @@
...
+	pp := cauthdsl.NewPolicyProvider(id.IdentityDeserializer)
+	policy, _, err := pp.NewPolicy(policyBytes)
+	if err != nil {
+		return err
+	}
+
+	return policy.EvaluateIdentities([]msp.Identity{identity})
+}
+
+func CanEndorse(identityBytes []byte, ccDef *lifecycle.QueryApprovedChaincodeDefinitionResult) error {
```
g2flyer, line 39: This seems a slightly different version of canEndorse from the PUMLs.  In that case we do it on the PK, not the (peer) identity.  

I realize, though, also that the canEndorse in the PUML is not completely consistent in the PUMLS. There is actually a slighly different case for the use in ercc during registration and ecc during key dist.  In the former (and contrary to what the comment says) we will call it _before_ registration in ercc, so (but steps before canEndorse ensure it is of proper verified mrenclave etc) whereas in the ecc case it the registration check is crucial as it transitively ensures validation of mrenclave etc.   Given that you currently use it for ercc your current implementation i guess is the only thing you need. That said, maybe call it slightly differently to not confuse with the (semantically different) `can_endorse` from the puml?

internal/utils/chaincode.go
mbrandenburger: removed

internal/utils/endorsement.go
mbrandenburger: you are right, we better rename this function and let ercc implement `canEndorse` function as described in the PUMLs. Here we should only focus on the primitives that are needed to implement `canEndorse`. I guess this function here is more useful Org-Enclave binding.

mbrandenburger: renamed.

internal/utils/endorsement.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+
+	"github.com/hyperledger/fabric-protos-go/peer/lifecycle"
+	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/common/cauthdsl"
+	"github.com/hyperledger/fabric/msp"
+	"github.com/hyperledger/fabric/msp/mgmt"
+)
+
+func NewPolicyEvaluator() *PolicyEvaluator {
```
g2flyer, line 13: Do we need PolicyEvaluator in its function public given that we have IsValidEndorserIdentity?

g2flyer:
Oops, approve button click was actually a bit premature as my test in HW mode was still running and I also wanted to add some comments to my approve:
- i still think we should rethink the `internal` vs `common` as i think it is confusing.  That said that seems better done outside of this PR as both directories already existed before this PR
- thanks for finding and fixing the annoying typo in core*.yaml

g2flyer:
> ...as my test in HW mode was still running 

ok, test finished and everything still works nicely in HW mode ...

internal/utils/endorsement.go
```diff
@@ -0,0 +1,42 @@
...
+	pp := cauthdsl.NewPolicyProvider(id.IdentityDeserializer)
+	policy, _, err := pp.NewPolicy(policyBytes)
+	if err != nil {
+		return err
+	}
+
+	return policy.EvaluateIdentities([]msp.Identity{identity})
+}
+
+func IsValidEndorserIdentity(identityBytes []byte, ccDef *lifecycle.QueryApprovedChaincodeDefinitionResult) error {
```
bvavala, line 39: This function is defined but only used in tests. Anyway, it should be helpful once the enclave's identity/key is handled through MSP.

internal/utils/chaincode.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/hyperledger/fabric/protoutil"
+)
+
+const MrEnclaveLength = 32
+
+func GetChaincodeDefinition(chaincodeId string, stub shim.ChaincodeStubInterface) (*lifecycle.QueryApprovedChaincodeDefinitionResult, error) {
+	function := "QueryApprovedChaincodeDefinition"
+	args := &lifecycle.QueryApprovedChaincodeDefinitionArgs{
+		Name:     chaincodeId,
+		Sequence: 0, // TODO get current sequence number
```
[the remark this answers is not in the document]
mbrandenburger: good catch! this needs to be updated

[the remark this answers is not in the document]
mbrandenburger: just double checked with fabric docu. see https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerlifecycle.html#peer-lifecycle-chaincode-querycommitted

internal/utils/endorsement.go
mbrandenburger: true, but before we introduce msp for the enclave, we can use this method to check whether a peer that hosts a chaincode enclave is valid endorsing peer (this only makes sense once we have some form of org-enclave binding).

g2flyer: it is required to implement `can_endorse`?

internal/utils/chaincode.go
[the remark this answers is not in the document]
g2flyer: @mbrandenburger @bvavala I'm a bit confused here: Bruno raises two good points but the merged seems to address only the first one but not the second. Looking at the link marcus sent i don't see how how that would address it. Was that the wrong link and you want to point to peer lifecycle chaincode queryapproved? But that link seems rather to confirm that Bruno has said . So with the current merged code still calling QueryApprovedChaincodeDefinition i am not sure this is really what we want. On the one hand it would seem to mean that retrieval might fail for peers which didn't approve even if the chaincode was committed but on the other hand it also could return something which is not committed?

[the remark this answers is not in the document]
bvavala: Here is the [QueryChaincodeDefinition](https://github.com/hyperledger/fabric/blob/17f5c7fb020a9228af8791e484783f6a8c5b6da2/core/chaincode/lifecycle/lifecycle.go#L622).

Michael, you're right. Marcus has correctly updated the protobufs, and the update of the most important line went  unnoticed. Surprisingly, the protobufs are similar enough that they do not trigger a type-check error.
I am going to fix it right away.

PR #415, opened 2020-07-25
fix get chaincode definition

bvavala:
This PR completes the fixes started and missed in #413 .


bvavala:
> Looks good! Only thing we might also change in `chaincode_test.go` is the two `When` clauses which says "approved" but probably should say "committed"?

Right. I double-checked with a _case-insensitive_ grep, this should complete the modifications.

g2flyer:
> Right. I double-checked with a _case-insensitive_ grep, this should complete the modifications.

That's actually how i found these references :-)

Issue #412, opened 2020-07-20
Design common attestation / crypto API


mbrandenburger:
**Description**

This work items defines a common attestation / crypto API that can be used in various components. The main goal of this task is to reduce redundant code and provide a single implementation usable from c/cc+ and go-based components. 

The implementation is based on the PDO crypto library which we have already linked in the project with PR pdo-crypto-lib integration #337

The main focus lies on the c/c++ interface as exposed to the FPC components and implemented by the PDO crypto lib. Ideally, we have this abstraction in case we need to change the underlying implementation and remove the PDO crypto dependency. Additionally, a go interface is designed and its implementation uses cgo to call the corresponding c/c++ interface.

Required functionality:

- create and verify signatures
- encryption and decryption (symmetric/asymmetric)
- SGX attestation evidence verification (support for EPID and DCAP)

TODOs:

- [x] Define Attestation API
- [x] Define Crypto API
- [X] Integrate PDO crypto Lib to implement the new interfaces
- [x] Implement go-based API using cgo
- [x] Create test cases
- [x] Integrate new attestation / crypto API in TLCC
- [x] Integrate attestation / crypto API in FPC_Enclave
- [x] Integrate attestation / crypto API in ERCC

Some ground work has been done already:

- pdo-crypto-lib integration #337
- port pdo crypto tests #367

Other related issues:

- Attestation mechanism: EPID #278
- Generalize attestation support #15
- Publicly verifiable attestation #114

g2flyer:
a few thoughts on the APIs and for which language:

- do we really need a cgo interface for crypto?   From what i can see we need crypto as part of processing of client request and response. However, i wonder whether for that case it wouldn't be better to have higher-level C APIs along the lines of `WrapRequest(..)`, `UnwrapRequest(...)`, WrapResponse(...)`, `UnwrapResponse` which we then can wrap with Cgo (and potentially, with swig for Node SDK).   

- In general, i think it would be a good idea to implement "sibling" functions like above `WrapRequest(..)` and  `UnwrapRequest(...)` in the same language and same library as it makes it easiest to keep stuff consistent and re-using code.

[...]

Issue #402, opened 2020-07-10
Ledger Enclave Strategy 


mbrandenburger:
The integration of the Ledger Enclave is one of the fundamental pillars in the FPC security architecture, however, the implementation is also challenging.

This issues continues the discussion based on the [FPC RFC](https://github.com/hyperledger/fabric-rfcs/pull/21).

- [ ] Define minimal set of functionality implemented by Ledger Enclave

- [ ] Develop short-/mid-/long-term working plan for development.

yacovm:
I think that instead of mimicking Fabric semantics in a complete manner, the strategy should be really deviating from Fabric by:
1.  Restricting functionality of what you can do
2. Further isolating FPC transactions from Fabric ones

By doing (1), validation logic is reduced and most importantly, doesn't need to play catch-up with Fabric changes.
In a sense, (2) is an indirect result of (1): Instead of validation all transactions of Fabric (and implement everything needed for that), maybe it's better to just make Fabric transactions and FPC transactions not be able to read each other's writes. 
The latter also means that in a manner of speaking, we are creating two "realms" inside the peer that prevents transactions to cross over. Surely, this would make several use cases infeasible, but my gut feeling is that 80% of the effort in support all Fabric validation logic would amount to catering to less than 20% of the use cases.

To achieve (1), there are several relaxations (which are, semantic restrictions) that can be taken into account: 
- Entirely remove the notion of "endorsement policy" by moving the trust to the endorsement step (since SGX is modeled as an honest party) and ensure that FPC instances only endorse transactions they are eligible of endorsing, and that it is enough to receive a single signature from an FPC instance. 
- For every namespace, restrict to a single autonomously monotonously increasing version number and at validation time always track the maximum version number and invalidate transactions that don't have this number. This can be done by adding into the read set of the transaction, the version number of the chaincode. 

By doing the above, we are essentially decoupling the FPC transaction validation from the Fabric configuration block processing and the Fabric transaction lifecycle and greatly simplifying the code.

To realize (2) I think the best way would be to introduce to Fabric a new transaction type and have a custom transaction processor for that transaction. While it does modify Fabric, I believe it is the right approach in the long term because it doesn't require changing any of the existing Fabric endorsement transaction logic which is already complex, and also allows flexibility and more freedom to the FPC transaction validation logic, as it is no longer bound by the format and structure of the endorsement transaction. 

[...]

PR #420, opened 2020-07-28
Api design

### LINKED ISSUE #410: Design TLCC API
**Description**

This work item summaries the new TLCC API exposed to a chaincode enclave based on the recent design diagrams.

**Link to feature or bug**

- Secure get_creator implementation #357
- Chaincode enclave / tlcc binding. #14
- Do TLCC cc2scc calls require reflection in r/w set of transactions? #361 

**Link to dependent issues**

- Enclave creation and registration #404

mbrandenburger:

This PR proposes the FPC component API and state design based on the most recent UML diagrams. The goal is to stablize the APIs and internal state before current code base can be refactored to align with the FPC design as specified by the UML diagrams.

Addresses #410 but excludes chaincode enclave / tlcc binding as this is covered by #434 


mbrandenburger:
@bvavala @g2flyer Can you please have a quick look at this? It's still incomplete; happy for initial thoughts. Thanks

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+}
+```
+
+
+# TLCC_Enclave
+
+## Interface:
+Defined in the TLCC EDL.
+```c++
+// TBD is this interface already sufficient to support multi-channel?
```
g2flyer, line 143: hmm, don't think so. To some extent it depends on how you define key (which could be defined as the concatenation of `channel-id || chaincode-id || chaincode-key`) but i think having the triplet explicit is nicer/easier to deal with and also easier to understand.

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+public int ecall_join_channel(
+        [in, size=gen_len] uint8_t *genesis, uint32_t gen_len);
+
+// pushes next block to tlcc
+public int ecall_next_block(
+        [user_check] uint8_t *block_bytes, uint32_t block_size);
+
+// verify state data
+public int ecall_get_state_metadata(
+        [in, string] const char *key, // TBD: should we add namespace separetely?
```
g2flyer, line 156: I think adding the namespace is useful. Certainly would allow ecc to access ercc state meta-data and might safe us a separate function which we currently have in the UML to validate ercc registration state.  And as mentioned above, might be good to already make the channel also explicit?  

BTW: in join the channel-id is implicitly the genesis block but in other cases/externally a channel-id is just a string: how do we do that mapping and is there any issue of having the string or would we need a cryptographic secure identifier (e.g., hash of genesis block)?  My intuition is that name probably should be ok (or fabric already would have some problem), in which case maybe the easiest is to just add a name parameter to the `join_channel`? [PS: seeing only now your `TODO` above, maybe this mapping channel data to channel name was maybe exactly the motivation for that?]

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+```
+
+## State:
+The internal TLCC state consists of the current ledger state (block number, integrity-metadata, and msp) and sessions information for ecc communication.
+
+```c++
+// TODO: this needs some more brain cycles with respect of #402 and #410
+typedef struct internal_tlcc_state_t {
+        const char* INTEL_PUB_PEM, // Intel IAS verification key
+
```
g2flyer, line 180: This should probably be outsourced to the attestation module?  Practically, i thought this also would really be some static variable/constant defined in that code and so wouldn't really need any runtime setting, other than the other data you have here. So keeping that separate seems anyway kindof nice ...

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+        [in, string] const char *comp_key,
+        [in, size=32] uint8_t *nonce,
+        [out] sgx_cmac_128bit_tag_t *cmac);
+
+// verify msp identities
+public int ecall_validate_msp_id(
+        [in, string] const char *msp_id,
+        [in, size=32] uint8_t *nonce,
+        [out] bool isValid, 
+        [out] sgx_cmac_128bit_tag_t *cmac);
```
g2flyer, line 170: Wouldn't this correspond to the `can_endorse` from the UML diagrams? So why not use the same name?  The `can_endorse` seems also to capture more closely what we validate?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+```c++
+// TODO: this needs some more brain cycles with respect of #402 and #410
+typedef struct internal_tlcc_state_t {
+        const char* INTEL_PUB_PEM, // Intel IAS verification key
+        
+        uint32_t sequence_number, // block number counter
+
+        // integrity metadata
+        kvs_t metadata_state,
```
g2flyer, line 184: this would capture fpc, ercc and lscc state? Maybe better to separate (as you already have done for MSP related info)?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+// ECC must trigger this when sealedState is available using load() and ecc starts; this is not really a public interface
+createEnclaveWithSealedState(sealedState []byte) error
+
+// Storage interface
+// TODO how to store sealedState, where and under what identifier (e.g. chaincode_id);
+load(chaincode_id string) (sealedState []byte, err error)
+store(chaincode_id string, sealedState []byte) error
+```
+
+This interface is implemented by ECC to let a chaincode enclave call into the peer
```
g2flyer, line 232: I think this is ideally outsourced into a separate sealedstorage module where you just have store and retrieve function exposed to ecc_enclave.  (with the enclave having some state-machine { no-keys, enclave-keys, cc-keys} and would make sure that certain function can be called only in a certain state and if not yet in that state, will try first to retrieve the state from sealed storage?)

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+        [in, out] sgx_cmac_128bit_tag_t *cmac,
+        [user_check] void *u_shim_ctx);
+
+
+// verify msp with the help of tlcc
+void ocall_verify_msp_id(
+        [in, string] const char *msp_id,
+        [in, size=32] uint8_t *nonce,
+        [out] bool isValid, 
+        [out] sgx_cmac_128bit_tag_t *cmac);
```
g2flyer, line 265: This would be exactly the same EDL as TLCC has? (at least i see no reason to have two different .edls ...)  If so, maybe just reference here tlcc rather than redefining?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+
+// attestation
+void ocall_get_spid(); // TBD: is spid only needed for EPID?
+
+// creates an attestation evidence; the evidence_request includes all the form of evidence (EPID/DCAP), the quote, and additional information required to complete the evidence request
+void ocall_create_evidence(
+        [in] evidence_request
+        [out] evidence
+);
```
g2flyer, line 275: As mentioned above, i think we should delegate these ocalls/ecalls to the implementation of  #412 and just rely on the (local call) API (to be) defined by #412.

To some extent you already assume the existance of such a module by the `attestation.EvidenceService` in your state?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+Note: `sealed credentials` is renamed to `sealed state`
+
+```c++
+// creates an new enclave
+public int ecall_create_enclave(
+        [in, size=cc_params_len] uint8_t *cc_params, uint32_t cc_params_len,
+        [out] uint8_t *credentials,     // TBD is this a fixed length type?
+        [out, size=max_sealed_state_len] uint8_t *sealed_state, uint32_t max_sealed_state_len,
+        [out] uint32_t *sealed_data_len);
```
g2flyer, line 309: It's a matter of taste but i think it would be nice to have a separate library/module for sealed state-management and don't put it here in these ecalls.  The nice part i think would be that we can start without sealed state and when adding sealed state none of the main ecall/ocalls here change but there will be this additional library with it's own edl and decisions like having separate write-once sealed blobs vs having single updateable blobs  can be all defined separatedly?
Admittedly, though, i guess the current UML seem to specify it this way, so we might have to fix it. But there seemt to be other small consistency issues with the diagrams which we might have to fix anyway at some point?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+// creates ("restores") an enclave with sealed state
+public int ecall_create_enclave_with_sealed_state(
+        [in, size=sealed_state_len] uint8_t *sealed_state, uint32_t sealed_state_len);
+
```
g2flyer, line 314: If we have a separate module, then this function should be internal to the module (and actually might be an ocall fetching state iff the module is "driven" from within the enclave; of course the library will always need code in enclave and outside, but only one side probably needs API to the rest of the code ...)

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+// TBD what is the input/output here?
+public int ecall_get_CSR(
+        [in]    
+        [out] csr);
```
g2flyer, line 318: I would ignore any CSR stuff and related validation for now in this context.  Certainly will not be MVP and very likely also after single endorsement peer and other road map features.  Arguably, this might eventually even be hidden by the attestation module?

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+typedef struct cc_params_t {
+        cc_id,          // chaincode id
+        version,        // chaincode version, this is the expected mrenclave
+        msp_id,         // msp_id that does what?
+        channel_id      // channel id
+        tlcc_mrenclave, // expected tlcc mrenclave
+};
+
+// this is create on enclave creation and kept inside the enclave
+typedef struct internal_ecc_state_t {
```
g2flyer, line 364: Maybe worth from to separate the sealed vs ephemeral state?  E.g., session key definitely would not be sealed.  CC_param certainly has to be bound to sealed state but it could actually be beneficial to bind it via authenticated data as the untrusted state anyway needs mapping of sealed blob to a subset of CC_Param?)   That said, might be driven also by the API of the seal/unseal library (if we go a modular route)?

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+```go
+// chaincode interface (exposed to ecc and clients)
+func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
+
+// functions below are dispatched by Invoke implementation
+
+// returns a set of credentials registered for a given chaincode id
+// TBD: adding the peer_id to the credentials would allow the clients to pick the set of FPC endorsing peers
+func queryCCInfos(chaincode_id string) (chaincode_ek []byte, allCredentials []Credentials) {}
+func queryCCInfo(chaincode_id string, enclave_id string) (ehaincode_ek []byte, credentials Credentials) {}
```
g2flyer, line 20:
```suggestion
func queryCCInfo(chaincode_id string, enclave_id string) (chaincode_ek []byte, credentials Credentials) {}
```

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+}
+
+message SessionClose {
+        byte context = 1;
+}
+
+message SessionCloseResponse {
+}
+
+message SessionRequest {
```
g2flyer, line 210: If we do on session for whole lifetime of ecc, we also need here a transaction identifier so we can make sure all state returned for a single transactions are all based on the same ledger state

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+        [in, string] const char *comp_key,
+        [in, size=32] uint8_t *nonce,
+        [out] sgx_cmac_128bit_tag_t *cmac);
+
+// verify msp identities
+public int validate_msp_id(
+        [in, string] const char *msp_id,
+        [in, size=32] uint8_t *nonce,
+        [out] bool isValid, 
+        [out] sgx_cmac_128bit_tag_t *cmac);
```
g2flyer, line 291: I guess this cmac and the cmac below would/should really be done by the session handling by ensuring request and response authenticity?  Similar all nonces can/should be done a single place as part of session handling?

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+// addtionally we may use the state sealing module to persists TLCC state
+// see below the data sealing module.
+include sealed_storage.edl
+```
+
+Session requests are depatched to the following interfaces:
+```c++
+// TODO refine
+// verify state data
+public int get_state_metadata(
```
g2flyer, line 274: Depending on how we handle state updates and consistency of requests inside a single transaction, we might have to surface some additional param here?

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+public int ecall_export_cc_keys(
+       [in] uint8_t *target_enclave_vk,
+       [out] uint8_t *export_msg);
+
+public int ecall_import_cc_keys(
+        // TBD: Why do we need cc_params?
+        [in, size=cc_params_len] uint8_t *cc_params, uint32_t cc_params_len, 
+        [in] uint8_t *export_msg,
+        [out] uint8_t *cckey_registration_msg,
+        [out] uint8_t *sealed_cckeys);
```
g2flyer, line 488: if we have a sealed storage module, then there would also not be any return here? (similar to `init`)?

docs/interfaces.md
g2flyer: also, might be worth adding explicit an `initialization_state` enum or alike which with potential states `uninitialized`, 'enclave_keys_exists`, `cc_key_exists` (after which probably invoke should work? Or would we also need a `committed` state or alike?)

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+typedef struct internal_tlcc_state_t {
+        uint32_t sequence_number, // block number counter
+
+        // integrity metadata
+        kvs_t metadata_state,
+        
+        // keep ercc and lifecycle separetely as TLCC provides
+        // additionality, such as `can_endorse`, requiring the information
+        kvs_t ercc,
+        kvs_t lifecycle
```
g2flyer, line 315: does lifecycle also cover the default policies? (TLCC would need the lifecycle policy to validate ERCC cc-def changes ...)

docs/interfaces.md
```diff
@@ -0,0 +1,571 @@
...
+
+```
+
+```c++
+// verify msp with the help of tlcc
+void ocall_tlcc_request();
+```
+
+```c++
+// Provided/implemented by a common sealed storage module
```
g2flyer, line 407: We currently have one. If created a corresponding issue #421 for that, you probably should reference that and eventually put details on internals there?

docs/interfaces.md
g2flyer: you added can_endorse. in that case, do we still need validate_msp_id? what exactly would it do?

g2flyer: What we _do_ need though additionally is something which helps in implementing `get_creator`.  Maybe easiest is a function which passes a SignedRequest and returns a (validated) tuple <msp-id, x509-principal/cert>`?

docs/interfaces.md
g2flyer: Probably worth also mentioning that this namespace-enhanced api makes obsolete the separate `check_ercc_commited_data` function we had in the `fpc-key-dist` UML diagram.

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Interface:
+The ERCC interface is implemented through a "normal" chaincode interface. All functions can be called by admin or via cc2cc.
+
+```go
+// chaincode interface (exposed to ecc and clients) implemented by invoke
+func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
+
+// returns a set of credentials registered for a given chaincode id
+queryCCInfos(chaincode_id string) (Chaincode_EK []byte, allCredentials []Credentials)
```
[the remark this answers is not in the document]
mbrandenburger: So what is the endorsing peer discovery protocol? I would imaging the following:

1. Client invokes `queryCCInfos`, and gets back the chaincode_ek and a list of enclave credentials. This list represents all possible endorsing enclaves/peers.

1. Client picks one (or more) enclaves using the information he got from credentials.

1. Last, clients prepare transaction proposal and submits it to the endorsing peers/enclaves.

With this protocol in mind, we would need to add some peer identifier to the enclave credentials, that allows the client to retrieve the peer endpoint. Also, since the credentials are attested, we have a nice way to show that an enclave is owned by a peer.

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+To store date we use the following key scheme. A variable is denoted using <variable_name> annotation.
+
+```go
+// returns chaincode encryption key
+namespaces/chaincode_ek/<namespace> -> chaincode_ek
+
+// returns all credentials for a given chaincode
+namespaces/credentials/<namespace> -> []Credentials
+namespaces/credentials/<namespace>/<enclave_vk> -> Credentials // see credentials definition below in ecc
```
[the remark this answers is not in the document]
mbrandenburger: changed to `enclave_id` and defined string encoding for that

docs/interfaces.md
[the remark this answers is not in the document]
mbrandenburger: A added `bytes peer_msp_id` to the credential message

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+type ExportMessage struct {
+        Chaincode_EK []byte,    // public chaincode encryption key
+        CCKeys_enc []byte,      // chaincode keys encrypted for the receiver
+        Enclave2_VK []byte,     // creator of this message
+        Enclave_VK []byte,      // receiver of this message
+        Sig_Enclave []byte,     // signature of the message creator
+}
+```
+
+ERCC keeps in instance of an attestation.Verifier to verify the attestation evidence. In case of EPID, the evidence is an `IASAttestationReport` object. This instance has the Intel attestation verification key.
```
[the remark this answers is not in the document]
mbrandenburger: added

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+// Intel IAS verification key (same as used by TLCC Enclave)
+const IntelPubPEM = `...`
+
+type IASAttestationReport struct {
+	EnclavePk                   []byte `json:"EnclavePk"`
+	IASReportSignature          string `json:"IASReport-Signature"`
+	IASReportSigningCertificate string `json:"IASReport-Signing-Certificate"`
+	IASReportBody               []byte `json:"IASResponseBody"`
+}
+```
```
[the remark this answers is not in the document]
mbrandenburger: Added reference to #412. No DCAP requirement in the near future so far. But with this design we can easily add DCAP support.

docs/interfaces.md
```diff
@@ -0,0 +1,431 @@
...
+
+// used by the admin to join the channel
+joinChannel(channel_id string) error
+```
+
+```c++
+// interface exposed to TLCC_enclave via cgo
+// logging
+void ocall_print_string([in, string] const char *str);
+```
```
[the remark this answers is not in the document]
mbrandenburger: added a note

docs/interfaces.md
mbrandenburger: added namespace. as a single tcc_enclave is responsible for a single channel, the dispatching using the channel id is done already outside tlcc_enclave, thus, no need to add it tlcc.edl

docs/interfaces.md
mbrandenburger: namespaces added. I also added a note saying that we have a tlcc_enclave for each channel, that is, there is no need to make the channel_id explicit when passing a block to tlcc_enclave. the `join_channel` provided by `tlcc` indeed requires the chaincode_id as known by the peer to start the block listener. 

Maybe tlcc_enclave should get a method to query the root-of-rust, that, is the cryptographic hash of the genesis block it has received from the pee. Ideally this statement is attested?

docs/interfaces.md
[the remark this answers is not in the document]
g2flyer: I forgot that for designated endorsing peer (rather than single peer), the client should know the correct peer unless she wants to broadcast the request (which works given the CC_EK).  That said, i don't really like the idea of adding msp_id to CC_Params for a number of reasons. (1) it is very designated endorsing peer specific and so i think an idea is better associated with the enclave key rather then CC_Pararms . (2) if you restrict it to MSP (rather than some form of peer id), it already exists in the endorsement policy of the cc, so really no need to replicate info?

So to me before changing anything in CC_Params and/or adding more discovery info to ERCC i think we should first answer the following question:  Is there a (org-independent/global) peer identity which (a) we can easily use to direct transactions, (b) is immutable and (c) can easily be retrieved either directly as part of create_enclave flows or by client to be additional input to create_enclave.

If there is none, we anyway have to broadcast so potentially you do need the enclave key as a client, only the cc_key. And if you want to narrow to orgs, you already can do that based on endorsement policy, again no need for additional discovery service from ercc?

docs/interfaces.md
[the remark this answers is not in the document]
g2flyer: from what i understood you, i thought we simply do not know yet? (And then whether it is a requirement would depend on importance of IBM Cloud and which HW they have. I think some HW might actually require DCAP as EPID might not be enabled anymore (EPID is primarily targeted at client systems.)  In any case, i guess we see the requirement when we see it, but if there are easy ways to already loosely couple attestation type, the easier and quicker will we be able to accommedate any popping up DCAP requirement.

docs/interfaces.md
g2flyer: did we decide at some point that multi-channel means multi-enclave?  I guess it is now somewhat implicit in TLCC state. Maybe then worth also explicitly adding the channel id to the tlcc_enclave state?

docs/interfaces.md
g2flyer: if we pass _only_ channel name, how do we ensure though that it really matches the genesis block seen by tlcc_enclave?  Couldn't a malicious org hosting the designated peer create a shadow ledger for tlcc which mimics everything on the real except being rooted in bogus orderer/msp root one and where the malicous org could inject bogus requests to reveal data when useful? somewhere/somehow we have to make sure  that untrusted and trusted ledger share the same genesis block.  One way could be that the gensis block is attested by TLCC when authenticating to ecc, ecc has genesis block baked in & attested to similar to tlcc-mrenclave which then both would be verified by ercc during registration?

docs/interfaces.md
g2flyer: just in case it was not clear. i think it would be useful to explictly add the state-machine i mentioned above. Yes, as other state it doesn't really have to be fully-specified but something like this will be a crucial state component?

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+// interface to establish/close a secure channel between an
+// ECC_Enclave and a TLCC_enclave.
+func session_setup(setup SessionSetup) (SessionSetupResponse, error) {}
+func session_close(close SessionClose) (SessionCloseReponse, error) {}
+
+// interface to invoke a TLCC request, such as, getMetadata and validateMSPId
+func session_request(request SessionRequest) (SessionRequestResponse, error) {}
+```
+
+## State:
```
g2flyer, line 169: Below is not really "State"?   The state tlcc would have is essentially just a lookup table of channel-id->tlcc enclave?

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+        [in, string] const char *str, 
+        [in, size=data_len] uint8_t *data, uint32_t data_len);
+```
+
+
+Additionally, ECC includes a common attestation module that abstracts
+
+```c++
+// attestation
+void ocall_get_spid(); // TBD: is spid only needed for EPID?
```
bvavala, line 422: I have some doubts on the implementation of this ocall.
In a real deployment, there will likely be one spid per mspid. With such an ocall, the only way (that comes to mind) to retrieve the spid is to leverage the external builder to grab a local spid packaged with the chaincode. So this imposes to package the chaincode in a certain way. Also, if ever the package has to be shared, it might have to be repackaged with a new spid.

For this reason I had originally proposed to send the spid in an Aux_Data field (i.e., a json blob) at create time in the invoke. The advantages are: the caller (i.e., the admin) knows the spid associated to its mspid; the json blob can be repurposed for different attestation types, and in that case it will simply not include the spid; finally, the spid does not ship with the package, hence it is separate from the chaincode and even the peer.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+        sgx_aes_gcm_128bit_key_t sek,      // ledger state encryption key
+        sgx_ec256_public_t chaincode_ek,   // argument encryption key
+        sgx_ec256_private_t chaincode_dk,  // argument decryption key
+
+        // tlcc session
+        sgx_cmac_128bit_key_t session_key; // is set during tlcc-ecc binding
+};
+```
+
+# FPC_SHIM
```
bvavala, line 520: Per the lengthy conversation in, and the small code of, #405 , the `get_mspid` and `get_channel_id` APIs can also be implemented conveniently, thereby expanding the domain of supported chaincodes.

docs/interfaces.md
bvavala: Additionally, the json blob can include the sig_rl, which are currently hard-coded.

docs/interfaces.md
mbrandenburger: good point!

docs/interfaces.md
mbrandenburger: added the channel id and a hash of the genesis block

docs/interfaces.md
mbrandenburger: mh ... passing the the entire proposal to tlcc to get back the creator is a bit too much. I suggest that we call this method `ecall_validate_identity`, and offer this to the chaincode as primitive to validate if the creator is valid or not.

docs/interfaces.md
mbrandenburger: ack! ok added a FSM state to the ecc state. Still need to define possible states.

docs/interfaces.md
mbrandenburger: TLCC would not need `ocall_get_state`? Ok, what is the best practise here? ECC needs the TLCC related to calls to perform TLCC `ecalls`. Having a single TLCC-integrity-metadata edl that contains, both, ecalls and ocalls, and is included by ECC and TLCC, would't the edg8er create stub code for all the functions in both enclaves? That is, we have used stub code in TLCC and ECC? Am I wrong here?

docs/interfaces.md
mbrandenburger: added reference to #412

docs/interfaces.md
mbrandenburger: ack! nonces and some for of message authenticator are part of the session messages in `SessionRequest` and `SessionRequestResponse`

docs/interfaces.md
mbrandenburger: we still want to return `export_msg` as this is part of the protocol and later stored at ercc. Removed all sealed messages as they are stored using the storage module.

docs/interfaces.md
mbrandenburger: added

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+        const char *namespace,
+        const char *key);
+
+public int get_multi_state_metadata(
+        [in, string] const char *namespace,
+        [in, string] const char *comp_key,
+        [out] uint8_t *metadata);
+
+// verify msp identities
+public bool validate_msp_id(
```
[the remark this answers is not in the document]
mbrandenburger: I changed this function to validate_identity, which can be used to implement get_creator and check if the creator is a valid identity with respect to the consortium defined in this channel.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+        [in, string] const char *msp_id);
+
+// verify msp identities
+public bool can_endorse(
```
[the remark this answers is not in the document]
mbrandenburger: right, it indeed needs the `enclave_id` and the `chaincode_id`

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+
+        // tlcc session
+        session_keys_t session_keys,
+};
+
+typedef struct kvs_t {};
+
+typedef struct session_keys_t {};
+
+// TLCC contains some attestation module that is responsable to verify
```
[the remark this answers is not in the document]
mbrandenburger: ack! My fault, I mixed current implementation with the new ercc model. :/ good catch

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+        byte context    = 1;
+        byte tx_context = 2;
+        byte nonce      = 3;
+        oneof request {
+                GetMetadataRequest metadata     = 1; 
+                validateMSPRequest validate_msp = 2;
+        }
+}
+
+message SessionRequeatResponse {
```
g2flyer, line 214:
```suggestion
message SessionRequestResponse {
```

docs/interfaces.md
g2flyer: i guess you refer to comment mentioning "signatures"?  Why signatures? We do have a shared key, so  lets just add a cmac field to both messages (or better, similar to signatures, make a sub-type of what is maced?  BTW: the two request/response types of the oneofs are not defined? But i guess they are a one-to-one mapping of the function signature ...

docs/interfaces.md
mbrandenburger: added

docs/interfaces.md
mbrandenburger: I removed the details of the attestation module used by ECC

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
+
+// functions below are dispatched by Invoke implementation
+// Note: Maybe use protobuf-to-json mapping (https://developers.google.com/protocol-buffers/docs/proto3#json)
+// to map the methods and their arguments into to chaincode
+// invocation form { "Function": "..", "Args": [....]}
+
+// returns a set of credentials registered for a given chaincode id
+func queryCCInfos(chaincode_id string) (chaincode_ek []byte, allCredentials []Credentials) {}
+func queryCCInfo(chaincode_id string, enclave_id string) (chaincode_ek []byte, credentials Credentials) {}
```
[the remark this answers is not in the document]
mbrandenburger: `queryCCInfo` is used by the FPC validation code.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
+# FPC Component Interfaces and States
+
+This document defines the interfaces exposed by the FPC components and their internal state.
+
+
+# ERCC
+
+## Interface:
+The ERCC interface is implemented through a "normal" chaincode interface. All functions can be called by admin or via cc2cc.
```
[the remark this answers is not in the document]
mbrandenburger: removed comment.

docs/interfaces.md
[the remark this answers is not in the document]
mbrandenburger: renamed it to validate_identity, which can be used to implement getCreator

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+
+```go
+// chaincode interface (exposed to admin/clients) implemented by invoke
+func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
+
+// functions below are dispatched by Invoke implementation
+
+
+// triggered by an admin
+func initEnclave(params CC_Parameters) (Credentials, error) {}
```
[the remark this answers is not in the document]
mbrandenburger: yes, this used to be called `createEnclave`. It is still part of the create enclave flow and is meant to be the initial step to provision the enclave with the `cc_params` and `host_params`. However, you are right. this method should not need the `cc_params` as input, but rather, the implementation should collect the cc_params (and host params) and pass it to the enclave.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+
+// key generatation 
+func generateCCKeys(params CC_Parameters) (SignedCCKeyRegistrationMessage, error) {}
```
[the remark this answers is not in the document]
mbrandenburger: ack! I took it from the UMLs. but also doesnt make so much sense.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+
+## Interface:
+The ECC_Enclave interface specifies interface of an FPC chaincode enclave. 
+
+Note: `sealed credentials` is renamed to `sealed state`
+
+```c++
+// initializes an enclave
+public int ecall_init(
+        [in, size=cc_params_len] uint8_t *cc_params, uint32_t cc_params_len,
```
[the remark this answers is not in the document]
mbrandenburger: this is meant to be the way how the enclave is provisioned with the cc_params.

docs/interfaces.md
mbrandenburger: right, edited

docs/interfaces.md
mbrandenburger: ack, edited

docs/interfaces.md
mbrandenburger: i removed the SPID ocall but kept the ocall_create_evidence with a note that more details are in #412 and yes therefore we also should keep the reference of attestation.EvidenceService

docs/interfaces.md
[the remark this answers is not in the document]
mbrandenburger: Ouch! this is a tricky thing. We indeed want the peer identity, which contains the peer msp_id + the peer public key. I added a host_params that contains the peer identity and is loaded with iniEnclave. Still need to figure out how we can pass the peer public key.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+// chaincode interface (exposed to ecc and clients)
+func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
+
+// functions below are dispatched by Invoke implementation
+// Note: Maybe use protobuf-to-json mapping (https://developers.google.com/protocol-buffers/docs/proto3#json)
+// to map the methods and their arguments into to chaincode
+// invocation form { "Function": "..", "Args": [....]}
+
+// returns a set of credentials registered for a given chaincode id
+func queryCCInfos(chaincode_id string) (chaincode_ek []byte, allCredentials []Credentials) {}
```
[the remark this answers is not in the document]
mbrandenburger: not so sure about this. You are proposing a `queryRegisteredEnclaves(chaincode_id string)` that then just returns the public keys. that's fine and would work for MVP. 

I thought `queryCCInfos` is more a general purpose query to get all information available for a certain FPC chaincode. As the credentials now also contains some enclave host identifier, it is also useful to pick a set of endorsing enclaves. At least this is what I had in mind.

To address 2., I suggest that we hide these protos as much as possible. Our FPC client SDK layer still should have no problems dealing with this. In case a client wants this data, we could present it as json format as @g2flyer suggested in this conversation.

mbrandenburger:
@bvavala @g2flyer Thanks for your new feedback! I tried to address everything where I can and commented you your feedback. Please have a look and click the resolve conversation btn where ever possible! THanks

docs/interfaces.md
g2flyer: you addressed my second comment, but not the first: the "State" heading should really be moved down just in front of the newly added map; the protobuf are really part of the interface section?

docs/interfaces.md
g2flyer: did you forget to push something here? Still see signatures (and no Request/Response type, although that probably could be handled just by a comment pointing to 1-1 mapping). BTW: i was referring to the sessionrequest/response, not the session setup (although, even in that case you wouldn't really would use signatures: the closest to signatures would be the DH key-exchange but that would be hidden behind the SDK API (see, e.g., https://github.com/intel/linux-sgx/blob/master/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveMessageExchange.cpp for some use of the API ...)

docs/interfaces.md
g2flyer: there are still some old references to ecall_create_enclave_with_sealed_credentials and other references to sealed state sprinkled around.  If you want to keep some (potential) ocalls/ecalls and reference to edl here, rather than just move to #421, then maybe having a separate section where all of it is consolidated would be clearer (and better manageable/kept consistent)?

docs/interfaces.md
```diff
@@ -0,0 +1,591 @@
...
+// ECC_Enclave and TLCC_Enclave. Once a secure channel is established,
+// Session Requests to TLCC can be issued to invoke TLCC functions such
+// as getMetadata or verifyMSPId.
+
+// TODO How to integrate the the message authenticator?!
+// should we abstract it as signature here even though we may
+// use CMAC once a session is established?
+
+message SessionSetup {
+        string channel_id = 1;
```
g2flyer, line 193: This should also be channel_hash. I think internally in our (enclave) code, we should use only channel_hash and never channel_id, i.e., channel_id only as needed to find channel_hash.  Insofar, if possible i would prefer to have channel_hash in CC_Params (and no reason then to have channel_hash in credentials)?

docs/interfaces.md
g2flyer: i would also remove `ocall_create_evidence`. Whether such a thing exists as ecall or just as a library function should really be defined by #412?

docs/interfaces.md
g2flyer: we definitely want to integrate this check into the shim's `GetCreator`. What we pass to tlcc i guess depends a bit on how much we already have to parse and decompose signed proposal.  your current validate_identity seems though not sufficient: you certainly would not send an identity string, at the very least you would need to send the certificate (from which ecc would have to extract the PK to verify the signature) and the corresponding MSP id it should match?

docs/interfaces.md
g2flyer: what about the states i have above as starters ;-)

docs/interfaces.md
g2flyer: oops, you are correct, the first three of course go to shim (and the fourth is gone)

g2flyer:
arrgh, wrong radio box again: this was not supposed to be approved. Not all of my comments are crucial but a number of them should hopefully, when addressed,help clarity, and i think the channel_id vs channel_hash it might be good to have some quick discussion?

docs/interfaces.md
mbrandenburger: good catch, `validate_identity` should take as input a serialzed identity as defined in fabric [here](https://github.com/hyperledger/fabric-protos/blob/master/msp/identities.proto#L15). I updated `validate_identity`

g2flyer: using serialized identity is probably easier. Given that both sides have to unserialize it and we also support only X509, one could also just send the X509 cert plus msp id string.  But the duplicate deserialization is a non-issue and using serialized identity would match the fabric formats (even though i have to admit i hate their nested serializations ...)

Apropos "hate": is there anyway to tell github _not_ to to hide conversations which contains unresolved discussions?

docs/interfaces.md
mbrandenburger: aha, heading moved down.

docs/interfaces.md
mbrandenburger: Can you please provide a patch fixing the session API with details based on the local attestation protocols?! :)

docs/interfaces.md
mbrandenburger: ok, will add TODO in the API document to show where UMLs need to be upgraded.

docs/interfaces.md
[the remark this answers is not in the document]
mbrandenburger: there is a new comment regarding the discovery of endorsing FPC-peers.

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+func queryCCInfos(chaincode_id string) (chaincode_ek []byte, allCredentials []Credentials) {}
+func queryCCInfo(chaincode_id string, enclave_id string) (chaincode_ek []byte, credentials Credentials) {}
+
+// register a new FPC chaincode enclave instance
+func registerEnclave(credentials Credentials) error {}
+
+// key registration
+func registerCCKeys(msg CCKeyRegistrationMessage) error {}
+
+// key distribution
```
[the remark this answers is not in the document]
mbrandenburger: added a comment

docs/interfaces.md
```diff
@@ -0,0 +1,568 @@
...
+
+
+// triggered by an admin
+func initEnclave(params CC_Parameters) (Credentials, error) {}
+
+// key generatation 
+func generateCCKeys(params CC_Parameters) (SignedCCKeyRegistrationMessage, error) {}
+
+// key distribution
+func exportCCKeys(credentals Credentials) (SignedExportMessage, error) {}
```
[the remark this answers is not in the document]
mbrandenburger: added a comment

g2flyer:
Below a dump regarding the tlcc-ecc session. Sorry, nothing integrated into text from this PR. Hopefully, it is at least parseable .. :-)

# Table of Contents

1.  [ECC - TLCC Tunnel Design Outline](#org1c1b917)
    1.  [public API](#org0d0a724)
    2.  [control flow of implementation](#orgaa4e54c)
    3.  [sgx_dh_* background](#orgd01ea82)


<a id="org1c1b917"></a>

# ECC - TLCC Tunnel Design Outline


<a id="org0d0a724"></a>

## public API

    - ecc_enclave
      - session_setup(channel_id) -> {ctx, channel_hash, tlcc_mrenclave, rc}
      - session_request(ctx, protobuf SessionRequest{TransactionCtx ctx, req anyof ...} req) -> {protobuf SessionResponse{ anyof ...} response, rc}
    - tlcc_enclave
      - register_request_handlers(function(SessionRequest)->SessionResponse f)
      - rest should be "under the cover", driven by ecalls


<a id="orgaa4e54c"></a>

## control flow of implementation

    - session_setup
      - ecc_enclave (#include <sgx_dh.h>)
        - ectx = EccTlccSessionCtx{sgx_dh_session_t dh, channel_id cid = cid, channel_hash ch, sgx_dh_session_enclave_identity_t tlcc_identity, sgx_key_128bit_t aek}
        - sgx_dh_init_session(initiator, ectx.dh)
        - init_req = protobuf SessionSetupInitMsg{channel_id cid = ectx.cid, enclave_id eid = eid}
        - ecc_enclave-tlcc_enclave ocall+ecall(init_req)
          - in tlcc_enclave (#include <sgx_dh.h>)
    	- tctx = TlccEccSessionCtx{sgx_dh_session_t dh, channel_id cid = init_req.cid, enclave_id eid = init_req.eid, sgx_dh_session_enclave_identity_t ecc_identity, sgx_key_128bit_t aek}
    	- sgx_dh_init_session(responder, tctx.dh)
    	- init_rsp = protobuf SessionSetupInitResponseMsg{sgx_dh_msg1_t msg1}
    	- sgx_dh_responder_gen_msg1(init_rsp.msg1, tctx.dh)
    	- return(init_rsp}
        - compl_req = protobuf SessionSetupCompleteMsg{sgx_dh_msg2_t msg2}
        - sgx_dh_initiator_proc_msg1(init_rsp.msg1, compl_req.msg2, ectx.dh)
        - ecc_enclave-tlcc_enclave ocall+ecall(compl_req)
          - in tlcc_enclave
    	- compl_rsp = protobuf SessionSetupCompleteResponseMsg{sgx_dh_msg3_t msg3, channel_id cid = tctx.cid, channel_hash ch = channel_hash_for(tctx.cid), enclave_id eid = tctx.eid, mac}
    	- sgx_dh_responder_proc_msg2(compl_req.msg2, compl.rsp.msg3, tctx.dh, tctx.aek, tctx.ecc_identity)
    	- compl_rsp.mac = GMAC(tctx.aek, <compl_req.cid, compl_req.ch, compl_req.eid>) ~= tag of (AES-GCM-Enc(tctx.aek, aad=<compl_req.cid, compl_req.ch, compl_req.eid>, msg={})
    	- return(compl_resp}
        - sgx_dh_initiator_proc_msg3(compl_rsp.msg3, ectx, ectx.aek, ectx.tlcc_identity)
        - GMAC_Verify(ectx.aek, <ectx.cid, compl_rsp.ch, eid>)
        - ectx.ch = compl_rsp.ch
        - return success
    
    - session_request(ectx, request)
      - ecc_enclave
        - req_msg = protobuf SessionRequestMsg{Bytes enc_req, nonce n=random()}
        - req_msg.enc_req = AES-GCM-Enc(ectx.aek, aad={req_msg.n}, msg={request})
        - ecc_enclave
          - ecc_enclave-tlcc_enclave ocall+ecall(req_msg)
    	- in tlcc_enclave
    	  - request = AES-GCM-Dec(tctx.aek, aad={req_msg.n}, msg={req_msg.enc_req})
    	  - rsp_msg = protobuf SessionRequestMsg{Bytes enc_rsp}
    	  - process(request) -> response
    	  - req_msg.enc_rsp = AES-GCM-Enc(tctx.aek, aad={req_msg.n}, msg={response})
    	  - return(req_resp}
        - response = AES-GCM-Dec(ectx, ectx.aek, aad={req_msg.n}, msg={rsp_msg.enc_rsp})
        - return response


<a id="orgd01ea82"></a>

## sgx_dh_* background

    - functions
      - [[https://download.01.org/intel-sgx/latest/linux-latest/docs/Intel_SGX_Developer_Reference_Linux_2.10_Open_Source.pdf][Developer Reference Guide v2.10]]
        - overview/flows: page 97-102
        - function list summary with pointers: 121ff
        - function reference: page 305ff
          - trusted (#include <[[https://github.com/intel/linux-sgx/blob/master/common/inc/sgx_dh.h][sgx_dh.h]]>) [[[/home/msteiner/src/projects/intel/sgx/sdk/public/src/linux-sgx.git/common/inc/sgx_dh.h][local]]]
    	- sgx_dh_init_session (role=initiator, ctx) -> status
    	- (app: session_request_ocall -> session_request_ecall)
    	- sgx_dh_init_session (role=responder, ctx) -> status
    	- sgx_dh_responder_gen_msg1(*msg1, ctx) -> status
    	- (app: session_X_ocall -> session_X_ecall; [missing from
    	  flow diagram, but essentially is return flow from above ocall/ecall!])
    	- sgx_dh_initiator_proc_msg1(msg1, *msg2, ctx) -> status
    	- (app: session_exchange_ocall -> session_exchange_ecall)
    	- sgx_dh_responder_proc_msg2(msg2, *msg3, ctx, *eak, *initiator_identity) -> status
    	- (app: session_X_ocall -> session_X_ecall; [missing from flow diagram?!])
    	  (      note: this flow can also piggyback first payload ..)
    	- sgx_dh_initiator_proc_msg3(msg3, ctx, *aek, *responder_identity) -> status
          - key-types
    	typedef struct _sgx_dh_session_enclave_identity_t {
    	    sgx_cpu_svn_t     cpu_svn;
    	    sgx_misc_select_t misc_select;
    	    uint8_t           reserved_1[28];
    	    sgx_attributes_t  attributes;
    	    sgx_measurement_t mr_enclave;
    	    uint8_t           reserved_2[32];
    	    sgx_measurement_t mr_signer;
    	    uint8_t           reserved_3[96];
    	    sgx_prod_id_t     isv_prod_id;
    	    sgx_isv_svn_t     isv_svn;
    	} sgx_dh_session_enclave_identity_t;
    
          - notes
    	- no way to bind explicit session context into the
    	  hand-shake, so this will need additional messages
    	- we also need a msg0 from initiator to responder, but it
    	  is not secured!
    	- above also means that _responder_ is first to be able to
    	  add payload, not _initiator_, i.e., you need 2 round-trip
    	  instead of only 1 to start :-(
    	  - you can, though, send public insecured content in msg0
    	    and just confirm in msg3
    	- two protocol versions (page 98-99 & 99-100, with v2
    	  version sligthly more context, all internal to impl of
    	  api & presumably more secure)
    	  - selected by #define SGX_USE_LAv2_INITIATOR
    	    - in below sample, done in Makefile!
    	       Enclave_C_Flags += -DSGX_USE_LAv2_INITIATOR
    	       Enclave_Cpp_Flags += -DSGX_USE_LAv2_INITIATOR
    	    - both versions exist, though, in library, so could be
    	      used concurrently
    	      #ifdef SGX_USE_LAv2_INITIATOR
    	      #define sgx_dh_initiator_proc_msg1 sgx_LAv2_initiator_proc_msg1
    	      #define sgx_dh_initiator_proc_msg3 sgx_LAv2_initiator_proc_msg3
    	      #else
    	      #define sgx_dh_initiator_proc_msg1 sgx_LAv1_initiator_proc_msg1
    	      #define sgx_dh_initiator_proc_msg3 sgx_LAv1_initiator_proc_msg3
    	      #endif

-   samples
    -   [LocalAttestation/](https://github.com/intel/linux-sgx/blob/master/SampleCode/LocalAttestation/)
        -   [EnclaveInitiator/EnclaveMessageExchange.cpp](https://github.com/intel/linux-sgx/blob/master/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveMessageExchange.cpp) ([sourcegraph](https://sourcegraph.com/github.com/intel/linux-sgx@master/-/blob/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveMessageExchange.cpp#L158))
        -   for verification (including debug flags, but not others, say 64-bit)
            -   [verify<sub>peer</sub><sub>enclave</sub><sub>trust</sub>](https://sourcegraph.com/github.com/intel/linux-sgx@ce4a18d9a0963b14f717cc7462241cfadd233f55/-/blob/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveInitiator.cpp#L144)

docs/interfaces.md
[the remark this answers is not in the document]
bvavala: Ok, it should be enough for now.

However, this [document (part 2)](https://hyperledger-fabric.readthedocs.io/en/release-2.2/txflow.html) corroborates my second point: we need a mechanism for checking the proposal (including its creator, and thus this API).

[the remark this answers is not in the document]
g2flyer: we definitely need to validate the signed proposal.  Using the proposed `validate_identity` we should be able that the creator -- as expressed by `serializedIdentity` which contains the msp-id and the x509 cert with pk, name and role [might help, though, explicitly mention that in the comment around `validate_identity` -- is a valid party grounded in the specified MSP-id root-of-trust and the rest of the signature verification of the proposal we can do locally to ecc based on the resulted validated public key?  Is there any other step in verifying the signedproposal in the enclave we need `tlcc` assistance?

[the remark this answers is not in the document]
bvavala: (see the link I have attached)
The endorsing peer verifies also that (i) the proposal has not been submitted in the past, (ii) the submitter respects the channel policy. In theory, TLCC should do this. In practice, we can rely on the validators for (i), and delegate (ii, or more precisely, any authentication/authorization) to the enclave.

docs/interfaces.md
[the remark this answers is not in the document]
bvavala: Again, from the `enclave_id` is should be possible to go back to the `chaincode_id`. However, though it appears redundant, it might be useful in the case the enclave is not specialized for a particular chaincode app.

docs/interfaces.md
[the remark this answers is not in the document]
g2flyer: ok, i see now what you mean (didn't initially get what you meant by "part 2").  Good points.

Regarding the replay protection, at first sight it certainly seems mainly a waste of resource you save by checking, as during submission of a replayed transaction it should fail due to rw-set conflicts?  This assumes, though, it is state-changing, maybe there could be issues of transactions which might be non-state-changing dependent on state and somebody could try to replay such a transaction at a later time when it would be state-changing? So integrity-wise there might or might not be issue?  For FPC maybe confidentiality could add additional twists: As an original tx/query submitter can always recreate a new query, so from overt channel (results to querier) perspective a replay shouldn't help?  If there would be covert channel, then a resubmit could potentially help in throttling attacks but that seems rather esoteric?  So i would somewhat agree that we can punt on replay protection to validators/rw-conflict (although i have not a good feeling yet how much this _this-tx-does-not-change-state-now-but-could-potentially-later scenario_ is an issue or not.)

regarding authorization, though, it becomes a bit more sensitive as we are talk also about confidentiality (for queries), not only integrity.  So maybe the `validated_identity` should maybe return which permissions the client has and ecc should immediately abort if client doesn't have `read` privs and abort on first write-attempt if it doesn't have `write` permissions?  Of course, we could also punt on that but would have to document that reader/writer polices could not be trusted to provide confidentiality of information protected fpc cc?

docs/interfaces.md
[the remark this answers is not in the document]
bvavala: I agree on the following: having a single call to get an enclave's credentials makes sense for validation.
That said, the `chaincode_ek` is unnecessary in that setting.

Also, back to another comment of mine (that I cannot find), now I see why `chaincode_id` is needed: it's because, in ercc's state,  `namespaces/registered/<enclave_id> -> SignedCCKeyRegistrationMessage` does not contain `cc_params`, which is under `<chaincode_id>/<enclave_id>`. So, by reorganizing ercc's state, we can get rid of the `chaincode_id` -- in fact, `enclave_id` is enough the purpose and, if `chaincode_id` is necessary, it can be found in the `credentials`.

Let me propose a possible improvement, having as objective the reduction of: input and output parameters,
* add `func queryChaincodeEncryptionKey(chaincode_id string) (chaincode_ek []byte) {}`
* `queryCCInfos` becomes `func queryListChaincodeEnclaveCredentials(chaincode_id string) (allCredentials []Credentials) {}`
* `queryCCInfo` becomes `func queryEnclaveCredentials(enclave_id string) (credentials Credentials) {}`

In ERCC state:
* `namespaces/registered/<chaincode_id> -> []enclave_ids`
* remove `namespaces/credentials/<chaincode_id>/<enclave_id>` (hence, no duplicate credentials)
* `namespaces/credentials/<enclave_id> -> Credentials` (ercc can still build a list of credentials from a list of ids)
* `namespaces/key_registration/<chaincode_id> -> SignedCCKeyRegistrationMessage` (the encryption key can be taken directly from here)
* in `CCKeyRegistrationMessage`, `enclave_vk` becomes `enclave_id`, which is used to look up `enclave_vk` for signature verification.

docs/interfaces.md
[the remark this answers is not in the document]
bvavala: > during submission of a replayed transaction it should fail due to rw-set conflicts?

**Probably yes**, in a concurrent execution: send proposal P1 and replay it immediately; you get two separate endorsements/transactions where the updated keys likely conflict with each other.
**Maybe not**, in a serialized execution: send proposal P1 (e.g., incrementCounter), commit transaction; replay P1, commit transaction; here the counter will be incremented by two with a single proposal.

The check that Fabric does should be [here](https://github.com/hyperledger/fabric/blob/c799dcc188d76c5d3f3e2cdab931f30a00273d17/core/endorser/endorser.go#L275). It's based a transaction id value computed over the proposal and stored (at some point) on the ledger.

> So maybe the validated_identity should maybe return which permissions the client has and ecc should immediately abort if client doesn't have read privs and abort on first write-attempt if it doesn't have write permissions?

In theory, yes.
However, this is all Fabric logic that TLCC should implement. Following Yacov's recommendation, we can simplify.

> Of course, we could also punt on that but would have to document that reader/writer polices could not be trusted to provide confidentiality of information protected fpc cc?

In practice, this is (part of) the solution. The other part (in my answer above) is to tell a developer how to leverage the fpc chaincode to implement an equivalent mechanism.

docs/interfaces.md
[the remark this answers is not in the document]
bvavala: > You are proposing a queryRegisteredEnclaves(chaincode_id string)

That's right, in the end... we call it "enclave registry"!!

Anyway, no strong opinion about the arguments above.
So, resolve this and continue a related conversation in [this comment](https://github.com/hyperledger-labs/fabric-private-chaincode/pull/420#discussion_r468235021).

docs/interfaces.md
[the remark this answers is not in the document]
g2flyer: > > during submission of a replayed transaction it should fail due to rw-set conflicts?
> [..]
> **Maybe not**, in a serialized execution: send proposal P1 (e.g., incrementCounter), commit transaction; replay P1, commit transaction; here the counter will be incremented by two with a single proposal.

Oops, of course, this is a much simpler examples where replay works (badly) than my contrieved sometimes-state-changes example. 

> The check that Fabric does should be [here](https://github.com/hyperledger/fabric/blob/c799dcc188d76c5d3f3e2cdab931f30a00273d17/core/endorser/endorser.go#L275). It's based a transaction id value computed over the proposal and stored (at some point) on the ledger.

This code seems to be executed by the endorsers. You mentioned above also "rely on the validators for (i)": Do the validators automatically also cross-check this?  If so, i think we can punt on them as good peers would ignore replays from a integrity perspective and from a confidentiality perspective there seems only esoteric direct leakage (and not indirect via integrity corruption, which would be caught by validators?)   If standard validators would not do it, this (contrary to authorizaton) seems a bit a tough to say we do not support replay protetion and  punt it to the app developer?   If we can live with some false-positives for replays, we could handle replay protection with a bloom-filter.  Actually, for channel & ercc transactions we cannot punt to app developers at all? So we more or less have something inside tlcc?   We currently have these three opaque kvs_t files as part of tlcc internal state but seems we might want to be more explicitl. And of course this ties now this to (some) resolutions of #402 ... :-(

> > So maybe the validated_identity should maybe return which permissions the client has and ecc should immediately abort if client doesn't have read privs and abort on first write-attempt if it doesn't have write permissions?
> 
> In theory, yes.
> However, this is all Fabric logic that TLCC should implement. Following Yacov's recommendation, we can simplify.

That's what i meant with punt below. But i agree that besides explicit telling you cannot rely on ledger-specified authorization rules (which we would have to do at a minimum) ...
 
> > Of course, we could also punt on that but would have to document that reader/writer polices could not be trusted to provide confidentiality of information protected fpc cc?
> 
> In practice, this is (part of) the solution. The other part (in my answer above) is to tell a developer how to leverage the fpc chaincode to implement an equivalent mechanism.

... pointing out how to do it themselves is even better.  Certainly, per-app authorization policies should be easy to handle via `get_creator_name` ...

docs/interfaces.md
```diff
@@ -0,0 +1,592 @@
+# FPC Component Interfaces and States
```
g2flyer, line 1: BTW: shouldn't we put this file into `docs/design/fabric-v2+` as the rest of the design docs?

docs/interfaces.md
g2flyer: will do. See below for a brain dump.  As this is really a separate module with (hopefully :-) clean interfaces, i will move the stuff to separate `.md` files ...

docs/interfaces.md
bvavala: It seems appropriate.

mbrandenburger: moved!

g2flyer:
FYI: to make sure we do not forget updating the UML diagrams as mentioned in comments here, i've created a corresponding issue #432

docs/interfaces.md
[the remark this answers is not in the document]
mbrandenburger: Thanks Bruno, I partially integrated your proposed changes.

- Introduced `queryChaincodeEncryptionKey` and removed the key from "old" `queryCCInfo(s)`
- Renamed `queryCCInfo` and `queryCCInfos`
- Add clarification that credentials are not stored twice on ercc state; my representation of how to store the credentials was just misleading and confusing.

Regarding removing the `chaincode_id` from the keys and and adding `namespaces/registered/<chaincode_id> -> []enclave_ids` I'm hesitating. Removing as many parameters as possible from the methods exposed by ERCC is a good. 

However, I see the following problem with the proposed layout:
```
namespaces/registered/<chaincode_id> -> []enclave_ids
namespaces/credentials/<enclave_id> -> Credentials
```

Concurrent registration operation:
Let's imagine concurrent enclave registrations for Chaincode A. Let's say a new FPC chaincode is installed and the orgs start setting up their chaincode enclaves and invoke `registerEnclave` at ERCC. (I ignore the fact that MVP does not support this situation but once we are moving forward this can happen).
The implementation of `registerEnclave` will perform the following:
```go
putState("namespaces/credentials/some_new_enclave_id", somecredentials)
registered := getState("namespaces/registered/some_chaincode_id")
registered = append(registered, some_new_enclave_id)
putState("namespaces/registered/some_chaincode_id", registered)
```
These storage operations may lead to write conflicts of concurrent registerEnclave transactions. Even with a single admin issuing the registration of a bunch of enclaves, the admin would need to wait until a register tx is committed to the ledger before registering the next enclave. By integrating the chaincode_id in the key the example code from above is reduced to:
```go
putState("namespaces/credentials/some_chaincode_id/some_new_enclave_id", somecredentials)
```
and no write conflicts will happen.

Another aspect is performance and clean code:
Using `namespaces/credentials/some_chaincode_id/some_new_enclave_id` allows us to perform a range query with `namespaces/credentials/some_chaincode_id` as prefix to query all registered chaincode enclaves with a single `getStateByRange` call instead of first get the all enclave ids and subsequently another getState operation for each enclave id. Note that for each getState invocation, the chaincode shim contacts the peer.

docs/interfaces.md
mbrandenburger: `ocall_create_evidence` removed

docs/interfaces.md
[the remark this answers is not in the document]
g2flyer: as mentioned on Tuesday, you could have the best of both worlds, though, with following schema?
```
namespaces/registered/<chaincode_id>/<enclave_ids> -> defined
namespaces/credentials/<enclave_id> -> Credentials
```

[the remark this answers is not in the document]
g2flyer: that said, i think we should also not overthinking this: Doesn't have to be perfect, good enough should work also :-)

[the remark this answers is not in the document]
bvavala: > as mentioned on Tuesday, you could have the best of both worlds, though, with following schema?
> 
> ```
> namespaces/registered/<chaincode_id>/<enclave_ids> -> defined
> namespaces/credentials/<enclave_id> -> Credentials
> ```

I agree

> that said, i think we should also not overthinking this: Doesn't have to be perfect, good enough should work also and what you propose definitely does (in particular as we don't really need any lookup `enclave_id->chaincode_id`, so above discussion is a bit academic ...:-)

I agree. Let's go ahead.

Issue #351, opened 2020-05-07
Build-tag selectable build-time fabric system-chaincode plugin configuration


g2flyer:
Is your feature request related to a problem? Please describe.**
Currently we apply the TLCC patch to fabric which are specific to FPC and work only when building Fabric including FPC, so is unlikely to be upstreamed.   The bigger issue in Fabric though is that there is no general (build-time) way to add system-chaincode in an easy and configurable way.

**Describe the solution you'd like**
The basic idea is to use build-tags for a particular system-chaincode configuration. Practically, it would require outsourcing from `internal/peer/node/start.go` list of to-be-configured system chaincodes into a struct defined in a separate, build-tag-controlled file which then would be used inside `start.go` to iterate through (instead of the explicit list as currently done).

- [X] strategy/design on how to handle it
- [X] FPC PR #422
- [ ] [Fabric PR 1822](https://github.com/hyperledger/fabric/pull/1822)
- [ ] Adapt FPC to Fabric (once Fabric PR is integrated)

**Describe alternatives you've considered**

Carrying around a patch of a specific file. The proposed solution offers a way to replace a built in file instead of modifying it.

**Additional context**

g2flyer:
See also [FAB-17564](https://jira.hyperledger.org/browse/FAB-17564?focusedCommentId=68563&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-68563) which touches on issue (but does not give any guidance as of 2020-05-09)

g2flyer:
Even better approach than above might be to define a package global static variable which is used in `internal/peer/node/start.go` and then populate/extend it via `init` functions (which golang supports multiples per package and can be defined in separate build-tag controlled files).  The standard set of course could be part of the static variable initialization and on extensions might be done in separate `init` functions ....

mbrandenburger:
FPC does not require custom system chaincode anymore. In case this integration concept comes back we can reopen this issue.

PR #422, opened 2020-08-01
TLCC via go build tags

g2flyer:

This PR adds a patch to fabric which enables to add additional system chaincodes to fabric controlled via go build tags.   Additionally, it changes the build-system to enable building via (mod) vendoring as fabric does and (mostly) enables also fpc mod-vendored via github rather than the GOPATH-referenced linkage to fpc as currently.

***SCC additions (and TLCC/qscc) s via go build tags***

This is done in four patches:
- `2_0001-Extension-point-for-additional-SCCs-opt_scc.patch`
- `2_0002-Add-FPC-SCCs-TLCC-via-opt_scc-extension-point.patch`
- `2_0003-Mock-SCC-for-opt_scc-extension-point.patch`
- `2_0004-Deploy-qscc-via-opt_scc-extension-point.patch`

The first is the general extension framework, the second is the use of it by FPC for `tlcc` and the third is a mock scc which is enabled using a build tag `mock_scc` and just echos (string) requests and the fourth would show how we could also use the framework for `qscc`.  (Not sure whether fabric would ever consider qscc to deployed that way, but as a separate patch it would be easy to opt-in or not and it was one "real" scc to test with...)

The parameters to the factory is a superset what tlcc needs and covers the set of common parameters by the standard SCCs.  It would also exactly match what qscc uses, so it could also be "outsourced" from `start.go` to `scc_opt`, but i haven't done that for now .

***Building everything***
The new setup builds like fabric in a vendored fashion.  The changes also make sure that fabric (without fpc) still properly builds after our patches with a simple `make` in ${FABRIC_PATH}  (didn't do so in the last version).  There are though a few caveats:
- go mod vendor does not copy all c-files. There is ugly hack which overcomes this. However, due to libraries would cannot be build directly via cgo, it still wouldn't work vendored
  This is currently hanled vie via CGO_{CFLAGS,LDFLAGS} variables pointing to header-file & library location (see fabric/Makefile or header of tlcc/enclave/enclave_stub.go)

- vendoring also causes a problem in presence of (needed) generated .go file like `ercc/attestation/provider.go`

  With local-referenced FPC repo, it is sufficient to just build it before fabric peer is built. However, with properly vendored version from github (as fabric almost certainly will insist), that wouldn't would not work. (You can make it temporariliy work by cheating and manually adding the file after vendoring that but certainly would not acceptable for fabric maintainers which would have to do this for each time they do go mod vendor ...)

So in the mid to long run i think we will have to
- split go and c into separate repos so we have only one version of go source as you build fabric utilities, and that is the vendored one 
- for the go repo we will have to ensure that we have no auto-generated go-file touched by peer build and, specificially, we will have a different way to define the attestation type. (auto-generated mrenclave.go should still be ok as none is baked into tlcc or validators)
- for the c repo, we should have a `make install` (or `make dist`) targets with install/package the libraries and header files required to build fabric peer which then can be used with (hopefully simpler) CGO_* flags. At the same time we probably would also want to think of packaging a binary distro for FPC CC developers.

***Other notes***
- pluginsenabled build tag doesn't exist anymore, so i've dropped it from Makefile (though not yet docu ..)
- i've split the fabric go mode patche so you can easily test with local-version of fpc vs github version

First step towards #351

To see the fabric diffs as source diffs and not diffs of the patches, see also the [diff on my fabric branch:](https://github.com/g2flyer/fabric/compare/v2.2.0...g2flyer:msteiner.opt-sccs?expand=1).  It includes all patches but you can select and view the individual commits to drill down ...


fabric/patches/2_0002-Add-FPC-SCCs-TLCC-via-SCC-Extension-point.patch
```diff
@@ -0,0 +1,49 @@
...
++	"github.com/hyperledger/fabric/core/aclmgmt"
++	"github.com/hyperledger/fabric/core/peer"
++	"github.com/hyperledger/fabric/core/scc"
++)
++
++type tlccFactory struct{}
++
++func (f tlccFactory) newOptSCC(aclProvider aclmgmt.ACLProvider, p *peer.Peer) scc.SelfDescribingSysCC {
++	return tlcc.New(p)
++}
```
mbrandenburger, line 39: Just wondering if we could replace this factory object with just a factory function, similar how http.handlers are used. The SSCRegistry would keep a list of factory functions instead of objects. This way we prevent any naming conflicts in the `node` package with other customSSC.

sccs_opt_fpc.go becomes:
```go
func init() {
    logger.Debug("Enabling tlcc-factory")
    optionalSCCs.add(func(aclProvider aclmgmt.ACLProvider, p *peer.Peer) scc.SelfDescribingSysCC {
        return tlcc.New(p)
    })
}
```

ssc_opt.go becomes:
```go
type FactoryFunc func(aclProvider aclmgmt.ACLProvider, p *peer.Peer) scc.SelfDescribingSysCC

type optionalSCCRegistry struct {
	registeredFactories []FactoryFunc
}

func (f *optionalSCCRegistry) add(ff FactoryFunc) {
	f.registeredFactories = append(f.registeredFactories, ff)
}

func (f *optionalSCCRegistry) list() []FactoryFunc {
	return f.registeredFactories
}
```
and in start.go
```go
for _, f := range optionalSCCs.list() {
		osi := f(aclProvider, peerInstance)
...
```

This is just an alternative implementation proposal. Not sure if this is actually better or not, but reduces the need to have a factory object for each custom scc and just provides a factory function. Maybe the real factory is a bit more future proof, in particular, when the factory would set some default values, but in this case we would introduce a factory for the factory. Let's keep it simple.

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
```diff
@@ -0,0 +1,79 @@
...
+@@ -595,6 +595,13 @@ func serve(args []string) error {
+ 		"_lifecycle": {},
+ 	}
+ 
++	optSCCInsts := []scc.SelfDescribingSysCC{}
++	for _, sf := range optionalSCCs.list() {
++		osi := sf.newOptSCC(aclProvider, peerInstance)
++		optSCCInsts = append(optSCCInsts, osi)
++		builtinSCCs[osi.Name()] = struct{}{}
++	}
```
mbrandenburger, line 63: As we want to have this upstream at some point we should also have some tests covering loading more than one optSSC.

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
```diff
@@ -0,0 +1,79 @@
...
+ internal/peer/node/sccs_opt.go | 31 +++++++++++++++++++++++++++++++
+ internal/peer/node/start.go    |  9 ++++++++-
+ 2 files changed, 39 insertions(+), 1 deletion(-)
+ create mode 100644 internal/peer/node/sccs_opt.go
+
+diff --git a/internal/peer/node/sccs_opt.go b/internal/peer/node/sccs_opt.go
+new file mode 100644
+index 000000000..d46fa4811
+--- /dev/null
++++ b/internal/peer/node/sccs_opt.go
```
mbrandenburger, line 17: looking in this directory `sccs_opt.go` looks a bit at the wrong place. Works fine for now, but for upstream maybe better in `github.com/hyperledger/fabric/core/scc`?

mbrandenburger:
@MHBauer can you have a quick look at this and share your thoughts on the general approach? Thanks

fabric/patches/2_0002-Add-FPC-SCCs-TLCC-via-SCC-Extension-point.patch
```diff
@@ -0,0 +1,49 @@
...
++
++var tf tlccFactory
++
++func init() {
++	logger.Debug("Enabling tlcc-factory")
++	optionalSCCs.add(tf)
++}
```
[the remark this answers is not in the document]
g2flyer: I've actually tried first to pass it without var but somehow got syntax wrong and then did the expedient way.  In fact, i thought first i could do it via an anonymous type but couldn't figure out that either, the key to anonomous handler is what you point out below is that i don't even need a type for the realization, a single lambda is enough.

fabric/patches/2_0002-Add-FPC-SCCs-TLCC-via-SCC-Extension-point.patch
g2flyer: didn't realize that you can just pass a function to an interface having a single function and don't need a type in that case. Definitely much better, will change!

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
g2flyer: i guess for test-cases having zero, one or more (say 3) optSSC should be sufficient.  Probably should also test that the handler returns non-nil and then do a corresponding test-case.  Any other error-condition i should check (and write test-case for it)?

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
g2flyer: yep, `internal` is not the best case, i would think though better to add as subdir inside scc, not directly in top-level?

g2flyer:
@mbrandenburger i've pushed an update which should address your comments except the testing part. Not 100% clear to me what exactly and what way to test is necessary and appropriate. Probably best to chat at some point in a call?   
Btw: besides addressing the comments, i've also opt-scc'ized (as a separate patch) `qscc`: -- which together with `tlcc` already give the test case of more than one you mentioned :-) -- and made sure that all patches pass `make linter` in `$FABRIC_PATH`

g2flyer:
btw: to see the fabric diffs as source diffs and not diffs of the patches, see also the [diff on my fabric branch:](https://github.com/g2flyer/fabric/compare/v2.2.0...g2flyer:msteiner.opt-sccs?expand=1).  It includes all patches but you can select the invdividual commits ...

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
mbrandenburger: we will figure out what the proper place is by creating this PR on the Fabric repo :) maintainers will tell :D

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
mbrandenburger: ah this test looks already good to me. No other error condition comes to my mind at the moment.
As a testing bonus maybe one could use ginkgo/gomega to make tests shorter (and more readable) as often used in Fabric. Here an [example](https://github.com/hyperledger/fabric/blob/master/core/endorser/endorser_test.go).

So instead of 
```go
ff = opt_scc.ListFactoryFuncs()
if ff == nil {
    t.Error("factory function list is nil")
}
if len(ff) != 3 { ...
```
you could write something like that
```go
import . "github.com/onsi/gomega"

ff = opt_scc.ListFactoryFuncs()
Expect(ff).ShouldNot(BeNil())
Expect(len(ff)).Should(Equal(3))
```

fabric/patches/3_0001-Add-FPC-validator-to-library.patch
```diff
@@ -9,25 +9,32 @@ peer.
...
++
 +package library
 +
 +import (
-+	validation "github.com/hyperledger/fabric/core/handlers/validation/api"
-+	decoration "github.com/hyperledger/fabric/core/handlers/decoration"
-+	"github.com/hyperledger-labs/fabric-private-chaincode/validation/fpc"
 +	"github.com/hyperledger-labs/fabric-private-chaincode/validation/ercc"
++	"github.com/hyperledger-labs/fabric-private-chaincode/validation/fpc"
++	decoration "github.com/hyperledger/fabric/core/handlers/decoration"
```
mbrandenburger, line 36:
```suggestion
+	"github.com/hyperledger/fabric/core/handlers/decoration"
```

Should also work like that right?

fabric/patches/9_0001-add-proxy-support-for-docker-in-peer.patch
```diff
@@ -10,42 +10,42 @@ Signed-off-by: Michael Steiner <michael.steiner@intel.com>
...
++		"HTTPS_PROXY",
++		"NO_PROXY",
++	}
++	for _, name := range proxyEnvVarNames {
++		if val, found := os.LookupEnv(name); found {
++			proxyEnv = append(proxyEnv, name+"="+val)
++		}
++	}
++	logger.Debugf("Util Proxy config used for docker: %v", proxyEnv)
++	return proxyEnv
```
mbrandenburger, line 44: did you use gofmt for this formatting?

fabric/patches/2_0003-Deploy-qscc-via-opt_scc-framework.patch
```diff
@@ -0,0 +1,77 @@
...
++	"github.com/hyperledger/fabric/core/scc/opt_scc"
+ 	"github.com/hyperledger/fabric/protoutil"
+ )
+ 
+@@ -226,3 +229,11 @@ func getBlockByTxID(vledger ledger.PeerLedger, rawTxID []byte) pb.Response {
+ func getACLResource(fname string) string {
+ 	return "qscc/" + fname
+ }
++
++func init() {
```
mbrandenburger, line 31: I would add a comment here, quickly explaining that qscc is instantiated as an optional system chaincode using the opt factory. Just as a comment for us, I doubt that this go upstream but definitely shows how the factory can also used for non-fpc components.

mbrandenburger (commented):
Thanks Michael, the refactored factory looks neat!

fabric/patches/3_0001-Add-FPC-validator-to-library.patch
g2flyer: maybe, only reason i touched that file (and only thing i did) is to make it properly formatted to pass `make linter`. Would prefer not changing other change anything else in this pr ..

fabric/patches/9_0001-add-proxy-support-for-docker-in-peer.patch
g2flyer: yep, the changes to the non-tlcc related patches are all just to properly `gofmt ..` so it passes `make linter`

fabric/patches/2_0001-Extension-point-for-go-build-tag-enabled-additional-.patch
g2flyer: gingko seemed overkill by maybe just gomega might help.  Seems gomega also supports `reflect.DeepEqual`. Let me give a try ...

g2flyer: doesn't work as outlined above: gomega requires a registered fail handler and the only pre-existing one i saw is the one with ginko. And if you register that but do not use ginko for the rest, then you don't get meaningful output if your test fails and actually get a stack trace. I.e., you would have to do the whole ginko shabang which i don't think is worth the effort (nor personally do i think it is really more readable).  I.e., i'll leave if for now ...

g2flyer: actually, what _does_ work is to do it xunit style. slightly different syntax but somewhat same semantics ..

fabric/patches/2_0003-Deploy-qscc-via-opt_scc-framework.patch
g2flyer: if you look at the patch it's obvious and there are also comments and it refers to the opt_scc package, really don't see what more to add ...

g2flyer:
> * would it make sense to add one line in the Fabric Makefile to build the FPC-enabled peer? (maybe later when we complete the refactoring)

With the current command i don't think it makes sense as it is somewhat transient until the go templating (and any yet unknown follow-up issues with `go mod vendor` are resolved. And hopefully in the end the only thing you would have to do is add fpc to GO_TAGS and run make on any arbitrary target.  So not sure whether in that case you really need a new target? (also going via GO_TAGS handles also automatically the situation whenever we might have to add fpc-stuff to other tools?)  Probably best to post-pone when we will try upstreaming the FPC-specific PRs?
 
> * the `qscc` thing seems redundant. First, I cannot find a reason why they may want it. Second, the TLCC part already represents an example on how to use the factory.

I would agree it's not very likely that `qscc` will but using the framework (but who knows where it goes).  But my primary motivation was that i think we do _not (yet) include FPC-specific_ patches into the PR, namely for two reasons: 
- it would be good to have an example which they can run and verify that it works end-to-end.  However, doing that with TLCC forces them to go also through the whole FPC installation which is a considerable barrier
- additionally, i think it would be good to separate actual fpc integration from this which is just a pre-requisite extension for us to be able to do so.  The actual integration would involve other patches not pertinent to SCCs and even if we give all of that only as an example, I fear it might still lead to a rathole of discussions going way beyond `opt_scc` and into fpc specific aspects.  Might be better to do that in divide'n'conquer.

> * the generated Go file is a temporary problem, since that part will be refactored

I've actually done the refactoring of that as a follow-up patch (turned out to be pretty straightforward; biggest thing was to consistently pass the go tags and, as side-effect, also tried to make makefiles a bit more consistent (e.g., everywhere using $(GO)). Have yet verify that this really solves the vendoring issue. WIP ....

fabric/patches/2_0003-Deploy-qscc-via-opt_scc-framework.patch
mbrandenburger: well, I was thinking about having somewhere some background story why qscc now has a `init` but maybe there is a better place. Maybe this will be part of the Fabric PR.

g2flyer: The init in query.go is now moot. Thinking a bit more on what i said below that we shouldn't use FPC as use-case, i realized that we should have a build-tag enabled version, so finally i did add now a mock scc to `opt_scc/mock` and then changed the qscc patch to have the same pattern: a file in `opt_scc` which just does the init, potentially build-tag enabled, and the actually scc code is elsewhere: so `query.go` is not touched anymore and the context is even more obvious now ...

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
```diff
@@ -0,0 +1,86 @@
+From 0505d4c4c4f7ac665afced25975840be299fe9dc Mon Sep 17 00:00:00 2001
+From: michael steiner <michael.steiner@intel.com>
+Date: Wed, 19 Aug 2020 15:40:39 -0700
+Subject: [PATCH 1/4] Deploy qscc via opt_scc extension point
```
MHBauer, line 4: any reason this shows as [Patch 1/4] even though it seems to be the fourth patch?

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
```diff
@@ -0,0 +1,86 @@
...
+-	qsccInst := scc.SelfDescribingSysCC(qscc.New(aclProvider, peerInstance))
+ 
+ 	pb.RegisterChaincodeSupportServer(ccSrv.Server(), ccSupSrv)
+ 
+@@ -755,7 +752,7 @@ func serve(args []string) error {
+ 	}
+ 
+ 	// deploy system chaincodes
+-	for _, cc := range append([]scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC}, optSCCInsts...) {
++	for _, cc := range append([]scc.SelfDescribingSysCC{lsccInst, csccInst, lifecycleSCC}, optSCCInsts...) {
```
MHBauer, line 80: It seems to me that the desired endpoint might be all systemCCs using the init add to registry? Is that the idea?

fabric/patches/3_0001-Add-FPC-validator-to-library.patch
MHBauer: Sometimes the formatter/importer programs does this duplicative thing, I don't know why.

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
```diff
@@ -0,0 +1,86 @@
...
+ 2 files changed, 29 insertions(+), 4 deletions(-)
+ create mode 100644 core/scc/opt_scc/qscc.go
+
+diff --git a/core/scc/opt_scc/qscc.go b/core/scc/opt_scc/qscc.go
+new file mode 100644
+index 000000000..334e75494
+--- /dev/null
++++ b/core/scc/opt_scc/qscc.go
+@@ -0,0 +1,28 @@
++//+build !test
```
MHBauer, line 19: seems weird to me. only build when not testing? what gets built when testing? The upstream qscc?

I may be missing some context, but maybe it could be solved by a comment with some of the intent described.

fabric/Makefile
```diff
@@ -20,7 +20,7 @@ build: native
 patch: $(FABRIC_PATCHED)
 
 $(FABRIC_PATCHED): patches/*.patch $(FABRIC_PATH)/.git
-	# Note: .git dependency hopefully captures significant changes in fabric dir requiring repatch (and rebuild)
```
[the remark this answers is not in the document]
g2flyer: it's the closest i could figure out in adding a dependency: it wouldn't catch anything you change without committing, but any commit/stage/pull etc should make sure it is rebuilt.

fabric/Makefile
```diff
@@ -33,13 +33,14 @@ $(FABRIC_PATCHED): patches/*.patch $(FABRIC_PATH)/.git
 	fi
 	cd $(FABRIC_PATH) && \
 	git am $(FPC_PATH)/fabric/patches/*.patch && \
-	$(FABRIC_CURRENT_DIFF_COMMITS_CMD) > $@
+	$(FABRIC_CURRENT_DIFF_COMMITS_CMD) > $@ && \
+	$(GO) mod vendor
```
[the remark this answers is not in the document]
g2flyer: the vendoring is required for integration of fpc into fabric's `go.mod` (ie. `fabric/patches/1_*`).  Note that shis wouldn't go into the PR for fabric -- that  PR to fabric would exclude FPC for now and only include `fabric/patches/2_000[134]*`.. -- but was necessary so we do not have to do the earlier (uglier) way of building fpc via `GOFLAGS="-mod=mod"`

fabric/Makefile
```diff
@@ -79,11 +80,15 @@ $(TLCC_ENCLAVE_LIB):
...
-	# build properly if tlcc enclave is not yet built ..
+        # Note: above does not have enough dependencies to trigger a rebuild if (recursively)
+        # any TLCC and dependent code changes. however, it should at least make sure it will
+        # build properly if tlcc enclave is not yet built ..
 	cd $(FABRIC_PATH) && \
-	GOFLAGS="-mod=mod" GO_TAGS=pluginsenabled,FPC $(MAKE) peer
+        env \
+            CGO_CFLAGS="-I${FPC_PATH}/common -I${FPC_PATH}/common/sgxcclib -I${FPC_PATH}/tlcc_enclave/trusted_ledger"  \
+            CGO_LDFLAGS="-L${FPC_PATH}/tlcc_enclave/_build/lib" \
+	   GO_TAGS=FPC \
```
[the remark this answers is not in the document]
g2flyer: not new in this PR, we have done that earlier to add decorators, validators and endorsements directly into code (rather than as goplugins). These plugins though could be added out-of-the-(fabric)-box as we could exploit what the plugins framework already offerred (see `fabric/patches/3_0001-Add-FPC-validator-to-library.patch` for details); that didn't work, though, for SCCs (i.e., TLCC), hence `opt_scc` extension point ...

fabric/patches/2_0002-Add-FPC-SCCs-TLCC-via-opt_scc-extension-point.patch
```diff
@@ -0,0 +1,45 @@
...
++	"github.com/hyperledger/fabric/core/scc"
++)
++
++func init() {
++	logger.Debug("Registring tlcc as system chaincode")
++	AddFactoryFunc(func(aclProvider aclmgmt.ACLProvider, p *peer.Peer) scc.SelfDescribingSysCC {
++		logger.Debug("Enabling tlcc as system chaincode")
++		return tlcc.New(p)
++	})
++}
```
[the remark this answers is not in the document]
g2flyer: yep, init was also not my preferred choice but i couldn't figure out any other way

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
g2flyer: good catch: i have to manually edit that given the way i create these format patches. Should (will) be fixed now ...

fabric/patches/2_0003-Mock-SCC-for-opt_scc-extension-point.patch
```diff
@@ -0,0 +1,102 @@
...
++// Invoke is called with args[0] contains the query function name, args[1]
++func (e *EchoSCC) Invoke(stub shim.ChaincodeStubInterface) pb.Response {
++	args := stub.GetStringArgs()
++
++	mscclogger.Infof("Invoke MSCC(%v)", args)
++
++	bytes := []byte(fmt.Sprintf("%v", args))
++	return shim.Success(bytes)
++}
+--
```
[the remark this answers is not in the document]
g2flyer: The type should be checked:   `New` does return a typed EchoSCC object that later in `mock.go` during `AddFactoryFunc` the the consistency of this type to scc.SelfDescribingSysCC interface is checked. I can verify that by commenting out Name or Chaincode (which would make EchoSCC not match interface SelfDescribingSysCC  anymore) and compilation fails.  So not sure what exactly you are referring to but the code section your comment references seems to not really match your text, so i guessed you was referring to a different code fragment but maybe i just misunderstood you?

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
g2flyer: ideally, yes, but all the other SCCs have so different parameters for creation that the common denominator seemed the better starting point rather than the grand union.

fabric/patches/2_0004-Deploy-qscc-via-opt_scc-extension-point.patch
g2flyer: yes, this is only needed to exclude qscc which you probably want to unconditionally build but for all the test cases to work you have to exclude qscc's init during testing (via `go test --tags test`).  I would guess the qscc patch will not be upstreamed but still thought it would be a good demonstration on how the framework would be used.   If it is not upstreamed, then you could also drop the `--tags test` from the test command ...

mbrandenburger:
I suggest to merge this and prepare the upstream PR. :)

fabric/patches/2_0003-Mock-SCC-for-opt_scc-extension-point.patch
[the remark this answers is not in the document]
MHBauer: The type assertion should be done as close to the implementation as possible to ensure it does not drift. That way one can still operate on the package as a semi isolated unit.

There's not a code fragment to reference besides the file as it should be an addition.

fabric/patches/9_0001-add-proxy-support-for-docker-in-peer.patch
```diff
@@ -1,7 +1,7 @@
-From ffabe25186d43af74bf8dd351d5237865cadc68c Mon Sep 17 00:00:00 2001
+From 2c0c61f6c35ef76d6425eb680bb8173a59453d40 Mon Sep 17 00:00:00 2001
 From: Michael Steiner <michael.steiner@intel.com>
 Date: Wed, 18 Mar 2020 19:42:55 -0700
-Subject: [PATCH] add proxy support for docker in peer
+Subject: [PATCH 1/1] add proxy support for docker in peer
```
MHBauer, line 4: This patch could probably be reviewed and get in on it's own merits, if there's no alternative provided by upstream.

fabric/patches/2_0001-Extension-point-for-additional-SCCs-opt_scc.patch
```diff
@@ -0,0 +1,172 @@
...
++
+ 	lsccInst := &lscc.SCC{
+ 		BuiltinSCCs: builtinSCCs,
+ 		Support: &lscc.SupportImpl{
+@@ -744,7 +755,7 @@ func serve(args []string) error {
+ 	}
+ 
+ 	// deploy system chaincodes
+-	for _, cc := range []scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC} {
++	for _, cc := range append([]scc.SelfDescribingSysCC{lsccInst, csccInst, qsccInst, lifecycleSCC}, optSCCInsts...) {
```
MHBauer, line 166: I wonder if this could be better factored by creating a slice of "requiredSCCInsts" to act as a counter to "optionalSCCInsts".

Not a fan of "opt" and "insts" but that is your code style.

fabric/patches/9_0001-add-proxy-support-for-docker-in-peer.patch
g2flyer: yes, i have that also in the back of my mind. However, i don't think it is really yet a clean and complete patch, just works for us and once all chaincode would migrate to external builder it also would become obsolete.  That said, i guess i could create a PR for it just to gauge if there is interest and also for the record if somebody else is ever running into the issue and so would have a starting point ....

fabric/patches/2_0001-Extension-point-for-additional-SCCs-opt_scc.patch
g2flyer: The `insts` come from original fabric code in `internal/node/peer/start.go`.  The `opt` i'm also not super enamored with but couldn't come up with something really better ...

PR #427, opened 2020-08-05
Make sure switching SGX_MODE work with just a "make clean"

g2flyer:

A `make clean` didn't clean up some SGX_MODE-sensitive build artifacts which resulted in strange crypto test failures when switching from SIM to HW.  This small patch should fix that ..

This should fix the problem Marcus was facing when testing PR #419 

### DIFF
--- a/common/crypto/Makefile
+++ b/common/crypto/Makefile
@@ -25,4 +25,6 @@ test: build
 	$(MAKE) -C $(BUILD_DIR) test
 
 clean:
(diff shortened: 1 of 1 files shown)


g2flyer:
> maybe remove the changes in `go.mod` for a clean PR :)

aargh, i actually did do a `git show` but the page break was so that i saw the complete intended change but not the additional one. go.sum is now removed ...

PR #428, opened 2020-08-05
enable conditional logging for pdo crypto lib tests

bvavala:
By default, PDO crypto lib tests have logging disabled.

This PR enables logging in these tests based on FABRIC_LOGGING_SPEC.
In particular, they are enabled when FABRIC_LOGGING_SPEC=DEBUG, and disabled otherwise.

### DIFF
--- a/common/crypto/CMakeLists.txt
+++ b/common/crypto/CMakeLists.txt
@@ -21,6 +21,15 @@ FILE(GLOB PROJECT_SOURCES
     "$ENV{FPC_PATH}/common/base64/base64.cpp"
     )
 
+#set debug flags for PDO based on FPC flags
+#(COMMON_CXX_FLAGS is inherited, and used, by the pdo crypto tests)
+if("$ENV{PDO_DEBUG_BUILD}" STREQUAL "1")
+    SET(COMMON_CXX_FLAGS "-Og" "-g" "-DPDO_DEBUG_BUILD=1")
+else()
+    SET(COMMON_CXX_FLAGS "-O2" "-DPDO_DEBUG_BUILD=0")
+endif()
+
+
 ###################################################################################################
 # Untrusted crypto adapt library
 ###################################################################################################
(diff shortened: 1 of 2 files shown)


common/crypto/CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ FILE(GLOB PROJECT_SOURCES
     "$ENV{FPC_PATH}/common/base64/base64.cpp"
     )
 
+#set debug flags for PDO based on FPC flags
+#(COMMON_CXX_FLAGS is inherited, and used, by the pdo crypto tests)
+if("$ENV{FABRIC_LOGGING_SPEC}" STREQUAL "DEBUG")
```
g2flyer, line 26: Given that FABRIC_LOGGING_SPEC can have legitimately much more complex expression when you debug (e.g., to ignore all the gossip messages), I wonder whether use it is the best way.  Why not just use a `PDO_DEBUG_BUILD` env var as PDO itself does (or seems to do quickly doing a `grep -r ...`)? 

In that context as you are already changing debug related cmake code: would be nice to also have an option to cause `cmake` to do a debug rather than release build in this context (and so it is compiled with -g)?  Seems cmake really needs it as cmake variables (`-DCMAKE_BUILD_TYPE=Release or `-DCMAKE_BUILD_TYPE=Debug`) so you cannot do that currently without changing makefile. Maybe easiest would be to define/use $(CMAKE) and have (optional) $(CMAKE_FLAGS)?  That said, the same issue applies also to other places we do cmake, but still might be worth doing?

bvavala: Updated: now you just have to set `PDO_DEBUG_BUILD`. Also, I have included a note in an echo command, right when/if the tests fail.

As for the rest, yes it's desirable, though has low priority.

mbrandenburger: that's good!

PR #429, opened 2020-08-05
Make epid linkable creds also working with docker

### LINKED ISSUE #426: Linkable EPID credentials don't work in anything based on docker/docker-compose, e.g., demo
**Describe the bug**

Whether epid linkable or unlinkable credentials are used are dependent on an `FPC_ATTESTATION_TYPE` env-variable.  
While we handle that properly for integration tests, we fail to propagate the variable, if defined, to the docker images which are all will be built with only support for unlinkable (the fallback if `FPC_ATTESTATION_TYPE` is undefined).

**To Reproduce**
- setup linkable credentials in `config/ias` and define `export FPC_ATTESTATION_TYPE=epid_linkable` in `./config.override.mk`
- do a `make`
- result: integration tests will pass but demo will fail

**Expected behavior**
- demo should succeed

**Log-filesScreenshots**

**Additional context**

g2flyer:
Subject says it all, see also #426.

Closes #426

A few notes:
- removed the definition of the default of `FPC_ATTESTATION_TYPE` and moved it to `config.mk` so this tweakable has some docu and a default definition at a central place in the code.
- as this was really part of SGX config, i also consolidated the cmake handling in `cmake/ConfigSGX.cmake`
- as peer and boilerplate images are "tainted" with `FPC_ATTESTATION_TYPE`, i correspondingly reflect that also in the image name ...

_PS: As Bruno correctly points out in #431, the compiled-in attestation-type is in a way the wrong approach and if we are going to do what #431 proposes, then this PR would become obsolete (and the docker-changes would have to be reverted).  So somewhat debatable whether it is worth merging this PR? That said, the bug can lead to confusion, the PR is here and reverting it as part of #431 should be easy?_

Different docker image name is the only user-facing visible change (except that it should now also work with linkable creds ;-)) and it should be backwards compatible ...

### DIFF
--- a/utils/docker/dev_peer_cc-builder/Dockerfile
+++ b/utils/docker/dev_peer_cc-builder/Dockerfile
@@ -1,31 +1,38 @@
 # Copyright IBM Corp. All Rights Reserved.
+# Copyright 2020 Intel Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
 
 # Description:
-#   Sets up fabric/fpc source environment and builds various derivatives in different stages
+#   Sets up fabric/fpc source environment and builds, in different stages, three different
+#   containers
+#   - peer: a container with an fpc-enhanced fabric peer
+#   - cc-builder: an environment to build fpc chaincode enclave.so
+#   - dev: a fpc-enabled interactive development environment
 #
-#  Configuration (build) paramaters
-#  - fpc image version:		FPC_VERSION (default: latest)
-#  - fabric repo:  		FABRIC_REPO (default: https://github.com/hyperledger/fabric.git)
-#  - fabric branch: 		FABRIC_VERSION (default: 2.2.0)
-#  - fpc repo:		 	FPC_REPO_URL (default: https://github.com/hyperledger-labs/fabric-private-chaincode.git)
-#  - fpc branch/tag/commit	FPC_REPO_BRANCH_TAG_OR_COMMIT (default: master)
-#  - git user:			GIT_USER_NAME (default: tester)
-#  - git user's email:		GIT_USER_EMAIL (default: tester@fpc)
-#  - sgxmode: 			SGX_MODE (default: SIM)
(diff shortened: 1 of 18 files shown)


utils/docker/dev_peer_cc-builder/Dockerfile
```diff
@@ -151,6 +158,8 @@ FROM common as cc-builder
 # import global vars
 ARG SGX_MODE
 ENV SGX_MODE=${SGX_MODE}
+# Note: the enclave.so does not depend on FPC_ATTESTATION_TYPE, so we only have to specialize on SGX_MODE
```
bvavala, line 161: Would you double check this please?

I think it's actually the opposite.
cc-builder builds the sdk, including sgxcclib in `ecc_enclave`, and it will build it with epid-unlinkable by default.
Then, cc-builder and its artifacts are used by the demo to build the enclave.so.
However, if I recall correctly, the demo skips the ercc stuff, so it does not fail.

g2flyer: cc-builder definitely builds a attestation-type dependent sdk. But the attestation part should be all only in the untrusted part, not in the trusted code (i.e., it is backed statically into the invocation of `sgx_get_quote` in common/sgxcclib/common-sgxcclib.c`).  So it shouldn't `enclave.so`, the only build artifact we use `cc-builder` for.  I've also tested with both linkable and unlinkable with `make clobber all` (which runs real attestation in the non-interactive auction demo including cc-builder) and even tested with wrong credentials which also failed as expected.   So not sure what i've overlooked?   But maybe i could be a bit more explicit in the Note: for the reasoning?

bvavala: That's fine, the "cc-builder" created some confusion.
Summarizing: just as we do for any FPC chaincode (see the examples), we generate the `enclave.so` and we use the pre-built `ecc` and `libsgxcc`. In this particular case, cc-builder creates the `enclave.so`.

PR #433, opened 2020-08-13
Common attestation api

bvavala:
This PR partially addresses #412 , describing the APIs for producing and verifying an attestation.


docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+int init_attestation(
+    uint8_t* params,
+    uint32_t params_length);
+
+```
+```
+int get_attestation(
+    uint8_t statement[32],
```
g2flyer, line 18: I guess you wanted to keep the API not FPC specific so it could more widely reused (e.g., in PDO)? This makes sense.  I wonder, though, whether you might not make it an arbitary-length byte array and then hash it internally to 32 bytes). Otherwise, you force the caller in most cases (e.g., fpc) to also have to do crypto on top.  I think it would be nicer if the caller only would have to do serialization?

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+This document defines the interface for producing and verifying an attestation.
+
+
+## Producing an attestation
+
+This interface is only available inside an enclave.
+
+```
+int init_attestation(
+    uint8_t* params,
```
g2flyer, line 12: What did you have in mind with these params?  Is this how you would pass spid and corresponding credentials to the library?

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+    [out, size=egid_len] uint8_t *egid, uint32_t egid_len);
+
+void ocall_get_quote(
+    [in, size=spid_len] uint8_t *spid, uint32_t spid_len,
+    [in, size=report_len] uint8_t *report, uint32_t report_len,
+    [out, size=max_quote_len] uint8_t *quote, uint32_t max_quote_len,
+    [out] uint32_t *actual_quote_len);
+```
+
+3. It must to be sent to IAS, which provides a publicly verifiable report. This step thus involves contacting and authenticating with a web service using the ISV's credentials. For this reason, its completion is delegated to a different entity.
```
g2flyer, line 43: But still would involve a, yet to be defined, ocall? 
BTW: thinking of how to get the credentials, i actually realized that the current way of passing it via decorators should work in this case (with all its warts, of course, but at least one viable way to manage the credentials)

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+This interface is only available inside an enclave.
+
+```
+int init_attestation(
+    uint8_t* params,
+    uint32_t params_length);
+
+```
+```
+int get_attestation(
```
mbrandenburger, line 17: Should we specify an error in case `init_attetstation` has not been called before?

docs/design/fabric-v2+/attestation-interface.md
mbrandenburger, line 18: Let's add a comment here what statement is. I guess this is statement ends up in the userdata of an sgx quote.

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+    uint8_t statement[32],
+    uint8_t* attestation,
+    uint32_t attestation_length);
```
mbrandenburger, line 20: attestation is what we call attestation_evidence in the uml diagrams right? just to clarify

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+
+This interface is available both inside and outside of an enclave.
+
+```
+int verify_attestation(
+    uint8_t* attestation,
+    uint32_t attestation_length,
+    uint32_t expected_statement[32],
+    uint8_t* expected_code_id,
+    uint32_t expected_code_id_length);
```
mbrandenburger, line 56: I guess code_id corresponds to mrenclave? If so I suggest to also name it `expected_mrenclave` to better align the uml diagrams and other interfaces.

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,63 @@
...
+
+3. It must to be sent to IAS, which provides a publicly verifiable report. This step thus involves contacting and authenticating with a web service using the ISV's credentials. For this reason, its completion is delegated to a different entity.
+
+
+## Verifying an attestation
+
+This interface is available both inside and outside of an enclave.
+
+```
+int verify_attestation(
```
mbrandenburger, line 51: How is the verification result decoded? We should define a set of error codes this method (and also others) can return. At least basic stuff like, VERIFICATION_SUCCEEEEDS, FAILED

g2flyer:
> Should we also defined here a more high-lvl API that is directly used by FPC components and to hide the complexity of bringing all the data in the right that it can be used with this low-level API?

Not sure whether we need a higher-level API: if we define the protobuf for AttestedData as from UML diagram and do the small change i suggested above (e.g., accept arbitrary lenght message and then hash inside module), then you just have to serialize before calling: not sure whether this warrants a separate API (not that it doesn't hurt, though).?

g2flyer:
_as discussed today, below my proposal .._

# Public API

_Note: Below is original proposal from this PR with a few minor changes:_
- _to match uml protocols i t replaced term attestation by attestation_evidence for the artifacts created and verified._
- _allow arbitrary length statements and put any potentially necessary hashing to 32 bytes inside library to simplify caller while trying not to add FPC-specifics  
  (ie., in our case ecc only has to do serialization of protobuf but no crypto.)
  This might also be useful to abstract away from SGX,_
- _added function to return attestation type as we seem to use that in some protobuf._
- _assuming that for near future it will be SGX/Intel specific, replaced type of code_id with mrenclave_t (but minor syntactic sugar, original version is also fine if we want to be more general.)_


```c++
// initialize attestation module
// - primarily used for ecc
// - params would be encoded in some attesation-specific format (e.g., json?)
// - for epid, this would be passing SPID, API key and linkable vs unlinkable,
//   IAS root PK is assumed to be baked into library
int init_attestation(
    uint8_t* params,
    uint32_t params_length);

// return the supported attestation type
// - callable in ecc, ecc_enclave and ercc
typedef enum {EPID_LINKABLE, EPID_UNLINKABLE, DCAP} attestation_type_t;
attestation_type_t get_attestation_type();


// retrieve attestation evidence over statement
// - called inside ecc_enclave
int get_attestation_evidence(
    uint8_t* statement,
    uint32_t statement_lenghth,
    uint8_t* attestation_evidence,
    uint32_t attestation_evidence_length);

// verify attestation evidence matches expected statement
// - called by ercc (via a cgo-wrapper)
int verify_attestation_evidence(
    uint8_t* attestation_evidence,
    uint32_t attestation_evidence_length,
    uint8_t* expected_statement,
    uint8_t* expected_statement_length,
    mrenclave_t expected_code);
```

# EDL

##  ECC <-> ECC_Enclave

_Note: Below is original proposal with additional ocall for retrieving IAS verification report_

```
	untrusted {

		void ocall_init_quote(
		    [out, size=target_len] uint8_t *target, uint32_t target_len,
		    [out, size=egid_len] uint8_t *egid, uint32_t egid_len);

		void ocall_get_quote(
		    [in, size=spid_len] uint8_t *spid, uint32_t spid_len,
		    [in, size=report_len] uint8_t *report, uint32_t report_len,
		    [out, size=max_quote_len] uint8_t *quote, uint32_t max_quote_len,
		    [out] uint32_t *actual_quote_len);

		void ocall_get_ias_report(
		    [in, size=quote_len] uint8_t *quote, uint32_t quote_len,
		    [out, size=max_verification_report_len] uint8_t *verification_report, uint32_t max_verification_report_len,
		    [out] uint32_t *actual_verification_report_len);
	}

```

# Discussion Pro's and Con's

- most minimal public interface
- matches the UML diagrams
- should map also nicely to DCAP (of course with other params to init_attestation and certainly not implemented for MVP)
- matches the IAS credential deployment model as currently done in FPC as well as PDO, i.e., credentials are associated with infrastructure. (Note: For DCAP it is inherently infrastructure-based  as the configs are associated with the platform.)
- requires network connectivity at peer/external builder but this is already a requirement right now in fabric to build chaincodes and retrieve dependent packages. Note also that external builder will control docker invocation once we move beyond the current `execute-chaincode-directly-on-host` model

docs/design/fabric-v2+/attestation-interface.md
bvavala: > I guess you wanted to keep the API not FPC specific so it could more widely reused (e.g., in PDO)?

Desirable, but that's secondary. The primary reason is that no FPC-specific thing should belong to a "common attestation" api, but rather they should be implemented above it.

> whether you might not make it an arbitary-length byte array and then hash it internally to 32 bytes?

First, I would argue that it is an implementation detail -- we could also make it optional.
Second, the rationale for not doing it is to avoid forcing a hash. The caller could choose to simply put an arbitrary 32-byte array.

> Otherwise, you force the caller in most cases (e.g., fpc) to also have to do crypto on top

Yes, that's the trade-off. Arguably a hash function is usually available.

docs/design/fabric-v2+/attestation-interface.md
bvavala: > What did you have in mind with these params? Is this how you would pass spid and corresponding credentials to the library?

I would expect this to be a json blob (or similar). So it would contain spid, and maybe sigrl and api key if necessary.

> given that these params are originally on the untrusted side and also are needed there given the current epid sgx apis, probably this function is better done in fpc_stub rather than (fpc_stub_)enclave

They are on the untrusted side, which means either at (i) the peer or at (ii) the admin.

In case (i):
* these parameters should be defined in the peer configuration (core.yaml), unless we use admin calls that terminate in the Go chaincode. If, for any reason, the parameters should be changed, the peer should be reconfigured (maybe restarting it, or through admin calls).
* putting the init in untrusted side means that the `get_attestation` would not work without calling first the init in an abstraction below. For this reason, it does not seem appropriate to separate the two functions.

In case (ii):
* the parameters are kept by the admin, and can be provided through the `createenclave` which terminates inside the enclave. Three benefits here: 1) we already have the `createenclave` in place; 2) we can remove configuration parameters from the core.yaml; 3) we keep init and get_attestation side by side, making the api self-contained.

docs/design/fabric-v2+/attestation-interface.md
bvavala: If the attestation output has to be publicly verifiable, then yes: we need another ocall and the IAS credentials passed somehow.

If the attestation output can be IAS-verifiable (i.e, the spid owner has to convert this into a publicly verifiable report by contacting IAS), then no: we don't need ocall and credentials; we delegate this to the admin.

The second case is viable since the admin is already the recipient of the `createenclave` response.

docs/design/fabric-v2+/attestation-interface.md
bvavala: In theory, yes.
In practice, it could be the quote to be converted in a verification report (which is the attestation evidence).

docs/design/fabric-v2+/attestation-interface.md
bvavala: `mrenclave` is SGX-specific, while the _code identity_ concept is generally accepted in the community.
Given that this is a _common_ API, I would leave it as it is. No strong opinion anyway.

docs/design/fabric-v2+/attestation-interface.md
bvavala: Yes, I have added a commit.

docs/design/fabric-v2+/attestation-interface.md
bvavala: Yes, I have added a comment.

docs/design/fabric-v2+/attestation-interface.md
bvavala: I have added a comment. Even more basic: return `0` on error, otherwise it is successful.

docs/design/fabric-v2+/attestation-interface.md
g2flyer: hmm, given that this API is an abstraction and, from discussion around `mrenclave_t`, you also aim to be independent not only of EPID and DCAP but even of SGX itself, then for this to work you will need an additional API function which does the conversion?  Otherwise you completely broke the abstraction? Or put differently(i.e., as-is), everybody looking at the public API and seeing a generation and a verification function with matching parameter names and no other function will assume to be output of `get_attestation` to be the input to `verify_attestation`.  This interpretation would also be re-enforced by looking a the current UMLs where there are evidence bytes passed from fpc stub to ercc with no documented manipulation?

docs/design/fabric-v2+/attestation-interface.md
g2flyer: Generalizing beyond SGX definitely would be nice. If we aim for it,it might be worth-while though to cross-check with some other types, e.g., i'm pretty sure forcing the statement to be exactly 32 bytes would not match the native format of everybody, in which case passing an arbitrary-length statement (and normalizing (via hashing) to the native format "behind" the api) would seem more general?

The main trade-off here seems to be around generality vs type-safety and code readability? With `mrenclave_t` there would be no need of doing type-casts and would also reduce the number of parameter validation steps you have to do? But obviously it would be SGX centric ...

docs/design/fabric-v2+/attestation-interface.md
g2flyer: Related more general question as i was thinking of return codes also in context of #434 (but then punted so far as i was not 100% sure).  Is our general strategy for return values in C:
- return 1/non-zero for true and 0 for false for functions evaluating a predictate
- return 0 for success and non-zero for errors for any other functions
Glancing through some code it seems that's more or less what we have done, but just cross-checking....

bvavala:
**Proposal**
1. Two APIs for producing an attestation, only available inside the enclave
```
int init_attestation(
    uint8_t* params,
    uint32_t params_length);

int get_attestation(
    uint8_t statement[32],
    uint8_t* attestation,
    uint32_t attestation_length);
```
2. One API for verifying an attestation, available both inside and outside of an enclave
```
int verify_attestation(
    uint8_t* attestation,
    uint32_t attestation_length,
    uint32_t expected_statement[32],
    uint8_t* expected_code_id,
    uint32_t expected_code_id_length);
```
3. requirement of no Internet connection to web services from chaincode enclave, shim, external builder, peer, etc.
4. requirement of no attestation-related configuration (in a) yaml file, or in a separate chaincode invocation, or anywhere in external builder, or packages
5. implement any ocall that is necessary to support such API (e.g., ocall_init_quote, ocall_get_quote, etc.). These do not belong to the presented proposal.

Some important benefits:
1. Fabric already requires (and might impose through containerization or other means) chaincode invocations not to result in the use of external (possibly internet) services. This proposal does not require any.
2. the peer (or chaincode, or external builder) does not have to be configured with information like SPID, API-key. The storage/usage of such info is delegated to the admin of the organization/ISV. Hence, less configuration data/logic.
3. in the FPC flow, any data required for init can be sent by the admin through the invocation of `createenclave` command. Hence all the infrastructure is already available for this to work.

One non-drawback:  
1. in EPID, the returned attestation is an IAS-verifiable quote. This must be converted to a verification report in order to be publicly verifiable. As the requestor of such attestation is the admin, the admin already knows that this operation is needed. Also the admin has the SPID and API-key to perform the translation operation with IAS, on behalf of the organization he represents.

Minor benefits:
1. the 32-bytes statement does not impose the the use of hash function, leaving its definition entirely to the caller. Hence the caller may elect to use the output of a hash function, or arbitrary data.

docs/design/fabric-v2+/attestation-interface.md
mbrandenburger: sounds good, even though I will be confused all the time and mix this stuff. maybe better use boolean return value to clearly differentiate between an error/non-error return value vs an evaluating function.

mbrandenburger:
@bvavala can you please a view details on the API and then we can close and merge this one.
Let's also add some more comments to the API, for instance, to `get_attestation` that this may return either a "privately"- or "publicly"- verifable attestation. And `verify_attestation` only accepts "publicly"-verifiable inputs, so the user of the API has to perform the transformatiton step. To check whether the returned attestation evidence is private or public some metadata would be nice, even though the caller should know as he/she provided the input for `init`. So maybe adding some words here to describe that the input of `init` will determine the output type of `get_attestation`?!

docs/design/fabric-v2+/attestation-interface.md
bvavala: changed to arbitrary length statement

docs/design/fabric-v2+/attestation-interface.md
bvavala: changed to boolean -- [the type is supported in C as of C99](https://en.cppreference.com/w/c/types/boolean)

bvavala:
> @bvavala can you please a view details on the API and then we can close and merge this one.
> Let's also add some more comments to the API, for instance, to `get_attestation` that this may return either a "privately"- or "publicly"- verifable attestation. And `verify_attestation` only accepts "publicly"-verifiable inputs, so the user of the API has to perform the transformatiton step. To check whether the returned attestation evidence is private or public some metadata would be nice, even though the caller should know as he/she provided the input for `init`. So maybe adding some words here to describe that the input of `init` will determine the output type of `get_attestation`?!

I have added a more detailed comment about that in the EPID details sections.

docs/design/fabric-v2+/interfaces.md
```diff
@@ -131,8 +131,9 @@ message Attested_Data {
 }
 
 message Credentials {
-        Attested_Data data    = 1;
-        bytes evidence_bytes = 2;       // serialized attestation evidence
+        Attested_Data data          = 1;
+        optional bytes attestation  = 2;    // optional attestation/quote
```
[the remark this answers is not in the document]
bvavala: not much to debate here!

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -38,7 +38,9 @@ note right Admin
       of FPC chaincode.
 end note
 
-Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id)
+Admin -> Admin: Attestation_Params <- <SPID>
```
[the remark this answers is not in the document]
bvavala: that's the beauty of specifying params in that way: you update it however you want.

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,95 @@
+# Attestation Interface
```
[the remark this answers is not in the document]
bvavala: done

docs/design/fabric-v2+/attestation-interface.md
```diff
@@ -0,0 +1,95 @@
...
+void ocall_init_quote(
+    [out, size=target_len] uint8_t *target, uint32_t target_len,
+    [out, size=egid_len] uint8_t *egid, uint32_t egid_len);
+
+void ocall_get_quote(
+    [in, size=spid_len] uint8_t *spid, uint32_t spid_len,
+    [in, size=report_len] uint8_t *report, uint32_t report_len,
+    [out, size=max_quote_len] uint8_t *quote, uint32_t max_quote_len,
+    [out] uint32_t *actual_quote_len);
+```
```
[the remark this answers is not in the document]
bvavala: It was meant to be solved later, but it's one-line now. done!

docs/design/fabric-v2+/attestation-interface.md
bvavala: > According to interfaces.md -- which as mentioned yesterday has not everything reflected yet in UML diagrams --, the createenclave ends in ecc which then calls ecc_enclave via init_enclave. So we would have to add the attestation parameter also there.

I would leave as it is now. Reason is: too many changes going on; once these stabilize, we can make them uniform.

g2flyer: hmm, `interfaces.md` should be stable now and should have been the reference regarding these ecc/ecc-enclave interfaces? So i don't see when would be a better time to change it than now?

Issue #361, opened 2020-05-19
Do TLCC cc2scc calls require reflection in r/w set of transactions?


g2flyer:
**Description**

Currently calls to TLCC from ECC (or potentially in the future for ERCC) are not reflected explicitly in an endorsement as other cc2(s)cc calls are.   We should investigate whether this is secure and/or safe from conflicts and otherwise might have to extend our protocols.  Probably also requires a bit more in-deptch study how fabric handles that exactly for other cc2scc calls.

**Link to feature or bug**
<--!
  Please add link to any applicable feature or bug.
-->

**Link to dependent issues**
<--! 
  Please link to any other work-item this issue depends on.
-->

mbrandenburger:
With Fabric it is possible to call another chaincode (chaincodeB) using chaincode2chaincode calls while your chaincode (chaincodeA) is executed. The interface is the same as a "normal" chaincode invocation. You send an operation with arguments and as a result it returns a response. If the called chaincode accesses it's state via getState calls, these calls and all accessed keys are presented in the read/write set.

For example:
```
<TxReadWriteSet>
  <NsReadWriteSet name="chaincodeA">
    <read-set>
      <read key="K1", version="1">
    </read-set>
    <write-set>
      <write key="K1", value="V1"
    </write-set>
  </NsReadWriteSet>
  <NsReadWriteSet name="chaincodeA">
    <read-set>
      <read key="K1", version="1">
    </read-set>
  </NsReadWriteSet>
<TxReadWriteSet>
```

In the case of TLCC, it does not access the state using getState ops but instead calls into the enclave and produces some integrity metadata that allows the FPC chaincode enclave to verify that it received "correct" state when ecc called getState.

Is it bad the the integrity metadata will not show up in the read/writeset? Since the implementation of the integrity metadata is a cmac that requires the corresponding key to verify it, even if we add it to the transaction proposal response, we cannot use to verify as the key is shared only between tlcc_enclave and ecc_enclave. So I guess we dont have an issue here.

mbrandenburger:
The other case where we may want to leverage cc2cc call is to let ercc retrieve the chaincode definition, in particular the MRENCLAVE, for a FPC chaincode. As ercc cannot directly access it from the peer when verifying an attestation report, we need to get it from somewhere. There are two options:

1) client provides it when invoking ERCC. This is not a good idea as every endorsing peer should verify an attestation report using it's own view of the chaincode definition.

[...]

PR #435, opened 2020-08-13
Ledger strategy

### LINKED ISSUE #402: Ledger Enclave Strategy

mbrandenburger:

This document defines the TLCC (trusted ledger) strategy.

Fixes #402 

```

### DIFF
--- a/docs/design/fabric-v2+/trusted_ledger.md
+++ b/docs/design/fabric-v2+/trusted_ledger.md
@@ -6,94 +6,21 @@ and enable ECC to verify ledger state receiveed from the (untrusted) peer.
 ## Requirements
 
 - ECC can query integrity metadata to validate ledger state received from the peer.
+- TLCC must maintain a view on the ledger
 - TLCC must be synronized with peer's ledger state
 - Secure (authenticated) channel between ECC and TLCC to query metadata;
-- ECC must perform queries on a snapshot during a single transaction invocation
-- TLCC must provide identity validation (orderer, msp, etc...)
+- ECC tx must perform read/write on a stable view of the ledger state
+    - (snapshot during a single transaction invocation)
+- TLCC must detect read/write inconsistency
+- TLCC must be able to perform identity validation (orderer, msp, etc...)
 - TLCC must provide channel metadata (chaincode definition)
 - Maintaining FPC chaincode state only is sufficient
 - Validate "normal" Enclave Registry (ERCC) transactions; however, ERCC comes with "hard-coded" endorsement policy
 
 
-
-## Approach
-
-Restrict Fabric functionality
-
-- Reduces validation logic to a minimum
-- Minimizes catch-up play with Fabric validation changes
-- Relax the notion of endorsement policy; 
-- Simplify chaincode versioning 
-- No support for custom endorsing/validation plugins for non-FPC chaincodes
-- Support for default MSP only
-- Any transactions/blocks with unsupported features are ignored/aborted
-
-Defined process to:
-- keep redundant validation code (in stock peer and TLCC) in sync
-- keep code relation traceable and changes trackable
-- consistent naming and code structure as much as possible
-- bonus: automated code-changes notification; notifies whenever Fabric code changes and FPC must be updates.
-
-### Non-features
-
-- No state-based endorsement
-- No custom MSP (no idemix)
-
-
-## Design
-
-Channel configuration:
-- Parse channel definition
-    - parse consensus definition
-- validate block signatures
-- Maintain msp metadata for signature validation
-- Config blocks are validated by the ordering service ([see code](https://github.com/hyperledger/fabric/blob/f27912f2f419c3b35d2c1df120f19585815eceb0/orderer/common/msgprocessor/standardchannel.go#L131))
-
-- parse Lifecycle policy 
-- validate Lifecycle policy
-    We only support: a majority of Orgs (See [Link](https://hyperledger-fabric.readthedocs.io/en/release-2.2/chaincode_lifecycle.html#install-and-define-a-chaincode))
-
-Access control:
-- check org is "writer"
-
-MSP:
-- Implement X509-based MSP
-
-Endorsing:
-- Phase1: Designated enclave only
-- Phase2: Any enclave that runs a particular FPC chaincode
-
-Versioning:
-- single autonomously monotonously increasing version number??
-
-Non-FPC Validation:
-- Transaction submitter identity validation
-    - submitter satisfies channel's writes policy
-- ERCC endorsement signatures verification
-- ERCC endorsement policy validation (same as lifecycle)
-    - Restrict to ERCC namespace only
-
-
-
-FPC Validation:
-- Introduce FPC transaction type (similar to introduce FPC namespace) and create
-a dedicated FPC tx processor; (removes the need of custom validation plugins and interference with existing Fabric validation logic; and also gives more freedom to FPC validation logic as no it not longer bound to the structure and format of endorsement transaction).
-
-- Support for (subset of) endorsing policies
-    - FPC Chaincode (see above)
-
-- Parse chaincode definitions
-- Transaction submitter identity validation
-    - submitter satisfies channel's writes policy
-
-- FPC endorsement policy validation
-    - Support only: ANY
-- FPC endorsement signatures
-
-### Fabric high-level validation 
+## Fabric high-level validation 
 
 TODO pseudo code here
-
 
 Config validation: [source](https://github.com/hyperledger/fabric/blob/f27912f2f419c3b35d2c1df120f19585815eceb0/common/configtx/validator.go#L163)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/design/fabric-v2+/trusted_ledger.md [hunk 1]
```diff
@@ -0,0 +1,162 @@
...
+- consistent naming and code structure as much as possible
+- bonus: 
+    - automated code-changes notification; notifies whenever relevant Fabric code changes and might FPC must be updates.
+
+### Non-features
+
+- No state-based endorsement
+- No custom MSP (no idemix)
+- No support for custom endorsing/validation plugins for non-FPC and FPC chaincodes
+- Authentication and decorators
```
g2flyer, line 72: i would explicitly say here "authentication _filters_" not to confuse it with the reader/writer authentication discussed later.

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+- TODO check for more non-features in RFC
+
+## Design
+
+### Chaincode execution support
+    - secure channel
+    - validate proposal
+        - check org is "write"
+        - replay protection
```
g2flyer, line 81: the view consistency/snapshot probably also belongs here?  

In that context, would be interesting to know how fabric currently does it: do they serialize update and endorsement or do they maintain different ledger states/views concurrently? the former is of course the easiest to implement (though requires an explicit notification by ecc that a tx is finished) but also means that a slow tx can hold of others (not that this is probably immediately the biggest concern for us :-))

(this in turn might ask for adding eventually the endorsement equivalent of the "Fabric high-level validation" section above?)

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+    - parse consensus definition
+- parse channel reader/writers
+- validate block signatures (Note that this check implies correctness as onfig blocks are validated by the ordering service ([see code](https://github.com/hyperledger/fabric/blob/f27912f2f419c3b35d2c1df120f19585815eceb0/orderer/common/msgprocessor/standardchannel.go#L131)))
+- Maintain msp metadata for signature validation
+- parse Lifecycle policy 
+    - validate Lifecycle policy
+    We only support: a majority of Org.member (See [Link](https://hyperledger-fabric.readthedocs.io/en/release-2.2/chaincode_lifecycle.html#install-and-define-a-chaincode))
+
+Access control:
+- delegate this "reader/writer/admin" check to admins;
```
g2flyer, line 96: We wouldn't really delegate the _(authorization) check_ to admins as this text seems to imply?  we would delegate to admins the  _policy check_ that the authorization policy does not allow ruling out an endorsement which otherwise would be a valid endorsement regarding our restrictions as mentioned above?

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+- TLCC only verifies endorsement policies and thereby implicitly the "writers" check is performed by the endorsers
+
+MSP:
+- One Org - One MSP mapping - One root cert
+- Implement X509-based MSP
+    - Restrict:
+        - no intermediate
+        - no CRLs
+        - only support member role
+    - any certificate will match to role member
```
g2flyer, line 106:
```suggestion
            - any certificate will match to role member
```

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+
+Endorsing:
+- Phase1: Designated enclave only
+- Phase2: Any enclave that runs a particular FPC chaincode
+
+Versioning:
+- single autonomously monotonously increasing version number??
```
g2flyer, line 113: monotonically increasing version numbers for us will be a bit a challenge with it being MRENCLAVE ... ;-)    [different story of course for ercc, though not sure whether for us the version really matters for ercc (compared to sequence numbers) as we look only at meta data changes and on that level version doesn't really have any effect? It is only relevant during the actual execution of the endorsement (which we delegate to normal peers for ercc)?

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+Non-FPC Validation:
+- Transaction submitter identity validation
+    - submitter satisfies channel's writes policy
+- ERCC endorsement signatures verification
+- ERCC endorsement policy validation
+    - we only support (explicityly) Majority endorsement policy
+    - Restrict to ERCC namespace only
+
+FPC Validation:
+- Introduce FPC transaction type (similar to introduce FPC namespace) and create
```
g2flyer, line 124: seems to me we should split this into 2: in the current design we essentially define the type somewhat implicitly by the presence of the fpc custom validation plugin.  Do we really want to tackle changes of fabric core in phase 1?  If it is possible with minimal changes it certainly could have various advantages but it could also be a rathole of changes also on client side etc?

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,162 @@
...
+
+- Support for (subset of) endorsing policies
+    - FPC Chaincode (see above)
+
+- Parse chaincode definitions
+- Transaction submitter identity validation
+    - submitter satisfies channel's writes policy
+
+- FPC endorsement policy validation
+    - Support only: ANY
```
g2flyer, line 135: Not sure what you mean by ANY. in https://github.com/hyperledger-labs/fabric-private-chaincode/blob/master/docs/design/fabric-v2%2B/fpc-management.md we say an OR with a single org (and from above, we would even further restrict it that the role must be `member`; in that document we would also allow other roles ...)

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,171 @@
...
+Restrict Fabric functionality:
+
+- keep Fabric validation changes in sync with TLCC
+- Reduces validation logic to a minimum
+    - minimizes catch-up with Fabric updates
+- Restrict the notion of endorsement policy; 
+- Simplify chaincode versioning (TODO)
+    - No chaincode updates for now
+- Support for default MSP only
+- Maintaining FPC chaincode state only is sufficient
```
g2flyer, line 57: not sure what exactly you mean here ..

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,171 @@
...
+- Validate "normal" Enclave Registry (ERCC) transactions; however, ERCC comes with "hard-coded" endorsement policy
+
+
+Any transactions/blocks with unsupported features are ignored/aborted
```
g2flyer, line 61: somewhere we definitely have to be more explicit where we abort and, in particular, where we (safely can) ignore. My feeling is that by default we should abort and ignore only for explicit cases where we know it's safe to ignore ..

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,171 @@
...
+
+### short term
+See approach above
+
+We restrict the supported endorsement policies for lifecycle, ERCC, and FPC chaincodes.
+
+### mid/long term
+- Introduce FPC transaction type (similar to introduce FPC namespace) and create
+a dedicated FPC tx processor; (removes the need of custom validation plugins and interference with existing Fabric validation logic; and also gives more freedom to FPC validation logic as no it not longer bound to the structure and format of endorsement transaction).
+
```
g2flyer, line 155: Related to above i guess is also to make tlcc maintain _all_ state, not only meta-data (thereby also removing any issue of synchronicity between peer and tlcc (at least as long as we do not go to cc2cc between fpc and non-fpc chaincode ...)

docs/design/fabric-v2+/trusted_ledger.md
```diff
@@ -0,0 +1,171 @@
...
+Re-use Fabric code components inside trusted ledger enclave. This requires further development on go-support for SGX. Although some PoC based on graphene for go are already available but seems not be stable yet.
+
+We may extend support for more enhanced endorsement policies in the future.
+
+### QA process
+
+## Implementation
+
+- nanopb to parse fabric protos (alternatively we could try to use real proto for c)
+- data state : leveldb
```
g2flyer, line 165: by data you mean actual ledger data and meta-data? 
also, adding persistance to tlcc is more mid-term rather than short-term? Or do you see a low-hanging fruit where we get that for (almost) "free"?  (If we restrict that state must fit into EPC and we write and read a snapshot in one go , we actually would not really have to do merkle tree and alike and a simple HMAC  (based on some seal-key) protected snapshot should be enough?)

g2flyer:
BTW: view consistency probably should also be addressed here? See also corresponding comments added to field `tx_context` in `message TLCCRequest` added as part of #434. Probably #361 is also somewhat relevant regarding the question whether view consistency is required just to guarantee progress or whether there are also integrity issues if we do not enforce view consistency and (more importantly) any confidentiality issues which could exploit integraty failures (e.g., integrity failures might be caught by peers when processing endorsements but it still could be that integrity issues during endorsement could lead to confidentiality breaches).   Related to this it probably would also be useful to know how fabric deals with view consistency (e.g., do peers ensure that chaincode have consistent views and, if so, how)

bvavala:
Outdated and needs further work.
Will be reopened as the discussion restarts in the future.

PR #438, opened 2020-08-31
Fix broken dev apt add pkg

mbrandenburger:
A naming inconsistency and missing ARG caused not installing any
additional packages as set with DOCKER_DEV_IMAGE_APT_ADD_PKGS in
config.override.mk


mbrandenburger:
> The import was definitely missing, good catch. The rename was imho a bit gratuitous and hides a bit the key part of the patch but if you like it better this way, i'm fine ...

It's not that I like the name better but otherwise it won't work.

PR #440, opened 2020-09-02
New protos

mbrandenburger:

This PR adds the new protos as specified in #420.
- Refactors existing protos (move protos to protos/)
- Set c-protos output to common/protos
- Set go-protos to internal/protos

Fixes #

As submodule has moved, you will have to do a `git submodule update --init` after checkout ...

```

### CODE REVIEW

File: common/protos/fpc.proto [hunk 1]
```diff
@@ -0,0 +1,74 @@
+syntax = "proto3";
```
g2flyer, line 1: quick thought: shouldn't we create here an fpc sub-directory and add our own stuff there? At least i would have expected that we also move the current google/fabric protobufs into common (including a common build procedure) and then we more or less would need an fpc subdir?

mbrandenburger: I was thinking the same! so we do it!

protos/fpc/attestation.proto
```diff
@@ -0,0 +1,21 @@
+// Copyright IBM Corp. All Rights Reserved.
+// Copyright 2020 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package attestation;
+option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos";
```
g2flyer, line 9: given that we modularlize attestation (and tl_session), i wonder whether we might not put the protobufs into module-specific packages?
```suggestion
option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos/attestation";
```

protos/fpc/tl_session.proto
```diff
@@ -0,0 +1,87 @@
...
+// Copyright 2020 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package tl_session;
+option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos";
```
g2flyer, line 11: See above comment on attesation
```suggestion
option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos/tl_session";
```

protos/fpc/fpc.proto
```diff
@@ -0,0 +1,36 @@
...
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package fpc;
+option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos";
+
+message CC_Parameters {
+    string chaincode_id   = 1;      // chaincode name
+    string version        = 2;      // chaincode version, this is the expected mrenclave
```
g2flyer, line 15: We probably should mention here (and for tlcc_mrenclave) what encoding happens (or just link to fpc-management.md for version and mention that tlcc_mrenclave is handled the same way? Couldn't find a definition on encoding of tlcc_mrenclave in the .md ...)

This more or less applies also to most other `bytes` where the encoding is not necessarily clear. (As discussed on phone, it might be worthwhile also to define them as types (sub-messages) if there is a good way to augment the protoc/nanopb-generated code with corresponding encoding/decoding functions ...

protos/fpc/fpc.proto
```diff
@@ -0,0 +1,36 @@
...
+    string sequence       = 3;      // chaincode sequence number
+    string channel_id     = 4;      // channel id
+}
+
+message Host_Parameters {
+    bytes peer_identity = 1;        // peer that hosts the enclave
```
g2flyer, line 21: what is the encoding here? Will this be a serialized msp.SerializedIdentity from Fabric?

protos/fpc/fpc.proto
```diff
@@ -0,0 +1,36 @@
...
+    CC_Parameters cc_params    = 1;
+    Host_Parameters host_params = 2;
+    bytes enclave_vk           = 3; // chaincode enclave public key
+    bytes channel_hash         = 4; // cryptographic hash of the channel genesis block;
+    string tlcc_mrenclave      = 5; // expected TLCC mrenclave
+}
+
+message Credentials {
+    google.protobuf.Any serialized_attested_data   = 1;  // serialization of type **Attested_Data**
+    bytes attestation           = 2;    // optional attestation/quote
```
g2flyer, line 34: purely a style suggestion but we either call both of them `..._bytes` or drop it from both?

protos/fpc/key_dist.proto
```diff
@@ -0,0 +1,33 @@
...
+option go_package = "github.com/hyperledger-labs/fabric-private-chaincode/internal/protos";
+
+message CCKeyRegistrationMessage {
+    bytes cc_params_hash = 1;       // defines the context as SHA2 of cc_params
+    bytes chaincode_ek   = 2;       // public chaincode encryption key
+    bytes enclave_id     = 3;       // creator of this message
+}
+
+message SignedCCKeyRegistrationMessage {
+    bytes msg_bytes = 1;            // public chaincode encryption key
```
g2flyer, line 18: shouldn't this also be ideally a `Any` (of CCKeyRegistrationMessage)? And comment seems wrong ...
```suggestion
    google.protobuf.An serialized_cckey_reg_msg = 1;            // serialization of type CCKeyRegistrationMessage 
```

protos/fpc/key_dist.proto
```diff
@@ -0,0 +1,33 @@
...
+message ExportMessage {
+    bytes cc_params_hash = 1;       // defines the context as SHA2 of cc_params
+    bytes chaincode_ek   = 2;       // public chaincode encryption key
+    bytes cckeys_enc     = 3;       // chaincode keys encrypted for the receiver
+    bytes enclave2_vk    = 4;       // receiver of this message
+    bytes enclave_vk     = 5;       // creator of this message
+}
+
+message SignedExportMessage {
+    bytes export_msg_bytes = 1;
```
g2flyer, line 31: Also target for Any?
```suggestion
    google.protobuf.An serialized_export_msg = 1;            // serialization of type ExportMessage
```

protos/fpc/key_dist.proto
```diff
@@ -0,0 +1,33 @@
...
+    bytes msg_bytes = 1;            // public chaincode encryption key
+    bytes signature = 2;            // signature of the message creator (enclave)
+}
+
+message ExportMessage {
+    bytes cc_params_hash = 1;       // defines the context as SHA2 of cc_params
+    bytes chaincode_ek   = 2;       // public chaincode encryption key
+    bytes cckeys_enc     = 3;       // chaincode keys encrypted for the receiver
+    bytes enclave2_vk    = 4;       // receiver of this message
+    bytes enclave_vk     = 5;       // creator of this message
```
g2flyer, line 27: The name doesn't really reflect the role, maybe better encode role also in name?
```suggestion
    bytes receiver_enclave_vk    = 4;       // receiver of this message
    bytes sender_enclave_vk     = 5;       // sender (creator) of this message
```

docs/design/fabric-v2+/interfaces.md
```diff
@@ -27,7 +27,7 @@ func Invoke(stub shim.ChaincodeStubInterface) pb.Response {}
 // Note: to get the endpoints of FPC endorsing peers do the following:
 // - discover all endorsing peers (and their endpoints) for the FPC chaincode using "normal" lifecycle
 // - query `getEnclaveId` at all the peers discovered
-// - query queryCCInfo with all received enclave_ids
+// - query `queryListEnclaveCredentials` with all received enclave_ids
```
g2flyer, line 30: Maybe we could drop here all protobuf definitions and just reference the .proto files?  Removes any duplication?

g2flyer (commented):
Looks already pretty good, a few suggestions for changes to improve doc and protobuf itself

protos/fpc/attestation.proto
mbrandenburger: oh if we do that we need to move the generated protos also to corresponding subfolders such as `internal/protos/attestation` etc ... otherwise we have multiple go packages in a single folder :/

protos/fpc/fpc.proto
mbrandenburger: in `fpc-management.md` we define:
```
* `--version <MRENCLAVE as hexadecimal string>`, the version of the FPC chaincode must contain a string that represent the identity of the enclave. This is the same identity which the trusted hardware computes and attests to. The string can be conveniently found in the `mrenclave` output file of the `generate_mrenclave.sh` script.
```

will add a comment here

protos/fpc/fpc.proto
mbrandenburger: added a reference to https://github.com/hyperledger/fabric-protos/blob/master/msp/identities.proto#L15

protos/fpc/fpc.proto
mbrandenburger: dropped

protos/fpc/attestation.proto
g2flyer: > oh if we do that we need to move the generated protos also to corresponding subfolders such as `internal/protos/attestation` etc ... otherwise we have multiple go packages in a single folder :/
> 
> Edit: just tested, protoc will not create these folders for us.

I would have hoped that protoc/nanopb would do create the folders.  As not, let's leave it as is ...
(Judging from https://github.com/golang/protobuf/issues/515 it seems though as iff at least for go the default is/was to generate it in the proper directories?)

protos/fpc/fpc.proto
g2flyer: that's the reference to version encoding i also referenced; what i meant it seems we haven't done the equivalent for tlcc anywhere (but naturally might use the same encoding, even though for this case the original binaries might work as well for the comparison, but then for debugging/logging makes sense to use same encoding ..)

protos/fpc/fpc.proto
```diff
@@ -11,26 +11,49 @@ package fpc;
...
 }
 
 message Credentials {
-    google.protobuf.Any serialized_attested_data   = 1;  // serialization of type **Attested_Data**
-    bytes attestation           = 2;    // optional attestation/quote
-    bytes evidence_bytes        = 3;    // serialized attestation evidence
+    // serialization of type **Attested_Data**
+    google.protobuf.Any serialized_attested_data = 1;
+
+    // serialized (optional) attestation/quote
```
g2flyer, line 54: the "serialized" doesn't really add much additional info from before ;-)
```suggestion
    // serialized (optional) attestation/quote as output by `get_attestatation`, see `interfaces.attestation.md`
```

protos/fpc/fpc.proto
```diff
@@ -11,26 +11,49 @@ package fpc;
...
+    bytes attestation = 2;
+
+    // serialized attestation evidence
```
g2flyer, line 57:
```suggestion
    // serialized attestation evidence as output by `AttestationToEvidence`, see `interfaces.attestation.md`
```

protos/fpc/attestation.proto
mbrandenburger: cool this works! updated

g2flyer (approved):
Looks good!

protos/fpc/fpc.proto
```diff
@@ -0,0 +1,59 @@
...
+
+    // version of chaincode, this is the expected mrenclave
+    // encoded as hexstring (as generated by common/enclave/generate_mrenclave.sh)
+    string version = 2;
+
+    // chaincode sequence number
+    int64 sequence = 3;
+
+    // name of channel
+    string channel_id = 4;
```
bvavala, line 25: Back to a comment of mine in #439 , here there is a mismatch with the fields. tlcc_mrenclave, channel_hash and/or mspid are now in attested_data, but... attested_data already contains cc_params. 
Either way, I don't think it changes much, it's only matter of consistency.

mbrandenburger: mismatch resolved in #443

Issue #445, opened 2020-09-25
Refactor logging for enclaves


bvavala:
The implementation of the logging mechanism is now spread out in the `common`, `enclave`, and other test folders. It would be appropriate and clearer to implement it in `common`, specifying any needed .edl, .c, .h file.

We probably also want to disable logging iff we build with SGX_BUILD=RELEASE (but i guess while we have right now a makefile variable for this, we don't have a C define? Though might be easy to add given that every C code (?) is using the shared cmake includes?

See also discussion in https://github.com/hyperledger-labs/fabric-private-chaincode/pull/444#discussion_r493943409

Check out a similar implementation of the attestation-api under crypto.

PR #447, opened 2020-10-06
Refactor of logging mechanism

bvavala:
This PR addresses #445 , providing a new logging mechanism for untrusted C/Cpp code, trusted enclaves and glue code for Go/CGo code.
Such unified capability is fully under `common/logging` and replaces the existing one. Trusted/untrusted code continue to use the `LOG_*` macros as originally defined. The implementation of these macros is located in the `tlogging` and `ulogging` libraries for the trusted and untrusted space, respectively. The trusted library always forwards the log to the untrusted one. The untrusted one must be configured with a callback function in order to forward any log to an API which is expected to match the prototype and behavior of the standard `puts`. If it is not configured, it simply drops any received log, and the logging call fails. This allows to send the logs to standard output (through `puts`) in test applications, or to enable Go/Cgo to set up a custom API to receive the any log from the C/Cpp/Enclave code being executed. 
 
Contributions are:
- refactor of C/Cpp logging, now compiled into two libraries (`ulogging` and `tlogging`) and then linked into an executable
- (very) basic test
- port of ECC, TLCC and crypto tests to use the new logging capability
- some renaming based on discussions in #444


common/crypto/attestation-api/test/CMakeLists.txt
```diff
@@ -23,12 +29,15 @@ ADD_EXECUTABLE(${VERIFY_EVIDENCE_APP}
     common/test-utils.cpp
     )
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SGX_COMMON_CFLAGS}")
+
```
g2flyer, line 33: I guess you added it so logging gets SGX_BUILD?  Wouldn't it be better/more robust to do that inside `${LOGGING_PATH}/CMakeVariables.txt` so it doesn't accidentally gets omitted in other new cmakefiles consuming logging eventually? Obviously might lead to duplicate definitions in some cases but that should not be a real problem?

common/logging/include/log-defines.h
```diff
@@ -0,0 +1,76 @@
...
+#endif
+
+#define LOC_FMT " (%s:%d) "
+
+#define NRM "\x1B[0m"
+#define CYN "\x1B[36m"
+#define YEL "\x1B[33m"
+#define RED "\x1B[31m"
+
+#ifndef DO_DEBUG
```
g2flyer, line 22: As you mentioned in discussion in PR #444, this is set only if SGX_BUILD is DEBUG, so it addresses the related aspect of issue #445.  Two things, though:
- might be good to add a comment here (or in the trusted logging.h) mentioning that in (PRE)RELEASE mode debug messages are guaranteed not to leak from enclave but any other log-message should not reveal anything sensitive?
- we also have a DEBUG flag from PDO (PDO_DEBUG_BUILD), we probably should add the setting (or in particular, unsetting) in `cmake/ConfigSGX.cmake`?

common/logging/include/log-defines.h
```diff
@@ -0,0 +1,76 @@
...
+extern "C" {
+#endif
+int loggingf(const char* fmt, ...);
+#ifdef __cplusplus
+}
+#endif
+
+#if DO_DEBUG == true
+#define LOG_DEBUG(fmt, ...) \
+    loggingf(CYN "DEBUG " LOC_FMT TAG YEL fmt NRM "\n", __FILE__, __LINE__, ##__VA_ARGS__)
```
g2flyer, line 48: Passing the logging level explicitly rather than implicitly inside the string would in principle allow you do better output filtering. That said, given that the flogger support in the go stub/peer is essentially dropped i guess it really doesn't matter. And need arises we obviously can add that at that time ...

tlcc/enclave/cgo_logging.go
```diff
@@ -0,0 +1,17 @@
...
+package enclave
+
+// extern void golog(char*);
+//
+// int golog_cgo_wrapper(const char* str)
+// {
+//      golog((char*)str);
+//      return 1;
+// }
+import "C"
```
g2flyer, line 17: This is a duplicate version.  I wonder whether could be handled via a small package in `internal` which does above and defines a single go function which gets a logger object as parameter, sets the callback plus based on a function pointer to the function defined with the logger object as a closure. Not 100% sure though whether that works and i guess as long as there are only two occurrence of this pattern it is certainly not worth spending much time on this. If we add  more places, might be worth though revisiting?

common/crypto/attestation-api/test/get_attestation_app/app/main.cpp
```diff
@@ -35,28 +35,32 @@ int main()
         exit(-1);
     }
 
+    // set logging callback
+    logging_set_callback(puts);
```
g2flyer, line 39: Should we maybe define `puts` as default in `common/logging/untrusted/logging.c`?   Note right now the callback is NULL by default and we do not test whether the callback is set, so we would crash  ...

g2flyer (commented):
Overall looks great.  A few suggestions below for potential improvements, but all take it or leave it ...

common/crypto/attestation-api/test/CMakeLists.txt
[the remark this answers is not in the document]
bvavala: > I guess you added it so logging gets SGX_BUILD? 

right.

>Wouldn't it be better/more robust to do that inside ${LOGGING_PATH}/CMakeVariables.txt

I would agree, but there is a problem.
The logging library is precompiled and only linked here. These flags are necessary at compile-time _of the current executable,_ when the logging headers are included, in order to define the `LOG_*` defines.

common/logging/include/log-defines.h
bvavala: > might be good to add a comment here (or in the trusted logging.h) mentioning that in (PRE)RELEASE mode debug messages are guaranteed not to leak from enclave but any other log-message should not reveal anything sensitive?

I have added some comments explaining that.
However, as I mentioned before, notice that `logging` (except for the `test`) does not use the `SGX_COMMON_CFLAGS`. The define is only necessary at compile-time when the ``log-defines.h" is included.
So, trusted logging, by itself, does not guarantee that there are no leaks.

bvavala: > we also have a DEBUG flag from PDO (PDO_DEBUG_BUILD), we probably should add the setting (or in particular, unsetting) in cmake/ConfigSGX.cmake?

I would be hesitant about that, because it would expose PDO details (that I think are not relevant) to FPC.
Right now, PDO_DEBUG_BUILD is only exposed in `common/crypto/CMakeLists.txt`, and disabled by default.
I needed that for the PDO integration, but hopefully it will not be used.

common/logging/include/log-defines.h
bvavala: I agree.

tlcc/enclave/cgo_logging.go
bvavala: > This is a duplicate version. 

True. It was convenient and necessary because cgo duplicates the preamble when it compiles the exported functions. 

> If we add more places, might be worth though revisiting?

Definitely.

common/crypto/attestation-api/test/get_attestation_app/app/main.cpp
bvavala: > Should we maybe define puts as default in common/logging/untrusted/logging.c?

No strong opinion on that.

With `puts`.
Pros: you should always get something displayed.
Cons: dependency on `puts`

Without `puts`.
Pros: no dependency on an output function.
Cons: if you forget to set it, nothing is displayed. 

> we do not test whether the callback is set, so we would crash

Nope, see line 26 of `common/logging/untrusted/logging.c`: if not set, the logging function returns 0/false.
Thing is that: the call is (usually) in a define, and the return value is not checked.

ecc/enclave/cgo_logging.go
```diff
@@ -0,0 +1,17 @@
...
+   SPDX-License-Identifier: Apache-2.0
+*/
+
+package enclave
+
+// extern void golog(char*);
+//
+// int golog_cgo_wrapper(const char* str)
+// {
+//      golog((char*)str);
```
mbrandenburger, line 14: should we also move `golog()` from `enclave_stub.go` also to this file?

mbrandenburger: this is is also the change to clean-up the unused `_logger` in `ecc/enclave/enclave_stub.go`.

mbrandenburger: and we even `var logger = flogging.MustGetLogger("ecc_enclave")` go be moved to this logging file. so we have everything related to logging in a single place

ecc/enclave/enclave_stub.go
```diff
@@ -305,6 +309,10 @@ type StubImpl struct {
 
 // NewEnclave starts a new enclave
 func NewEnclave() Stub {
+	r := C.logging_set_callback(C.log_callback_f(C.golog_cgo_wrapper))
+	if r == false {
+		panic("error initializing logging for cgo")
+	}
```
mbrandenburger, line 315: I have the feeling this would fit pretty well in a `init` method inside `logging.go`

ecc/enclave/cgo_logging.go
mbrandenburger: this file could also be just called `logging.go` I think

mbrandenburger (commented):
This is pretty cool! thanks Bruno, below some comments on the go-side logging.

common/crypto/attestation-api/test/CMakeLists.txt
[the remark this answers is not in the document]
g2flyer: but you _do_ include ${LOGGING_PATH}/CMakeVariables.txt above? So moving it to there would give the desired effect for both the library itself and consumers? Are there cases we do currently do not include it but would have to include it with my suggested move?

common/logging/include/log-defines.h
g2flyer: ok

common/crypto/attestation-api/test/get_attestation_app/app/main.cpp
g2flyer: ok, overlooked the check.  But is there any reason where `puts` is currently not available? Just seems to me having it as a default saves you having to add this in all test programs and if you forget you might be confused first at time you want to look at log and wouldn't see anything. I..e., i see only benefits but no real drawbacks

ecc/enclave/cgo_logging.go
g2flyer: @mbrandenburger if you move the logger here then you loose the context specific logger we currently have.  As mentioned above ("This is a duplicate version ...") i think you might be able to get the best of both worlds by passing the logger via some closure.

g2flyer: > then you loose the context specific logger we currently have.
admittedly, for any particular program, there is only one context/logger, so it really doesn't matter if for enclave logging it says tlcc_enclave or ecc_enclave or alike as the context is clear. So just having to logger object in internal/logging.go would be considerably easier than what i suggest and perfectly sufficient :-)

common/crypto/attestation-api/test/CMakeLists.txt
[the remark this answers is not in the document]
bvavala: That's right -- thought it was cmakelist -- and it's now in cmakevariables.
Also, I have revised the DO_DEBUG part. As it belongs to logging, I moved it there.

common/crypto/attestation-api/test/get_attestation_app/app/main.cpp
bvavala: > i see only benefits but no real drawbacks

In that case... done!

ecc/enclave/cgo_logging.go
mbrandenburger: not sure if I really understand this. But moving `var logger = flogging.MustGetLogger("ecc_enclave")` into `cgo_logging.go` will retain the contexts specific logger as it is still in the same package just in a different file.

bvavala:
@mbrandenburger @g2flyer take a look at 7bc5a4 for the refactoring of the go log part.
It's definitely cleaner now, particularly with the use of `init`. 
The logging is however split in two files. The cgo one is necessary because it defines a function that must be visible externally -- and this cannot stay together with an exported function (i.e., golog). The go one includes all the logging code that was previously in the stub.

g2flyer (approved):
Didn't realize that you could not define in a single cgo-file a go export used in cgo as well as a cgo export used in go.  I guess probably has to do something with the sequencing of the processing of //export ...
Anyway, two minor observation, otherwise it e2e works nicely in HW mode ...

mbrandenburger (approved):
LGTM! Thanks Bruno

ecc/enclave/logging_cgo.go
```diff
@@ -14,4 +14,7 @@ package enclave
 //      golog((char*)str);
 //      return 1;
 // }
+//
+// int golog_cgo_wrapper(const char* str);
+//
```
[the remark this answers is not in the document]
bvavala: That's right, leftover -- removed.

tlcc/enclave/logging_cgo.go
```diff
@@ -14,4 +14,7 @@ package enclave
 //      golog((char*)str);
 //      return 1;
 // }
+//
+// int golog_cgo_wrapper(const char* str);
+//
```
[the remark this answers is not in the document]
bvavala: That's right, leftover -- removed.

bvavala:
> Didn't realize that you could not define in a single cgo-file a go export used in cgo as well as a cgo export used in go. I guess probably has to do something with the sequencing of the processing of //export ...
> Anyway, two minor observation, otherwise it e2e works nicely in HW mode ...

(verbatim)
_"Using //export in a file places a restriction on the preamble: since it is copied into two different C output files, it must not contain any definitions, only declarations. If a file contains both definitions and declarations, then the two output files will produce duplicate symbols and the linker will fail. To avoid this, definitions must be placed in preambles in other files, or in C source files."_

PR #450, opened 2020-10-20
ERCC refactoring

mbrandenburger:

- implements RegiterEnclave
- implemented as a chaincode as a service

Addresses ERCC related work items in #404

```

### CODE REVIEW

File: ercc/registry/registry.go [hunk 1]
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	if err := checkAttestedData(ctx, rs.Verifier, rs.PEvaluator, &attestedData, &credentials); err != nil {
+		return err
+	}
+
+	chaincodeId := attestedData.CcParams.ChaincodeId
+	enclaveId := utils.GetEnclaveId(&attestedData)
+
+	key, err := ctx.GetStub().CreateCompositeKey("namespaces/credentials", []string{chaincodeId, enclaveId})
```
g2flyer, line 167: Wouldn't you have to do here also some check whether there are already registered enclaves for that chaincode and, if so, verify that they are consistent?

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// check that registration transaction creator has same mspid as the enclave owner
+	if err := pe.EvaluateCreatorIdentity(creatorIdentityBytes, attestedData.HostParams.PeerIdentity); err != nil {
+		return fmt.Errorf("creator identity evaluation failed: %s", err)
+	}
+
+	// TODO add more checks
+	// channel_hash should correspond to peers view of channel id (POST-MVP)
+	// TLCC_MRENCLAVE matches the version baked into ERCC (POST-MVP)
+	// check org-enclave binding (POST-MVP)
```
g2flyer, line 230: hmm, didn't we just check that with above HostParams checks?
```suggestion
```

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+	// deployment validation (POST-MVP)
+	// validate FPC deployment (restriction) policy
```
g2flyer, line 232: what's the difference between the two?

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return enclaveIds, nil
+}
+
+// returns the chaincode encryption key for a given chaincode id
+func (rs *Contract) QueryChaincodeEncryptionKey(ctx contractapi.TransactionContextInterface, chaincodeId string) ([]byte, error) {
+	//input chaincodeId string
+
+	//return chaincode_ek []byte
+	return nil, nil
```
g2flyer, line 127: Hmm, some TODO is missing here (or even better, the corresponding implementation: should be straightforward?

g2flyer (commented):
Some initial comments.  I haven't looked at the testing stuff, you have to give a bit more context on that (see corresponding comment).

ercc/connection.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
mbrandenburger: I agree, see d14b41c, it is already moved somewhere.

ercc/attestation/interface.go
```diff
@@ -0,0 +1,5 @@
+package attestation
+
+type VerifierInterface interface {
+	VerifyEvidence(evidenceBytes, expectedStatementBytes []byte, expectedMrEnclave string) error
+}
```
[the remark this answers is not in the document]
mbrandenburger: right, I also think that we will only have a single implementation of this interface that contains the cgo part calling the real attestation api. However, having this interface is just here to autogenerate mock stubs and use them in testing.

ercc/chaincode.env
```diff
@@ -0,0 +1,10 @@
+# CHAINCODE_SERVER_ADDRESS must be set to the host and port where the peer can
+# connect to the chaincode server
+CHAINCODE_SERVER_ADDRESS=ercc.org1.example.com:9999
+
+# CHAINCODE_ID must be set to the Package ID that is assigned to the chaincode
+# on install. The `peer lifecycle chaincode queryinstalled` command can be
+# used to get the ID after install if required
+CHAINCODE_ID=ercc_1.0:26135f3588187d357d35b2e3b6e1c3b2805860815c10aa37b1e90a3b44ba47d2
+
+FABRIC_LOGGING_SPEC=DEBUG
```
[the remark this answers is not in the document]
mbrandenburger: ack! moved

ercc/registry/mocks/chaincodestub.go
```diff
@@ -0,0 +1,2878 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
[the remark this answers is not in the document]
mbrandenburger: yes correct, theres files are generated using `go generate ./...` based on the interfaces specified in `registry_test.go` using [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter). I added it also to our tools.go, so it should be already there.

I think for the test mocks here we can follow the fabric approach and check these generated files in. However, we can come back on that later.

For documentation I only used https://github.com/maxbrunsfeld/counterfeiter and looking around in the fabric repo, for instance [here](https://github.com/hyperledger/fabric/tree/master/core/chaincode).

internal/utils/endorsement.go
```diff
@@ -3,14 +3,22 @@ package utils
...
+	pb "github.com/hyperledger/fabric-protos-go/peer"
 	"github.com/hyperledger/fabric/bccsp/factory"
-	"github.com/hyperledger/fabric/common/cauthdsl"
 	"github.com/hyperledger/fabric/msp"
 	"github.com/hyperledger/fabric/msp/mgmt"
+	"github.com/pkg/errors"
 )
 
-func NewPolicyEvaluator() *PolicyEvaluator {
+type PolicyEvaluatorInterface interface {
```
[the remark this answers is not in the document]
mbrandenburger: we have an interface here to add mocks so in `registry_test.go` we can better control the behaviour of the policy evaluator. I think this is also the normal way to go in fabric. Once you implement a component that is composed of other components, you better use interfaces so you can concentrate on the implementation and testing of the component you are developing.

ok, the main reason for the change here is that the previous implementation of `EvaluateIdentity` relied on a identityDeserializer provided by default msp provider (not sure if this is the right name?!?! anyway). In particular, the old version called the following code to get the deserializer.

```go
manager := mgmt.NewDeserializersManager(factory.GetDefault())
eturn &PolicyEvaluator{
		IdentityDeserializer: manager.GetLocalDeserializer(),
``` 

However, using this code a chaincode environment we get a deserializer that cannot parse the identities simply because the chaincode environment has no access to the msp information that a peer has. 

Therefore, this change here. What we do now is essentially, unmarshal the policy and the serialized identity and check if the msp id of is part of the policy.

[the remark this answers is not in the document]
mbrandenburger: will add a few comments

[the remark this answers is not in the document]
mbrandenburger: Not sure what you mean with different implementation of this evaluator?

internal/utils/chaincode.go
```diff
@@ -18,14 +18,14 @@ import (
 
 const MrEnclaveLength = 32
 
-func GetChaincodeDefinition(chaincodeId string, stub shim.ChaincodeStubInterface) (*lifecycle.QueryChaincodeDefinitionResult, error) {
+func GetChaincodeDefinition(chaincodeId, channelId string, stub shim.ChaincodeStubInterface) (*lifecycle.QueryChaincodeDefinitionResult, error) {
```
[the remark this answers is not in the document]
mbrandenburger: mh, yeah I think this change is not really necessary. I guess some relict some playing with the code :)

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+	PEvaluator utils.PolicyEvaluatorInterface
+}
+
+// returns a set of credentials registered for a given chaincode id
+// Note: to get the endpoints of FPC endorsing peers do the following:
+// - discover all endorsing peers (and their endpoints) for the FPC chaincode using "normal" lifecycle
+// - query `getEnclaveId` at all the peers discovered
+// - query `queryListEnclaveCredentials` with all received enclave_ids
+// this gives you the endpoints and credentials including enclave_vk, and chaincode_ek
+func (rs *Contract) QueryListEnclaveCredentials(ctx contractapi.TransactionContextInterface, chaincodeId string) ([][]byte, error) {
```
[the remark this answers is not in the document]
mbrandenburger: so by default the contract API only support a bunch of input and output primitive data types. Our credential data type is not supported. There is a way to provide custom serializer to make this possible but I haven't yet looked into it. 

Also, maybe we also don't care in this function because getState(ByPar...) returns a serialized Credentials object. So if this function would return an array of []Credentials. we would need to deserialize them after getState and returns. However, the custom serializer would serialize them again. Let me add more comments to the code.

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+			return nil, err
+		}
+
+		allCredentials = append(allCredentials, q.Value)
+	}
+
+	return allCredentials, nil
+}
+
+func (rs *Contract) QueryEnclaveCredentials(ctx contractapi.TransactionContextInterface, chaincodeId, enclaveId string) ([]byte, error) {
```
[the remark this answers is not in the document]
mbrandenburger: see comment above

ercc/registry/registry.go
mbrandenburger: added todos

ercc/registry/registry.go
mbrandenburger: I added a TODO.

ercc/registry/registry.go
mbrandenburger: same

ercc/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM hyperledger/fabric-private-chaincode-base-rt:latest
```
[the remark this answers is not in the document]
mbrandenburger: yes, we should move the Dockerfile somewhere else. I would do that in a follow up PR once it's clear where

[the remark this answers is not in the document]
g2flyer: as mentioned, it should be part of the sample net stuff, i.e., in that pr. And regarding location i would suggest we put all of the sample net into 'util/sample-net' or alike as this is also were we have docker and docker-compose, with the latter presumably eventually replaced by the sample-network stuff ...

ercc/Makefile
```diff
@@ -6,32 +6,28 @@
...
 
-ERCC_NAME?=ercc
-DOCKER_IMAGE?=$$(docker images | grep -- -$(ERCC_NAME)- | awk '{print $$1;}')
+docker:
+	docker build --no-cache -t fpc/ercc .
 
-docker-clean:
-	-if [ ! -z "$(DOCKER_IMAGE)" ]; then docker rmi -f $(DOCKER_IMAGE); fi
-	-if [ ! -z "$(DOCKER_IMAGE)" ]; then echo "\033[0;33mWARNING: ercc not deleted (peer still up?)\033[0m"; fi
+run:
```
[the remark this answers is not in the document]
g2flyer: you moved only the docker stuff. I was referring to also all the packaging stuff which is also sample-net-test specific ...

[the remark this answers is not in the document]
g2flyer: the docker part though imho is also part of sample-net-test part.  But obviously that has to be handled together with whereever the Docker file is

ercc/registry/mocks/chaincodestub.go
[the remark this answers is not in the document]
g2flyer: what is the reasoning to check in the mocks?! and what makes sure that they are kept in sync with `registry_test.go`. Honestly i don't really like that. At the very least we should have a make target which creates them and they shouldn't appear out of thin air without documentation ...

ercc/registry/registry.go
g2flyer: Can't we just implement it? should be pretty similar to the other queries?

ercc/registry/registry.go
```diff
@@ -0,0 +1,269 @@
...
+
+	chaincodeId := attestedData.CcParams.ChaincodeId
+	enclaveId := utils.GetEnclaveId(&attestedData)
+
+	key, err := ctx.GetStub().CreateCompositeKey("namespaces/credentials", []string{chaincodeId, enclaveId})
+	if err != nil {
+		return err
+	}
+
+	// TODO check if this enclave is already registered
```
g2flyer, line 179: what are we doing if already registered? fail?

ercc/registry/registry.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// TODO perform the MVP check that only a single enclave is registered per chaincode
```
g2flyer, line 181:
```suggestion
	// TODO perform the (enclave) endorsement policy specific tests:
	// - MVP (designated chaincode): verify no other enclave (of other peers) is registered or fail
	// - Post-MVP: check consistency with potentially existing enclaves
```

ercc/registry/registry.go
mbrandenburger: I can be your PR once this here is merged!

ercc/registry/registry.go
mbrandenburger: good question, let's figure out later.

ercc/registry/registry_test.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/hyperledger-labs/fabric-private-chaincode/internal/utils"
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	"github.com/hyperledger/fabric-contract-api-go/contractapi"
+	"github.com/hyperledger/fabric-protos-go/ledger/queryresult"
+	"github.com/hyperledger/fabric-protos-go/peer/lifecycle"
+	"github.com/hyperledger/fabric/protoutil"
+	"github.com/stretchr/testify/require"
+)
+
+//go:generate counterfeiter -o mocks/transaction.go -fake-name TransactionContext . transactionContext
```
g2flyer, line 23: was there a strong reason change the outpudle directory and fake-name?  I actually like the default name and distinguishes from the normal mocks (see my other comment that i got confused by the name "mock" because this mock doesn't match the other mocks which are usually self-contained functional replacements for a module rather than the kindof  "remote-controlled" ones here.   That said, no strong reason but it confused me and using defaults even makes these comments more consice?

ercc/registry/registry_test.go
```diff
@@ -0,0 +1,339 @@
...
+type transactionContext interface {
+	contractapi.TransactionContextInterface
+}
```
g2flyer, line 26: Wouldn't this in principle be unecessary if you use the third-party variant of generation where the argument is <package-path>.<Interface>. (Not that i'm asking for change, just trying to understand ....)

ercc/registry/registry.go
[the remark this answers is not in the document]
g2flyer: ok, completely missed when looking at it that these parameters come from the request, not one of our other go functions, and hence are the encoded version of these objects (i.e., serialized protobuf and as discussed in this morning calls also base64 encoded (on top) so on the client side we have strings which can be handled by peer cli script.
Clarifying that in documentation though would help. => all add a corresponding proposal for a change ...

ercc/registry/registry.go
```diff
@@ -0,0 +1,271 @@
...
+// returns a set of credentials registered for a given chaincode id
+// Note: to get the endpoints of FPC endorsing peers do the following:
+// - discover all endorsing peers (and their endpoints) for the FPC chaincode using "normal" lifecycle
+// - query `getEnclaveId` at all the peers discovered
+// - query `queryListEnclaveCredentials` with all received enclave_ids
+// this gives you the endpoints and credentials including enclave_vk, and chaincode_ek
+//
+// Note that this implementation returns a set of serialized credentials in order to send it to the receiver.
+// That is, the receiver needs to deserialize the return value into []Credentials
+// TODO look into custom serializer to make this call more aligned with the FPC API in `interfaces.md`
```
g2flyer, line 40: well, we didn't really define the encoding in `interfaces.md` (there is only a "maybe" which is anyway different from what we talk now).  From discussion this morning it seems we have agreed on base64-encoding of (standard binary) protobuf encoding of Objects?

ercc/registry/registry_test.go
mbrandenburger: first reason, I got inspired by some package in fabric using counterfeiter with this naming convention. And by looking at the output of the default setting, I like this non-standard format more. I think we maybe rename the `mocks` folder to `fakes` to do something against you mental picture of mocks and "remote-controlled" fakes. :)

CHanging also the interface names to default with FakeEverything naming I don't like so much as the fake package already notes that this is a "mock" or "fake" implementation.

ercc/registry/registry.go
g2flyer: well, we can do it in a separate PR; just seemed random why we do all the others but not these and add the overhead of a separate PR? But i guess as long as we have the TODO my main concern is addressed

ercc/registry/registry.go
```diff
@@ -0,0 +1,255 @@
...
+	// validate FPC deployment (restriction) policy
+
+	return nil
+}
+
+// registers a CCKeyRegistration message that confirms that an enclave is provisioned with the chaincode encryption key.
+// This method is used during the key generation and key distribution protocol. In particular, during key generation,
+// this call sets the chaincode_ek for a chaincode if no chaincode_ek is set yet.
+func (rs *Contract) RegisterCCKeys(ctx contractapi.TransactionContextInterface, ccKeyRegistrationMessageBase64 string) error {
+	//input msg CCKeyRegistrationMessage
```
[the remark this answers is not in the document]
g2flyer: It seems you approved the comment but didn't add the TODO?

ercc/registry/registry.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// returns a set of credentials registered for a given chaincode id
+// Note: to get the endpoints of FPC endorsing peers do the following:
+// - discover all endorsing peers (and their endpoints) for the FPC chaincode using "normal" lifecycle
+// - query `getEnclaveId` at all the peers discovered
+// - query `queryListEnclaveCredentials` with all received enclave_ids
+// this gives you the endpoints and credentials including enclave_vk, and chaincode_ek
+//
+// Note that this implementation returns a set of serialized credentials in order to send it to the receiver.
```
g2flyer, line 38: Might as well make the encoding explicit?
```suggestion
// Note that this implementation returns a set of (base64-encoded) protobuf-serialized `Credential` objects in order to send it to the receiver.
```

ercc/registry/registry_test.go
mbrandenburger: maybe. cannot answer this without trying out.

ercc/registry/registry.go
```diff
@@ -0,0 +1,271 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+Copyright 2020 Intel Corporation
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package registry
```
g2flyer, line 8:
```suggestion
// this package defines the client-facing interface of ERCC as defined in the ERCC Interface section in [specifications](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/master/docs/design/fabric-v2%2B/interfaces.md)
package registry
```

internal/utils/endorsement.go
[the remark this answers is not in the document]
g2flyer: Actually, i think i know where the disconnect comes. Originally before lightweight the chaincode endorsement policy was really policy which decided on validation rules for attestations, i.e., what in our proposal we call `enclave endorsement` policies.  However, now we have this `enclave endorsement policy` separate from the `validation endorsement policy` and the endorsement policy of the chaincode is the latter whereas the former is for MVP `designated enclave` policy implicitly defined  (and for post-mvp specified in some yet to be determined fashion).  What you reference/verify here is hence against the `validation endorsement policy`.  However, what we do have to verify in errc is the `enclave endorsement policy`.  Hence my reference to above that i was speculating that we might have (eventually) multiple implementations depending on what enclave endorsement policy we support. Not that i think this is really necessary and i now also understand why you have an interface even if there is only one implementation, the current implemention imho is doing the wrong stuff (at least as used by ercc)

ercc/registry/registry.go
```diff
@@ -0,0 +1,271 @@
...
+	// validate FPC deployment (restriction) policy (POST-MVP)
+
+	return nil
+}
+
+// registers a CCKeyRegistration message that confirms that an enclave is provisioned with the chaincode encryption key.
+// This method is used during the key generation and key distribution protocol. In particular, during key generation,
+// this call sets the chaincode_ek for a chaincode if no chaincode_ek is set yet.
+func (rs *Contract) RegisterCCKeys(ctx contractapi.TransactionContextInterface, ccKeyRegistrationMessageBase64 string) error {
+	//input msg CCKeyRegistrationMessage
```
g2flyer, line 252:
```suggestion
func (rs *Contract) RegisterCCKeys(ctx contractapi.TransactionContextInterface, ccKeyRegistrationMessageBase64 string) error {
         // TODO: Implement me for MVP
         
	//input msg CCKeyRegistrationMessage
```

g2flyer (changes requested):
Overall great step forward for ERCC and probably (actually, almost certainly?) the most comprehensive test we have anywhere in FPC :-)

However, i think there is an issue with the policy evaluation where the old model of (chaincode) endorsement policies doesn't really represent the (enclave) endorsement policy we had in our spec and hence the implementation seems to do somewhat the wrong checks. See also some more text in the discussion on `endorsement.go`. That said, probably, best discussed in some interactive way instead of going through github?

We also should definitely add the Makefile target for the counterfeit generation.  I actually tried to add it as a proposal but then run into the problem i simply cannot build FPC right now (and hence not test any make target as `go generate` also failed).  I understand that integration tests and alike do not work but shouldn't `make build` work?  Do i do something wrong or did you maybe forget to check something in?
```
...
/hyperledger/fabric-contract-api-go v1.1.1
# github.com/hyperledger-labs/fabric-private-chaincode/internal/utils
../internal/utils/utils.go:65:33: undefined: protos
../internal/utils/utils.go:66:7: undefined: sha256
../internal/utils/utils.go:67:9: undefined: hex
Makefile:12: recipe for target 'ercc' failed
```

mbrandenburger:
Thanks @g2flyer. I updated the `fakes` generation, including renaming and add a build target for go generate. Fixed also the include issue in `utils.go`.

Let's have a quick sync offline to resolve the identified issue regarding EP validation.

g2flyer:
FYI: tried again to build after your update and i'm still running in issues with `make build`. 
- it fails in fabric during peer build in validators. I've solved it by a close-to-minimal fix of moving `fabric/patches/3_0001-Add-FPC-validator-to-library.patch` to `fabric/patches/3_0001-Add-FPC-validator-to-library.patch` after which fabric build. (Alternatively we could of course for now also just remove `fabric` from the subdir list in the top-level makefile) 
- `protos` has to be added to `FPC_SDK_DEP_DIRS` in the top-level makefile is needed or cc-builder fails to build
With above two changes, `make build` does work successfully from `$FPC_PATH`

For `make test` i would expected it to fail in `demo` (which we probably should just comment out for now in Makefile) and probably in `integration-test` (maybe also comment out for now?) but it already failed very quickly when trying to build `internal`:
```
....
ok      github.com/hyperledger-labs/fabric-private-chaincode/internal/testing/executor  (cached)
?       github.com/hyperledger-labs/fabric-private-chaincode/internal/testing/executor/worker   [no test files]
# github.com/hyperledger-labs/fabric-private-chaincode/internal/utils_test [github.com/hyperledger-labs/fabric-private-chaincode/internal/utils.test]
utils/chaincode_test.go:54:44: too many arguments in call to utils.GetChaincodeDefinition
        have (string, string, *shimtest.MockStub)
        want (string, shim.ChaincodeStubInterface)
utils/chaincode_test.go:62:44: too many arguments in call to utils.GetChaincodeDefinition
        have (string, string, *shimtest.MockStub)
        want (string, shim.ChaincodeStubInterface)
utils/chaincode_test.go:76:44: too many arguments in call to utils.GetChaincodeDefinition
        have (string, string, *shimtest.MockStub)
        want (string, shim.ChaincodeStubInterface)
FAIL    github.com/hyperledger-labs/fabric-private-chaincode/internal/utils [build failed]
FAIL
Makefile:9: recipe for target 'test' failed
make[1]: *** [test] Error 2
make[1]: Leaving directory '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/internal'
Makefile:22: recipe for target 'test' failed
```


BTW: just noticed that you called the branch `ercc-refactor`. Just to make sure, this is not an ercc-specific branch but is the intended 'refactor-to-new-ecc-ercc+lightweight' branch we talked earlier?

mbrandenburger:
Regarding building fabric peer this is a good observation! We can disable it for the moment and also add proto.

Regarding the issue in utils, this should be resolved.

internal/utils/endorsement.go
```diff
@@ -3,40 +3,89 @@ package utils
...
 
 type PolicyEvaluator struct {
-	msp.IdentityDeserializer
 }
 
+func NewPolicyEvaluator() PolicyEvaluatorInterface {
+	return &PolicyEvaluator{}
+}
+
+// EvaluateIdentity checks that the provided identity is a valid endorser as specified in the endorsement policy
```
bvavala, line 24: My understanding is that this part is to check whether the enclave is a valid endorser.

That said, do we need this whole part at all?

We definitely _need_ that in an FPC/Fabric integration where the enclave acts as a _Fabric_-endorser and signs the response.
However, in the new architecture, this is not the case anymore. The enclave acts as an _FPC_-endorser and its signature is checked by the FPC (_not Fabric_) validation chaincode. Most importantly, FPC uses ERCC to control which enclaves can FPC-endorse a transaction and, therefore, pass the validation check. Hence, it is not necessary to check whether the Fabric validation logic can recognize an enclave's signature through MSP.

bvavala (commented):
Some comments and a question about endorser validation.

internal/utils/endorsement.go
mbrandenburger: yes, all correct. We can keep this code for FULL FPC ercc implementation. Note that the policy evaluated got invalided with c4a0ace during review session with Michael :D

ercc/Makefile
```diff
@@ -6,32 +6,23 @@
...
 
-ERCC_NAME?=ercc
-DOCKER_IMAGE?=$$(docker images | grep -- -$(ERCC_NAME)- | awk '{print $$1;}')
+generate:
+	$(GO) generate ./...
 
-docker-clean:
-	-if [ ! -z "$(DOCKER_IMAGE)" ]; then docker rmi -f $(DOCKER_IMAGE); fi
-	-if [ ! -z "$(DOCKER_IMAGE)" ]; then echo "\033[0;33mWARNING: ercc not deleted (peer still up?)\033[0m"; fi
+docker:
```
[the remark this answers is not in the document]
mbrandenburger: oh this docker-clean cleans ercc images produced by a peer. Since we use either chaincode as a service or chaincode started using external builder this special clean command is not needed anymore. But I think a docker-clean is still helpful here. Let me bring back a docker clean.

g2flyer (approved):

bvavala (approved):

mbrandenburger:
Squashed and ready for merge.

PR #451, opened 2020-10-20
Add fabric-sample/test-network

mbrandenburger:

- add as git submodule
- provide script to setup test-network using external launcher
- provide install script for ercc and ecc_mock
- add initial ecc_mock implementation as a skeleton for the new enclave
  creation/registration flow and lightweight FPC flow

Fixes #

Please see `integration/README.md`

```

### CODE REVIEW

File: integration/README.md [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+cd $FPC_PATH/integration
+./setup.sh
+```
+
+Start network:
+```bash
+cd $FPC_PATH/integration/fabric-samples
+./network.sh up createChannel -c mychannel -ca
```
g2flyer, line 36: This command fails for me as the docker-compose file manupulation with yq seems to have a bug.
the script fails with 
```
....
rror: failed to create deliver client for orderer: orderer client failed to connect to localhost:7050: failed to create new connection: connection error: desc = "transport: error while dialing: dial tcp [::1]:7050: connect: connection refused"
Channel creation failed
Create channel failed
```
but i think the real cause is a much earlier error message (not causing immediately an abort) which says 
```
...
ERROR: The Compose file './docker/docker-compose-test-net.yaml' is invalid because:
services.peer0.org1.example.com.volumes value ['/var/run/:/host/var/run/', '../organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/msp:/etc/hyperledger/fabric/msp', '../organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls:/etc/hyperledger/fabric/tls', 'peer0.org1.example.com:/var/hyperledger/production', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml'] has non-unique elements
services.peer0.org2.example.com.volumes value ['/var/run/:/host/var/run/', '../organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/msp:/etc/hyperledger/fabric/msp', '../organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/tls:/etc/hyperledger/fabric/tls', 'peer0.org2.example.com:/var/hyperledger/production', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0//src/github.com/hyperledger-labs/fabric-private-chaincode/:/opt/gopath/src/github.com/hyperledger-labs/fabric-private-chaincode', '/home/msteiner/src/projects/intel/PoET-et-al/fabric/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/fabric-samples//config/core.yaml:/etc/hyperledger/fabric/core.yaml'] has non-unique elements

...
```


Actually, figured out the problem: the `yq` scripts are not idempotent for the volume manimpulations for peers! (They are for the orderer).

integration/package.sh
```diff
@@ -0,0 +1,81 @@
...
+  #}
+
+  jq -n '$ARGS.named' \
+    --arg address "${address}" \
+    --arg dial_timeout "${timeout}" \
+    --argjson tls_required "${use_tls}" \
+    >>"$out_dir/connection.json"
+}
+
+DEPLOYMENT_PATH=$(PWD)/_deployment
```
g2flyer, line 76: Same problem as at other place: did that really work for you that way?
```suggestion
DEPLOYMENT_PATH=${PWD}/_deployment
```

integration/installFPC.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
g2flyer, line 1: this script fails to me (even after fixing package.sh). Haven't debug yet what causes it ..

fabric/externalLauncher/bin/build
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
g2flyer, line 1: Calling them `externalLauncher` next to a dir `externalBuilder` seems to me a bit confusing.   

More importantly though, i think for clarity it would be good to keep stuff like these artifacts which are specific to the our particular use of fabric-sample network for our testing all as below a sample-network specific sub-dir.  Given where we currently keep some of that network stuff `utils/sample-network-testing` or alike would be one choice or it could also be a subdir of `integration`  or alike.  And as mentioned on the ercc PR, i think also stuff like ecc_mock is best moved there. Minimizes clutter on top-level filenamespace and also makes it easy to see what all fits together rather than having it spread out the whole file tree?

integration/core_ext.yaml
```diff
@@ -0,0 +1,6 @@
+chaincode:
```
g2flyer, line 1: before, `integration` directory contained only the tests but none of the scaffolding of the test.  I think it would help clarity if you could move scaffolding files into some subdirectory ...

g2flyer (commented):
Some initial comments which i managed tonight, so you have already to digest something your day. Couldn't make it run yet, so also couldn't cross-check whether maybe there is a way that we could override docker-compose settings for volumes by "injecting" a new "delta"-file (as we do in `utils/docker-compse` e.g., to had SGX HW support, couchdb and/or explorer.

integration/package.sh
mbrandenburger: oh on a mac this works; inside the docker-dev container, it doesn't. good catch

integration/README.md
mbrandenburger: oh true, this is a problem of this script. Added a quick fix

integration/installFPC.sh
mbrandenburger: I updated it. I think it had similar issues related to `$(PWD)` :)

fabric/externalLauncher/bin/build
mbrandenburger: well, right now it is not clear if externalLauncher will be only used for the test-network or eventually replace externalBuilder here. Since you also have suggested to try to keep `main.go` of ercc and ecc flexible to support both modes, I think we can also keep externalLauncher and externalBuilder together in the `fabric` folder. We can come back to this later, but I don't think this needs to be discussed now.

integration/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Before you start the network make sure you build ercc and ecc_mock. In
+
+```bash
+cd $FPC_PATH/ercc
+make && make docker
+
+cd $FPC_PATH/ecc_mock
+make && make docker
+```
```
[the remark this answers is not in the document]
mbrandenburger: ok, this currently describes my testing flow. Will be removed later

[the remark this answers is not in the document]
mbrandenburger: and indeed, it seems that we are missing again some go changes which are in another PR

utils/docker/dev_peer_cc-builder/Dockerfile
```diff
@@ -191,4 +191,7 @@ RUN apt-get install -y -q \
 	docker-compose\
 	${APT_DEV_ADD_PKGS}
 
+RUN GO111MODULE=on \
```
g2flyer, line 194: Do we need it from source (and latest)?  On my host it was convenient (and providing a well-defined version) to do a `snap install yq`.  Mainly asking whethere we _must_ have the go getted binary in our PATH. I guess if we don't fear version incompatibility then it probably shouldn't matter which version we take; if do fear version incompatibility, then i guess we should also ask here for a specfici version.

(Asking also as package.sh gives me problems with `jq` (yes, i know it is `jq`, not `yq`) and was wondering whether that problem might maybe be related to versions of jq as presumably the script does work for you ...)

g2flyer (commented):
I like the directory reorg and it seems you fixed the script problems i mentioned.  Alas, i still run into some problem with installFPC (or more specifically, package.sh), see corresponding error message. Also a few clickable suggestions to make scripts more robust and path fixes in README ...

integration/test-network/package.sh
```diff
@@ -0,0 +1,87 @@
...
+  #  "address": "ercc.org1.example.com:9999",
+  #  "dial_timeout": "10s",
+  #  "tls_required": false
+  #}
+
+  jq -n '$ARGS.named' \
+    --arg address "${address}" \
+    --arg dial_timeout "${timeout}" \
+    --argjson tls_required "${use_tls}" \
+    >>"$out_dir/connection.json"
```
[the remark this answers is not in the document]
mbrandenburger: oh interesting. Inside the dev-container I see the same error! But on mac this script works. 

In dev-docker I have:
```
jq --version
jq-1.5-1-a5b5cbe
```

On mac:
```
jq --version
jq-1.6
```

utils/docker/dev_peer_cc-builder/Dockerfile
mbrandenburger: Oh I tried to install snap in dev docker but for some reason it just did not work, so I stoped bothering with it and tried via go get. 

BTW on Mac:
```
yq --version
yq version 3.4.0
```

g2flyer: via snap on 18.04 i get `3.4.1`  which is the same i get when i do right now a go get install. So at least right now all three version differ only by patch-level which probably should be fine (contrary to jq where we differ in major version)

integration/test-network/setup.sh
```diff
@@ -0,0 +1,59 @@
...
+# Also there is another issue with this approach here. When working completely inside the FPC dev-container,
+# the volume mounts won't work. The reason is that the docker daemon provided by the host cannot parse volume paths.
+# For instance, ${FPC_PATH} inside dev-container is `/project/src/github.com/hyperledger-labs/fabric-private-chaincode/`.
+# The correct path from the docker daemon perspective is something like
+# `/Users/marcusbrandenburger/Developer/gocode/src/github.com/hyperledger-labs/fabric-private-chaincode/` (in my case on the mac).
+
+# Using DOCKERED_FPC_PATH fixes our custom volumes but still there is an issue default mounts in `docker-compose-test-net.yaml`
+# TODO need to be fixed
+
+if [[ -n "${DOCKERD_FPC_PATH}" ]]; then
```
g2flyer, line 49:
```suggestion
if [ ! -z ${DOCKERD_FPC_PATH+x} ]; then
```
so it works also if var is not defined

integration/test-network/installFPC.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Prepare deployment
+
+PKG_PATH=${FPC_PATH}/integration/test-network/_deployment
+. package.sh "${PKG_PATH}"
+
+# test network settings
+
+cd ${NETWORK_PATH} || exit
+. ./scripts/envVar.sh
```
g2flyer, line 32:
```suggestion
. ./scripts/envVar.sh
# scripts defined in fail if some vars are not defined (due to 'set -e...')
# -> make sure they are at least defined as empty (but allow override by caller)
OVERRIDE_ORG=${OVERRIDE_ORG:-}
VERBOSE=${VERBOSE:-}
```

integration/test-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Install FPC components:
+```bash
+cd $FPC_PATH/integration/test-network
+./installFPC.sh
+```
+
+Start FPC container
+```bash
+docker-compose up
```
g2flyer, line 47: this doesn't work but that's because mock_ecc doesn't build (yet) which i understand is because parts of #450 and/or #452 are missing. So we can resolve that as part of #452 ....

g2flyer (approved):
Looks good. After applying the two clickable suggested script changes, network setup and installation also works nicely.  What does not work is build of `ecc_mock`  (and hence also the start of ecc/ercc chaincode containers via `docker-compose up`) but as far as i understand, this is due to some necessary changes from #450 and/or #452, so seems best to just resolve it in #452 ...

mbrandenburger:
Squashed and ready for merging.

PR #453, opened 2020-10-24
Interface update

g2flyer:

This reflects the client<->chaincode encoding decision from thursday meeting in design docs, also updates the dispatching and adds missing protobufs for chaincode requests/responses.

Follow-up to discussion on PR #450

### DIFF
--- a/docs/design/fabric-v2+/fpc-cc-invocation.puml
+++ b/docs/design/fabric-v2+/fpc-cc-invocation.puml
@@ -39,18 +39,24 @@ group prepare FPC proposal
 CLI -> ERCC1 ++: queryChaincodeEncryptionKey(CC_Id)
 return Chaincode_EK
 
-CLI -> CLI : r_key <- keyGen()
+CLI -> CLI : ccreq_msg.rsp_key <- keyGen()
 note right
     generate a symmetric key to
     encrypt the FPC chaincode result
 end note
 
-CLI -> CLI : args_enc <- encrypt <args, r_key> with Chaincode_EK
+CLI -> CLI : ccreq_msg.encryped_request <- encrypt <args, r_key> with Chaincode_EK
 
 end
 
 group invoke
-  CLI -> Peer ++: send proposal <- <chaincodeInvoke, CC_Id, args_enc, ...>
+  CLI -> ERCC1 ++: queryListEnclaveCredentials(CC_Id)
+  return list of Credentials
+  note right
+    HostParam in Credentials allows client to identify peers which host enclaves for this Chaincode
+  end note
+
+  CLI -> Peer ++: send proposal <- <chaincodeInvoke, CC_Id, ccreq_msg>
 
   alt FPC Chaincode is not running
       note right Peer
@@ -110,7 +116,7 @@ group invoke
       return
 
 
-      Enclave1 -> Enclave1 : <args, r_key> <- decrypt args_enc with Chaincode_DK
+      Enclave1 -> Enclave1 : <args, r_key> <- decrypt ccreq_msg.encryped_request with Chaincode_DK
 
       note right
         If proposal contains optional Req_ledger_enclave_inst_id parameter,
@@ -127,14 +133,24 @@ group invoke
       end ref
 
 
-      Enclave1 -> Enclave1 : result_enc <- encrypt result with r_key
+      Enclave1 -> Enclave1 : ccrsp_msg.encrypted_response <- encrypt result with r_key
       note right
         The encryption is via authenticated symmetric cipher like aes-gcm
       end note
-      Enclave1 -> Enclave1 : Sig_Enclave <- sign <args, read/writeset, result_enc> with Enclave_SK
-      return <result_enc, Sig_Enclave>
+      Enclave1 -> Enclave1 : ccrsp_msg.signature <- sign <args, read/writeset, result_enc> with Enclave_SK
+      Enclave1 -> Enclave1 : ccrsp_msg.enclave_id <- Enclave_Id
+      return <ccrsp_msg>
 
-    return <result_enc, Sig_Enclave>
+    alt lightweight FPC
+      ccrsp_msg.rw_set <- encrypted_rw_set
+      note right
+        with chaincode-hosted validation, the stub will keep track of the 
+        R/W set (of cleartext keys but encrypted values) and pass it on to
+        the validation transaction as part of the response message.
+      end note
+    end
+
+    return <ccrsp_msg>
 
     alt  MVP
       Peer -> Peer : create proposal response <args, read/writeset, result=<result_enc, Sig_Enclave, Enclave_VK>, ...>
--- a/docs/design/fabric-v2+/fpc-registration.puml
+++ b/docs/design/fabric-v2+/fpc-registration.puml
@@ -40,7 +40,7 @@ end note
 
 Admin -> Admin: Attestation_Params <- <.e.g, SPID, sig-rl, ....>
 
-Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params)
+Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params, Peer_Endpoint)
 
 group enclave creation
   note right Peer_CLI
@@ -56,12 +56,15 @@ group enclave creation
     chaincode definition on the channel as well as in ERCC.
   end note
 
-  Peer_CLI -> ECC1 ++: initEnclave (Attestation_Params)
+  Peer_CLI -> Peer_CLI : init_enclave_msg.attestation_params <- Attestation_Params
+  Peer_CLI -> Peer_CLI : init_enclave_msg.peer_endpoint <- Peer_Endpoint
+
+  Peer_CLI -> ECC1 ++: initEnclave (init_encalve_msg)
   ref over ECC1
     This might trigger [FPC Chaincode is not running] from fpc-cc-invocation.puml
   end ref
 
-  ECC1 -> ECC1: fetch CC_Params and Host_Params
+  ECC1 -> ECC1: fetch CC_Params and Host_Params objects
   note right ECC1
     The CC_Params help binding the enclave to a specific FPC chaincode,
     as it is registered in Fabric.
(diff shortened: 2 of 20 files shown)


protos/fpc/fpc.proto
```diff
@@ -66,3 +66,36 @@ message Credentials {
...
+
+message ChaincodeRequest {
+    // function to be invoked on FPC chaincode
+    string function = 1;
+
+    // function arguments of above
+    repeated string arguments = 2;
+
+    // key to (aes128-gcm) encrypt ChaincodeResponse
+    bytes returnEncryptionKey = 3;
```
mbrandenburger, line 78:
```suggestion
    bytes response_encryption_key = 3;
```

mbrandenburger: https://developers.google.com/protocol-buffers/docs/style

We do not follow the style guide entirely in this file :/ For instance: `message Host_Parameters` We can revisit this later.

docs/design/fabric-v2+/interfaces.md
```diff
@@ -237,7 +237,7 @@ func importCCKeys() (SignedCCKeyRegistrationMessage, error) {}
 func getEnclaveId() (string, error) {}
 
 // chaincode invoke
-func chaincodeInvoke(transactionProposal []byte, ctx interface{}) (response, error) {}
+func chaincodeInvoke(transactionProposal EncryptedChaincodeRequest) (EncryptedChaincodeResponse, error) {}
```
mbrandenburger, line 240:
```suggestion
func chaincodeInvoke(request EncryptedChaincodeRequest) (EncryptedChaincodeResponse, error) {}
```

protos/fpc/fpc.proto
```diff
@@ -66,3 +66,36 @@ message Credentials {
...
+
+    // A payload that can be used to include metadata with this response.
+    bytes payload = 3;
+}
+
+message EncryptedChaincodeResponse {
+    // an aes128-gcm encryption of the serialization of ChaincodeResponse
+    bytes payload = 1;
+}
+
```
mbrandenburger, line 101:
```suggestion
message ProposalResponse {
    EncryptedChaincodeResponse encrypted_response = 1;
    bytes signature = 3;
    bytes enclave_vk = 4;

    // TODO
    // read/writeset
}
```

maybe this as well as a replacement for https://github.com/hyperledger-labs/fabric-private-chaincode/blob/master/internal/utils/utils.go#L19

mbrandenburger (approved):
LGTM. Some minor changes.

protos/fpc/fpc.proto
g2flyer: yep, realized that also overnight that we need the r/w set and the signature.  For R/W-set (and actual response probably also best to re-use existing fabric messages ...

g2flyer: btw: why would we send the enclave _key_ and not the enclave _id_ ?

protos/fpc/fpc.proto
g2flyer: i noticed also that Fabric is also slightly inconsistent. Have to admit i haven't seen the link you mentioned but reading it now it seems kindof important to follow the guide so the mapping is consistent across languates:
=>
If we ever want to make it consistent, it's probably better to do that sooner than later: should i do that already as part of this PR?

docs/design/fabric-v2+/interfaces.md
g2flyer: changed ..

protos/fpc/fpc.proto
g2flyer: see revision where i use now fabric types as appropriate

protos/fpc/fpc.proto
g2flyer: Ok, after the offline discussion, i've added the corresponding changes as a separate commit ...

protos/fpc/fpc.proto
mbrandenburger: > btw: why would we send the enclave key and not the enclave id ?

we should send the enclave_id!!!

mbrandenburger (approved):
Updated PR looks good to me.

g2flyer:
@bvavala @mbrandenburger See the last commit for the changes in protobuf and flows to reflect peer/enclave discovery

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -127,14 +133,22 @@ group invoke
...
       note right
         The encryption is via authenticated symmetric cipher like aes-gcm
       end note
-      Enclave1 -> Enclave1 : Sig_Enclave <- sign <args, read/writeset, result_enc> with Enclave_SK
-      return <result_enc, Sig_Enclave>
+      Enclave1 -> Enclave1 : enc_ccrsp_msg.signature <- sign <args, read/writeset, result_enc> with Enclave_SK
+      return <enc_ccrsp_msg>
+
+    alt lightweight FPC
+      enc_ccrsp_msg.encrypted_rw_set <- encrypted_rw_set
```
mbrandenburger, line 144: I would just call it `rw_set`. Values of the rw set are may be encrypted but not the entire rws_set. In contrast, encrypted_response is a ciphertext of a response msg. for the rw_set I think this is not the case.

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -40,7 +40,7 @@ end note
 
 Admin -> Admin: Attestation_Params <- <.e.g, SPID, sig-rl, ....>
 
-Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params)
+Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params, [Peer_Endpoint])
```
mbrandenburger, line 43: why is peer endpoint in `[]`? is it a array and can contain multiple endpoints? 

Also, I though the endpoint is "for now" a simple string, which also makes it easy to handle in the cli scripts? Below the endpoint is defined using some [fabric protos](https://github.com/hyperledger/fabric-protos/blob/master/discovery/protocol.proto#L232).

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -70,6 +73,11 @@ group enclave creation
     CC_Id, version, MSPID, and channel id can be retrieved through Fabric
     shim functions
 
+    The Host_Params has how two purposes: enable endpoint discovery for enclaves
```
mbrandenburger, line 76: The Host_Params has two purposes

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -127,14 +133,22 @@ group invoke
...
 
-      Enclave1 -> Enclave1 : result_enc <- encrypt result with r_key
+      Enclave1 -> Enclave1 : enc_ccrsp_msg.encrypted_response <- encrypt result with r_key
       note right
         The encryption is via authenticated symmetric cipher like aes-gcm
       end note
-      Enclave1 -> Enclave1 : Sig_Enclave <- sign <args, read/writeset, result_enc> with Enclave_SK
-      return <result_enc, Sig_Enclave>
+      Enclave1 -> Enclave1 : enc_ccrsp_msg.signature <- sign <args, read/writeset, result_enc> with Enclave_SK
+      return <enc_ccrsp_msg>
```
mbrandenburger, line 141: I am missing where `enc_ccrsp_msg.enclave_id` is set; maybe it's somewhere here but can't find it.

protos/fpc/fpc.proto
```diff
@@ -78,7 +90,7 @@ message ChaincodeRequest {
     bytes return_encryption_key = 3;
 }
 
-message EncryptedChaincodeRequest {
+message EncryptedChaincodeRequestMessage {
```
mbrandenburger: I think we can drop the `Encrypted` Prefix for ChaincodeRequest and Response Messages. For the request, the only field it contains already suggests that it is encrypted.

For the `ChaincodeResponseMessage` some fields, such as the signature, enclave_id, ... are in clear and only the response is encrypted.

I could imaging that if we wanna support cleartext requests and responses, we may use a simple one_of for the request.

docs/design/fabric-v2+/fpc-cc-invocation.puml
g2flyer: ok, i guess the comment in fpc.proto that it is rw set of encrypted values i guess should be enough; i guess i can add also a comment here ...

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: this was meant to denote an optional parameter. But i guess i'll just drop that as initially we probably always need it explicitly from admin ...

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: fixed ..

docs/design/fabric-v2+/fpc-cc-invocation.puml
g2flyer: well, we didn't before; added now ...

protos/fpc/fpc.proto
g2flyer: i see your point and also had similar feelings but when dropping the `Encrypted` it differs from the cleartext only by `Message` which is also not nice. I solved that now by making the inner one explicitly a Cleartext Request ..

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: > why is peer endpoint in `[]`? is it a array and can contain multiple endpoints?

this was meant to denote an optional parameter. But i guess i'll just drop that as initially we probably always need it explicitly from admin ...

> Also, I though the endpoint is "for now" a simple string, which also makes it easy to handle in the cli scripts? Below the endpoint is defined using some [fabric protos](https://github.com/hyperledger/fabric-protos/blob/master/discovery/protocol.proto#L232).

We didn't talk about format and it seemed natural to use the fabric definition.   as long as we have an `initEnclave` message then i don't think that using a string naturally makes it fundamentally?

mbrandenburger (approved):

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -40,7 +40,7 @@ end note
 
 Admin -> Admin: Attestation_Params <- <.e.g, SPID, sig-rl, ....>
 
-Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params)
+Admin -> Peer_CLI ++: lifecycle chaincode\n createenclave (CC_Id, Attestation_Params, Peer_Endpoint)
```
bvavala, line 43: I don't have a better solution for this "peerendpoint", but it would certainly be better if the peer itself could supply it -- rather than an external entity.

bvavala (approved):

PR #455, opened 2020-10-28
Various client-sdk related improvements

g2flyer:

* implement callout to attestation-to-evidence conversion in admin api and corresponding changes in tests/mock
* various fixes that a `make` works (and hence travis should now succeed)
* as part of that removed obsolete ercc validators and decorators (and related references in fabric patch)
* some script and docu improvements for test-network

In my testing i can now successfully setup test-network and run `client_sdk/go/test/test`. Note though it will fail by design as we havent' implemented yet the validation transaction.  Also, not sure whether this is an output bug but it seems the `QueryListEnclaveCredentials` call to ercc doesn't seem to return the registered enclave. ...

```


g2flyer:
Wrong branch ...

PR #457, opened 2020-11-02
Update Spec with FPC Lite variant

g2flyer:

This PR should do the necessary changes to match FPC Lite as outlined in [FPC externalized endorsement validation](https://docs.google.com/document/d/1RSrOfI9nh3d_DxT5CydvCg9lVNsZ9a30XcgC07in1BY/edit#heading=h.3moiev8yvxkn) and [FPC without Trusted Ledger](https://docs.google.com/document/d/1jbiOY6Eq7OLpM_s3nb-4X4AJXROgfRHOrNLQDLxVnsc/).

Also as discussed in review for this PR, it changes the enclave signature to cover the whole (signed) proposal, not only the args.  See also related issue #460.

This should hopefully cover the design changes we have to do in our MVP implementation and should also be able to serve for the RFC where we can (in an appendix) punt all design and internals to these UMLs and .mds, the two googlde docs on TL-less and external validation as well as referencing the code ...

In some cases there is a considerably amount of white-space changes but the files where rather inconsistent before and for easier edititing i normalized them ...

```

### CODE REVIEW

File: docs/design/fabric-v2+/fpc-cc-invocation.puml [hunk 1]
```diff
@@ -141,35 +146,35 @@ group invoke
       Enclave1 -> Enclave1 : ccrsp_msg.enclave_id <- Enclave_Id
```
mbrandenburger, line 146: Since the enclave creates the endorsement signature using `args, read/writeset, and result_enc` we need to ensure that all these values are available when chaincode-hosted endorsement is performed. In particular, our current flow misses the `args`. With traditional custom validation, we have full access of the transaction proposal, which contains the entire proposal. With our new validation, we only send the `ccresp_msg` to validation, hence, we are missing the `args`.

I suggest that we just add the `proposal` to `ccresp_msg`.

mbrandenburger: Also, instead of signing just the `args`, why don't we include the entire proposal in the signature?

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
```diff
@@ -92,17 +92,21 @@ group approveformyorg
...
+    NOTE: the (fabric) endorsement policy has different meanings depending
+    on whether FPC runs in "light" (MVP) or "full" mode (post-MVP):
+    - in former case it specifies "validation endorsement policy", i.e. 
+      how many organization have to be trusted to validate the enclave 
+      endorsements in the validation chaincode transaction. For MPV, the 
+      "enclave endorsement policy" is implicit and requires an enclave
+      signature from a single enclave registered at ERCC and hosted at
+      an organization participating in the underlying channel.
+    - in the latter case, it specifies the "enclave endorsement policy",
+      i.e., how many (and with whic proporties) enclaves are expected
```
mbrandenburger, line 106:
```suggestion
      i.e., how many (and with which properties) enclaves are expected 
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -52,88 +52,107 @@ group endorsement
...
+                ECC1 -> ECC1 : verify Sig_Enclave over <args, read/writeset, ECC_result> with Enclave_VK
+                ECC1 -> ECC1 : Enclave_Id <- Hash(Enclave_VK)
+                create participant ERCC order 30
+                ECC1 -> ERCC++ : queryEnclaveCredentials(CC_Id, Enclave_Id)
+                    note right
+                        If FPC enclave is registered, ERCC will
+                        return the corresponding Credentials,
+                        otherwise Credentials is empty.
+                    end note
+                return Credentials
```
mbrandenburger, line 70: here some minor changes are needed. From `ccresp_msg` we already get `Enclave_Id`, so no need to hash `Enclave_VK`. In particular, we also first need to fetch `Enclave_VK` from ERCC before we can verify the signature.

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -52,88 +52,107 @@ group endorsement
...
+                ECC1 -> ECC1 : extract CC_Params from Credentials.attested_data
+                ECC1 -> ECC1 : check CC_Params matches chaincode definition
+                ECC1 -> ECC1 : check CC_Params.MSPID matches MSPID of endorser
```
mbrandenburger, line 73: > check CC_Params.MSPID matches MSPID of endorser

Is this still the right thing to do with FPC light?

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -52,88 +52,107 @@ group endorsement
     end ref
     return proposal response
 
+    alt FPC "light"
+        User1 -> ECC1 ++: invoke "validateEnclaveEndorsement" (proposal response)
+            group enclave endorsement validation
+                ECC1 -> ECC1 : <args, read/writeset, result> <- extract from transaction
+                ECC1 -> ECC1 : <Enclave_VK, Sig_Enclave, ECC_result> <- extract from result
```
mbrandenburger, line 59: Should we replace `proposal response` with `ccresp_msg` here? 

As mentioned before, without the "full" proposal response from the actual chaincode invocation we are missing the args. I would just add the entire (signed and serialized) proposal to `ccresp_msp` and we have everything we need.

mbrandenburger (changes requested):
This looks good to me. I found some minor things I think we should change.

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -141,35 +146,35 @@ group invoke
...
           This step is performed using an FPC-specific endorsement plugin.
       end note
-    end
+  end
 
-    return proposal response
+  return proposal response
 end
 
 group complete transaction
```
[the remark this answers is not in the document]
mbrandenburger: a note, or alt should be added

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -141,35 +146,35 @@ group invoke
...
 
-    alt lightweight FPC
-      ccrsp_msg.rw_set <- encrypted_rw_set
-      note right
-        with chaincode-hosted validation, the stub will keep track of the 
-        R/W set (of cleartext keys but encrypted values) and pass it on to
-        the validation transaction as part of the response message.
-      end note
+     alt FPC "light"
+       ECC1 <-> ECC1 : ccrsp_msg.rw_set <- encrypted_rw_set
```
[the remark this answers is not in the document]
mbrandenburger: without this change the diagram is not correctly rendered (on my machine)

docs/design/fabric-v2+/fpc-cc-invocation.puml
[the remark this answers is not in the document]
g2flyer: i added the validation transactions inside the validation.uml, where it would seem more natural. But it is true the handover between cc-invocation and fpc-validation has to be fixed. It essentially means referring to enough enclave endorsements and moving the submit transaction to validation

docs/design/fabric-v2+/fpc-cc-invocation.puml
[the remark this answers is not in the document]
g2flyer: will change, but curious in what respect did it not correctly render?  When i test via visual code, the two cases are indistinguishable ...

docs/design/fabric-v2+/fpc-cc-invocation.puml
g2flyer: good find: besides obviously needing the parameters to verify, we also should sign the proposal to capture also the creator which can influence the chaincode and hence should be audited ..

docs/design/fabric-v2+/fpc-lifecycle-v2.puml
g2flyer: will fix ...

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: good find, order of the two steps has to be swapped and no hash computation needed

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: it's misleading text. It has to be the identity of the transaction creator (which essentially represents as an (conceptual, not Fabric) endorsement of this enclave to represent its organization. Will rephrase ...

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: well, seeing it now here you actually _would_ have received everything according to the UMLs. But it wouldn't have matched interfaces.md and having `FPC_Stub` add it to `ChaincodeResponseMessage` rather than having the client send it separately in the validation transaction is certainly the easier way to solve it.

docs/design/fabric-v2+/fpc-cc-invocation.puml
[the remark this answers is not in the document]
mbrandenburger: `ccrsp_msg.rw_set` and  `encrypted_rw_set` were rendered as new entities on the left hand side of the diagram. there was also a line from `ECC` to `ccrsp_msg.rw_set`.

g2flyer:
@mbrandenburger hope the fixup i just pushed addresses your concerns

docs/design/fabric-v2+/fpc-validation.puml
mbrandenburger: > ECC1 -> ECC1 : check CC_Params.MSPID matches MSPID of proposal creator & creator has role Admin

I think something is wrong here. With this check we enforce that only members of the org owning the enclave and members that have admin rights can invoke FPC transactions. 

What we want:
- only admins can create fpc enclaves
- every org in the channel can invoke the fpc chaincode hosted by any org

mbrandenburger:
> @mbrandenburger hope the fixup i just pushed addresses your concerns

Thanks Michael, all changes are good! The only open concern is regarding MSPID check at FPC transaction validation. See comment above.

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: oops, good catch. This seems to have been wrong for quite a while already before my changes; i just was thinking of delta from your earlier review comment. Will fix ...

g2flyer: Good catch. This is wrong and now thinking not only about the delta from your earlier comment, i realize that this was wrong even beforehand. CC_Params doesn't have MSPID, it's now in credentials.   So i guess one could conceptually try verify that the signature from the enclave still comes from an endorsing peer consistent with the registration. But this will be more challenging now in externalized validation and i also don't think the test is really necessary: as part of the registration an organization has taken ownership, i.e., consider itself accountable, and so any mischief of that enclave might be laid blame on the org, making it in their interest to protect the enclave (keys) well
=>
Propose to just drop that line?

mbrandenburger: drop it!

g2flyer: just pushed a fixup.  Realized there were also a few other inconsistencies which are hopefully now fixed (or at least improved ;-))

mbrandenburger (approved):
Thanks! LGTM!

docs/design/fabric-v2+/fpc-registration.puml
```diff
@@ -98,27 +107,33 @@ group enclave creation
...
-    return
+  group attestation preparation
+    alt FPC "light"
+      Enclave1 -> Enclave1 : AttestedData <- <CC_Params, Host_Params, Enclave_VK, Channel_hash, onCreation=true>
+      note right Enclave1
+      In the future an admin might want to request re-attestations: For EPID this would be
+      the only way to check for revocations (In DCAP, checking CRLs would enable this as well).
+      For security reasoning, it is crucial that the TCB level (CPUSVN) of the seal operation is
+      reflected in the attestation (and, hence, ERCC registration) as this determines whether
+      enclave secrets might be exposed in case of newly discovered future SGX vulnerabilities.
```
bvavala, line 118:
```suggestion
      reflected in the attestation (and, hence, ERCC registration).
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -43,97 +37,127 @@ ref over User1, Orderer
...
+                        note right
+                            If FPC enclave is registered, ERCC will
+                            return the corresponding Credentials,
+                            otherwise Credentials is empty and validation will abort.
+                        end note
+                    return Credentials
+                    ECC1 -> ECC1 : extract Enclave_VK from Credentials.attesated_data
+                    ECC1 -> ECC1 : verify Sig_Enclave over <proposal, read/writeset, ECC_result> with Enclave_VK
+                    ECC1 -> ECC1 : extract CC_Params from Credentials.AttestedData
+                    ECC1 -> ECC1 : check CC_Params matches chaincode definition
```
bvavala, line 66:
```suggestion
                    ECC1 -> ECC1 : check CC_Params matches chaincode definition stored in Credentials
```

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -43,97 +37,127 @@ ref over User1, Orderer
...
-    deactivate Orderer
+Orderer -> ECC1+ : broadcast
+& Orderer -> ECC2+ : broadcast
+& Orderer -> ECC3+ : broadcast
+deactivate Orderer
 
-    group validation & commit
-        ECC1 -> ECC1 : validate block signature
-        loop for each tx in block
+group validation & commit
```
bvavala, line 104: This entire block is "Post-MVP", as we have no FPC-logic built into the peer.
So, there is nothing to specify here.
```suggestion
group Post-MVP: validation & commit
```

bvavala (changes requested):
Minor clickable changes.

docs/design/fabric-v2+/fpc-validation.puml
```diff
@@ -43,97 +37,127 @@ ref over User1, Orderer
...
-
-        loop for each tx in block
-           ECC1 -> ECC1 : apply valid tx writeset
-        end loop
-
-        ECC1 -> ECC1 : append block to ledger
 
+                ECC1 -> ERCC++ : queryEnclaveCredentials(CC_Id, Hash(PeerCertificate.Public_key))
+                return Credentials
+                fail if no credential is return
```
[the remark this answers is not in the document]
g2flyer: oops, i guess i overlooked the red ! in vs code (maybe also reason to put documation build back into make and .travis ;-)
should be fixed now

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -137,74 +142,69 @@ group invoke
       note right
         The encryption is via authenticated symmetric cipher like aes-gcm
       end note
-      Enclave1 -> Enclave1 : ccrsp_msg.signature <- sign <args, read/writeset, result_enc> with Enclave_SK
+      Enclave1 -> Enclave1 : ccrsp_msg.signature <- sign <proposal, read/writeset, result_enc> with Enclave_SK
```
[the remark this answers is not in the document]
g2flyer: There is an implied hash in `sign`.  In the code we might actually double hash args (no change to poposal yet) but not what double-hashing really would buys us?  Certainly doesn't change security (at least in non-academic ways) and for omitting it here certainly makes it clearer and simple (which i guess you agree from the second part of your comment)

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -137,74 +142,69 @@ group invoke
...
+       end note
+       ECC1 -> ECC1 : ccrsp_msg.proposal <- proposal
     end
 
     return <ccrsp_msg>
 
-    alt  MVP
-      Peer -> Peer : create proposal response <args, read/writeset, result=<result_enc, Sig_Enclave, Enclave_VK>, ...>
+  alt  MVP
+    Peer -> Peer : create proposal-response <proposal, read/writeset, result=ccrsp_msg, ...>
```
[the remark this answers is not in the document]
g2flyer: Note this is the (vanilla Fabric) Peer, not FPC_stub (the untrusted part of our stub/shim), the response payload is ccrsp_msg.  I've added though now also an explicit remark to the comment that this is vanilla fabric ..

docs/design/fabric-v2+/fpc-cc-invocation.puml
```diff
@@ -137,74 +142,69 @@ group invoke
...
       end note
-    end
+  end
 
-    return proposal response
-end
+  return proposal-response
 
-group complete transaction
+  loop for all/"enough" enclave endorsers
```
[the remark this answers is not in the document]
g2flyer: Well, it's just with FPC Lite (and current implicit enclave endorsement policy), you have enough after a single one, so not really wrong as is.

docs/design/fabric-v2+/fpc-integration-touchpoints.puml
```diff
@@ -76,44 +79,61 @@ group Per FPC chaincode FCC
...
-        Peer2 <- ecc2
-        & Client1 <- Peer2 : endorsement Peer2 & result
-        Client1 -> Orderer : {endorsement Peer1, endorsement Peer2}
-        Peer1 <- Orderer : ordered {endorsement Peer1, endorsement Peer2}
-        note over Peer1
-  	  Validate enclave signatures, endorsement policies evaluated
-          over enclave signatures, not peer signatures!
-        end note
-        Peer2 <- Orderer : ordered {endorsement Peer1, endorsement Peer2}
+        & Client1 <- Peer1 : endorsement Peer1, enclave signature & result
```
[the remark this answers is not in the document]
g2flyer: that is my understanding.  Was this just some question or a suggestion for a change? If the latter, not sure what you meant.  Did you mean adding "endorsemt Peer1" also further up on line 100? (we could be for easier reading i would prefer not mentioning it, in particular as nobody makes any use of it?

docs/design/fabric-v2+/fpc-registration.puml
g2flyer: done

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: `CC_Params` are already extracted from `Credentials`, so that test would always be true. What we have to test that CC_Param matches ECCs view of the chaincode definition. I've handled that by adding a "own" after "matches".

PS: made me realize that we somewhat mix up peer and fpc-stub here, maybe also leading to your confusion below regarding fpc validation. Split now the two functionalities.

docs/design/fabric-v2+/fpc-validation.puml
g2flyer: not really true: even for FPC Lite the validation and commit has to happen, just only in the `default validation` group, not in `fpc-vscc`, and the final `apply valid tx writeset` (but not in TL)

docs/design/fabric-v2+/fpc-integration-touchpoints.puml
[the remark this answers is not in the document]
bvavala: > Did you mean adding "endorsemt Peer1" also further up on line 100?

Yes, no big deal.

docs/design/fabric-v2+/fpc-validation.puml
bvavala: > not really true: even for FPC Lite the validation and commit has to happen

of course it has to, but that's vanilla Fabric, so there is nothing here that add or remove.
Anyway, minor.

PR #462, opened 2020-11-06
update nanopb 0.4.3

g2flyer:

nanopb 0.4.3 fixes the issue in 0.4.2 which prevented us to go to (then) latest version in PR #436, fixes numerous bugs, brings some performance optimization and, for the C++ fans among us, provides `Better support C++ types in generated structs (#577)`
For more info see [CHANGELOG](https://github.com/nanopb/nanopb/blob/master/CHANGELOG.txt)

To test you will have to update your nanopb setup:
```
cd  $NANOPB_PATH
git fetch --all --prune
git checkout nanopb-0.4.3
cd generator/proto
rm -f nanopb_pb2.py*
make
```
and then also make sure that do `make clean`, at least in 'protos' dir ..

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ Make sure that you have the following required dependencies installed:
 
 * Protocol Buffers
     - Protocol Buffers 3.0.x needed for the Intel SGX SDK
-    - Protocol Buffers 3.11.x or higher and [Nanopb](http://github.com/nanopb/nanopb) 0.4.1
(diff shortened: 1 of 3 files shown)


mbrandenburger (approved):

bvavala:
As discussed, let's table this for discussion, since the change makes it difficult to work with this branch and master at the same time.

Issue #411, opened 2020-07-20
Upgrade SGX SDK and SSL


mbrandenburger:
**Description**
Upgrade the SGX SDK and SSL used in FPC. Currently, FPC uses the Intel SGX SDK v2.6 for linux and SSL for SGX SDK v2.4.1. 

Most recent available version of the SGX SDK is v2.10. See https://github.com/intel/linux-sgx/releases and https://github.com/intel/linux-sgx/releases

**Link to feature or bug**

**Link to dependent issues**

g2flyer:
Intel just released v2.12.   The main noteworthy feature is that now Ubuntu 20.04 is also supported. Nothing else noteworthy which should push to upgrade.   In some quick tests, `make` for `flow-refactoring` works with a 2.12 machine; running `make` on `cr2.0.0` will fail when running the test in `demo/chaincode/fpc`, but that seems to be an issue which also happens with earlier sgx versions.

A few caveats:
- `apt update` -- which (semi-)automagically upgrades to the latest sgx packages -- can result in a broken upgrade, a simple `apt --fix-broken install` does provide remedy, though.
- it seems on FLC-enabled NUCs, you now need the dcap driver (now as version `1.36.2`), the old-style driver, now as version `2.11`) does not seem to work anymore.  Note, though, that [download site](https://download.01.org/intel-sgx/sgx-linux/2.12/distro/ubuntu18.04-server/) provides both.
- there is also a newer version of sgxssl  ( `lin2.11_1.1.1.g` vs currently used  `lin2.10_1.1.1.g`). Nothing important for us to upgrade; in fact right now upgrade wouldn't work as some pthread stuff changed which make pdo crypto fail on that version.
-  regarding 20.04, a quick fix (PR #463, matching a similar change for PDO) makes all our docker images build.

g2flyer:
fyi: [SGX SDK&PSW 2.13](https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-linux-2.13-release) got released. From release notes perspective there should be only bug-fixes as far as FPC use is concerned. As we haven't noticed any issue i guess there is no reason force version increase nor should there be an issue if somebody uses a newer version ...

PR #463, opened 2020-11-12
Dockerfile fix for 20.04

g2flyer:
Small fix to dockerfile so they work also with 20.04.  
`DOCKER_BUILD_OPTS="--build-arg SGX=2.12 --build-arg UBUNTU_VERSION=20.04 --build-arg UBUNTU_NAME=focal" make -C utils/docker`
should build all images from `utils/docker` successfully based on 20.04

See comment on issue #411 for some context.


mbrandenburger:
Uh this is nice! Though, we may run into similar issues as @bvavala pointed out in #462 that switching between master and flow development may be cumbersome.

g2flyer:
> Though, we may run into similar issues as @bvavala pointed out in #462 that switching between master and flow development may be cumbersome.

This should not affect switching between branches (at least not more than what you have to inevitably do,  i.e., rebuild). It also should be backwards compatible if need would arise to backport.  In any case, master is effectively frozen before renitegrating flow-refactoring so there are not really any reason to go back and forth?   That said, this PR is obviously not urgent, it was just an artifact of stumbling yesterday over the SGX version update and the related apt update break and then multi-tasking during various meetings on assessing the impact of it...

mbrandenburger (approved):
LGTM! Thanks!

PR #468, opened 2020-11-18
Godoc for SDK functions

g2flyer:

This adds godoc comments for the Client and Management SDK API functions we reference in the RFC.
Also includes some minor refactoring of api to make more uniform, minimize exports, closer to what Fabric does (e.g., no Interface suffix for interfaces) and get rid of some go-lint warnings.
Also added two missing .gitignores and a convenience .gitignore for the case you have a local copy of the helloworld example

```


mbrandenburger (approved):
LGTM! Thanks Michael

mbrandenburger:
Travis check complains about missing license headers in `.gitignore`, we can ignore this.

mbrandenburger:
@bvavala any objections to merge? :)

g2flyer:
> Travis check complains about missing license headers in `.gitignore`, we can ignore this.

Strangely, it didn't complain when i tested in my VM.  But added it now also as exception to the licence check script

bvavala (approved):
need to add some more exceptions for .gitignore

g2flyer:
> need to add some more exceptions for .gitignore

I too naively copy/pasted from the `.gitmodules` rule which i've added during the re-integration to solve another licence-header issue. Removed now the preceeding '^' made the rule match only on the top-level. Hopefully now it goes through ....

PR #473, opened 2020-11-25
Fix Go refs for non-dockerized environments

bvavala:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ addons:
   apt:
     packages:
 
-go_import_path: github.com/hyperledger-labs/fabric-private-chaincode
-
 env:
   global:
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 dist: bionic
-language: go
-
-go:
-  - 1.14.4
 
 sudo: required
 services:
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ Make sure that you have the following required dependencies installed:
 
 * CMake v3.5.1 or higher
 
-* [Go](https://golang.org/) 1.14.4 or higher
+* [Go](https://golang.org/) 1.15.4 or higher
 
 * Docker 18.x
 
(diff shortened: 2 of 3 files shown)


g2flyer:
if we require it everywhere, we probably should also change `go.mod` which still references `1.14`?

.travis.yml
```diff
@@ -3,10 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 dist: bionic
-language: go
-
-go:
-  - 1.14.4
```
g2flyer, line 9: If we remove this (which makes sense as long as we do only docker build) then we probably should also remove below `go_import_path`? 

BTW: as this is far away from a change, i'can do clicky suggestion. So seems you can do clicky suggestsions. Seems while you can do them outside of changes, it can only be done at most two lines away from a change?

bvavala: done

bvavala (commented):
> if we require it everywhere, we probably should also change `go.mod` which still references `1.14`?

I have tried that, but it fails -- more precisely, I had also updated the go mods.
Problem should be that we use 1.15 but Fabric is still at 1.14.

g2flyer:
> > if we require it everywhere, we probably should also change `go.mod` which still references `1.14`?
> 
> I have tried that, but it fails -- more precisely, I had also updated the go mods.
> Problem should be that we use 1.15 but Fabric is still at 1.14.

In that case, should we really bump the requirement up to 1.15?  (i've actually asked that question to Marcus on his change ;-)   Note even with Fabric 2.3.0, Fabric is still with 1.14 (although bumpted it now to 1.14.12 ...

g2flyer:
> In that case, should we really bump the requirement up to 1.15? (i've actually asked that question to Marcus on his change ;-) Note even with Fabric 2.3.0, Fabric is still with 1.14 (although bumpted it now to 1.14.12 ...

Actually, it seems we _do_ need now 1.15 due to a new `ecdsa.VerifyASN1` we are using now? Hmm, ....

bvavala:
> Actually, it seems we _do_ need now 1.15 due to a new `ecdsa.VerifyASN1` we are using now? Hmm, ....

That's correct.
Anyway, just tested go.mod upgrade (without upgrading the packages). Works fine.

g2flyer (approved):

PR #474, opened 2020-12-01
Make fabric-samples work again and easier to use

g2flyer:
test-networks test do not work in master. This is PR fixes some obvious issues, updates docu and also adds a patch to the test-network network.sh which makes it easy to look at peer logs and alike.   It should also display now logs from cgo.

PS: it's strange, in a simple test program, the `puts` used in cgo does display info on stdout, but not for `ercc` where puts output is simply swallowed and you have to explicitly set the output callback to something else.  
BTW: given that we used flogging so far, why are we using now "log"? Would be nice to be somewhat consistent ....

PR #472, opened 2020-11-25
chaincode ek message/flow merge

bvavala:
Signed-off-by: Bruno Vavala <bruno.vavala@intel.com>

PR #475, opened 2020-12-01
Cleanup + bug fix + re-enable (sim-mode only) integration tests

bvavala:
This PR 
* does some cleanup (remove cmac code, disable fabric patch)
* fixes some bugs
* re-enables the integration tests (temporarily, waiting to switch to the test network)


ercc/main.go
```diff
@@ -34,39 +34,42 @@ func main() {
...
+	addr := os.Getenv("CHAINCODE_SERVER_ADDRESS")
 
-	server := &shim.ChaincodeServer{
-		CCID:    config.CCID,
-		Address: config.Address,
-		CC:      ercc,
-		TLSProps: shim.TLSProperties{
-			Disabled: true,
-		},
-	}
```
mbrandenburger, line 50: This is good; instead of wrapping the code that comes after in an `if` we could just panic/exit here if `ccid` and `addr` are not set. This makes the change set smaller and code easier to read.

ercc/attestation/verifier.go
```diff
@@ -37,7 +37,7 @@ func (v *VerifierImpl) VerifyEvidence(evidenceBytes, expectedStatementBytes []by
 
 	expectedMrEnclaveBytes := []byte(expectedMrEnclave)
 	expectedMrEnclavePtr := C.CBytes(expectedMrEnclaveBytes)
-	defer C.free(expectedStatementPtr)
+	defer C.free(expectedMrEnclavePtr)
 	expectedMrEnclaveLen := len(expectedMrEnclaveBytes)
```
mbrandenburger, line 41: good catch! #474 contains the fix already.

mbrandenburger (approved):
LGTM. Please rebase and see suggestion or `ercc/main.go`

ercc/main.go
bvavala: I would agree but cannot panic (check out the code below): if those parameters are not set, the code still proceeds with the original initialization.

ercc/attestation/verifier.go
bvavala: The "update branch" automatically solved the conflict.

ercc/main.go
```diff
@@ -34,39 +34,42 @@ func main() {
...
-	//}
-	//
-	//if err := ercc.Start(); err != nil {
-	//	log.Panicf("Error starting registry chaincode: %v", err)
-	//}
+	if len(ccid) == 0 && len(addr) == 0 {
+		// alternatively we can start the chaincode in the normal way and let it connect the its peer
+		// TODO integrate the code below
+		// some switch is needed (e.g., via --arg or ENV_VAR) to start the chaincode in one or the other mode
+		// note that this code will also be shared with ecc
```
g2flyer, line 67: This is obsoleted by your change!
```suggestion
		// start the chaincode in the traditional way
```

fabric/Makefile
```diff
@@ -90,9 +90,9 @@ $(FABRIC_PEER_BIN): $(TLCC_GO_SRC) $(TLCC_ENCLAVE_LIB) $(FABRIC_PATCHED)
 	   GO_TAGS=FPC \
         $(MAKE) peer
 
-native: peer
+native:
 	cd $(FABRIC_PATH) && \
-	GOFLAGS="-mod=mod" $(MAKE) -j orderer cryptogen configtxgen
+	GOFLAGS="-mod=mod" $(MAKE) -j orderer cryptogen configtxgen peer
```
g2flyer, line 95: Are you sure we do not need the proxy patch anymore? I think it is still needed for the integration test which mixes fpc with standard go chaincode, the latter being directly run by peer (at least if you don't yet have the corresponding docker image i think you could run into trouble behind a proxy)?  In that case, might be easier to leave Makefile as-is and just `git mv` every patch except proxy to `fabric/patches.disabled/`?

g2flyer (commented):
Looks good and is great to have integration test back, even without travis and even if right now they don't do yet validation and work only in sim.  That can be handled in separate PRs afterwards.

A number of suggestions below, some of them clickable.  The main concern i have whether we might not need for intel folks the proxy patch in some circumstances?

Also, `make linter` fails for me. 
```
INT: Running code checks for Go files...
Checking with gofmt
stat : no such file or directory
Makefile:31: recipe for target 'linter' failed
make: *** [linter] Error 2
```
Not immediately clear now why, but i haven't really investigated much. from `bash -vx scripts/golinter.sh` it seems as if the script tries to run `gofmt` on the empty string ...
```
+ read -r source_dir
go list -f '{{.Dir}}' ./... | sed s,"${fabric_dir}".,,g | cut -f 1 -d / | sort -u
++ go list -f '{{.Dir}}' ./...
++ sed s,/home/msteiner/src/poet-et-al/go-trusted-chain-code.v2.0/src/github.com/hyperledger-labs/fabric-private-chaincode.,,g
++ cut -f 1 -d /
++ sort -u
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir
+ source_dirs+=("$source_dir")
+ IFS='
'
+ read -r source_dir

echo "Checking with gofmt"
+ echo 'Checking with gofmt'
Checking with gofmt
OUTPUT="$(gofmt -l -s "${source_dirs[@]}")"
++ gofmt -l -s '' client_sdk demo ecc ecc_mock ercc internal tlcc utils validation
stat : no such file or directory
+ OUTPUT=
```

fabric/Makefile
bvavala: Good point, I have not run it in HW mode.
* For IAS, we don't need it anymore. Since that part has moved to the client side.

* For standard chaincode, I guess that would allow us to install any package in the container. However, the solution is pre-external-builder. Should we move to an external builder for regular chaincodes behind proxy?

g2flyer: i don't think it's worth to go to the effort of doing so. Given that for integration tests we anyway compile our own peer, i wouldn't worry about that patch added. Certainly the least effort for us (by leaving makefile as is and a simple git mv (or git rm) to (re)move the non-wanted patches?

integration/auction_test.sh
```diff
@@ -32,20 +32,12 @@ auction_test() {
 
     PKG_ID=$(${PEER_CMD} lifecycle chaincode queryinstalled | awk "/Package ID: ${CC_ID}/{print}" | sed -n 's/^Package ID: //; s/, Label:.*$//;p')
 
-    # first call negated as it fails due to specification of validation plugin
-    try_fail ${PEER_CMD} lifecycle chaincode approveformyorg -o ${ORDERER_ADDR} -C ${CHAN_ID} --package-id ${PKG_ID} --name ${CC_ID} --version ${CC_VER} --sequence ${CC_SEQ} --signature-policy ${CC_EP} -E mock-escc -V fpc-vscc
```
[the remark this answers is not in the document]
bvavala: We did redefine it, but still we cannot really prevent the use of the plugins (without modifying Fabric/client).

Anyway, no strong opinion keeping it.

fabric/Makefile
bvavala: sounds good, though for development in this repo.
In the future we should really move to the external builder.

g2flyer: fabric is not moving to external builder by default for legacy chaincode, so why should we spend effort on it?

bvavala: Because, before, Fabric lacked a way to deal with proxies, now it does not.

Alternatively, we could try to push the patch to Fabric.

g2flyer: still seems to me rather low on the priority list, at least as long as the current patch works for us as-is? (i definitely would agree though that it's not worth spending any time maintaining/updating the patch)

ercc/main.go
bvavala: > instead of two if might be more robust to do a if .. elsif .. else ..

fixed

bvavala:
with the latest commits, everyone should be now satisfied with this PR.
I am going to merge to proceed with #472

PR #480, opened 2020-12-05
peer cli assist & integration test in HW mode

g2flyer:

This PR provides a small utility which assists the peer cli with message manipulations based on the client sdk code. So far it implements the credential conversion in `initEnclave` and correspondingly simplifies `peer.sh` , makes it work in HW^1 and also ensures what there is a single place which does the actual message manipulation.  It also defines stub for commands for the peer cli to hand off (complicated) client request and response processing.

^1: the major challenge here was that we used inconsistent hex encodings. Took me probably as much time as doing all the rest to figure that out ... :-(


common/crypto/attestation-api/evidence/verify-evidence.cpp
```diff
@@ -187,6 +187,8 @@ bool verify_ias_evidence(
                  extract_hex_from_report(ias_report,
                      offsetof(sgx_quote_t, report_body) + offsetof(sgx_report_body_t, mr_enclave),
                      sizeof(sgx_measurement_t), hex_id));
```
bvavala, line 189: should we put here a normalization step (toUpper or to Lower) for both terms?

bvavala (approved):
Very nice, works like charm.
Just a comment, oth. good to do.

common/crypto/attestation-api/evidence/verify-evidence.cpp
g2flyer: i documented it now in the .md as supposed to be upper-bound.  if we do normalization, i think we would have to do it in peer cli and management API; doing it only here might still bring up issues elsewhere if you confuse cases (e.g., consider org1 installs lowercase, the other one uppercase; from an ERCC case they would now be considered the same but from a ledger perspective they are different!)

bvavala: I understand, but that upper-case/lower-case encoding is irrelevant -- security-wise -- and just raises implementation issues.
We can discuss that, and I can do the change if we decide so.

g2flyer: irrelevant security-wise, yes; but functionality-wise it's not as fabric is case-sensitive (at least i thought) for versions.   In that case i don't really see how we could normalize, certainly not transparent to admins? (non-admins should never see the version). But we can talk tomorrow in the call if you want ...

PR #481, opened 2020-12-07
Purge more stuff not necessary in FPC 1.0 (aka FPC Lite)

g2flyer:

One step more along the agreement of last week to purge FPC "Full" stuff.
More specifically, it does remove tlcc* and validation plugins, some related fabric makefile stuff. It also removes `demo` and `utils/docker-compose` (with the included but anyway outdated node-sdk code)

Other potential clean-up actions which probably require further discussions:
- docu (see issue #487 for separate issue on that)
- stuff in `ecc*`. This needs coordination with `ecc`-`ecc_mock` merge; also seems to contain other obsolete code around "setup" and code in `ecc/crypto/` obsoleted by new code in `internal` ...
- remove of auction example (not really supported, but our "main" test program in integration test,
  so we would have to replace it with something else, we cannot just remove it). There is now a separate issue #491 for that ....
- a number of what is in `common`, e.g., `common/enclave`, is really more and more `ecc` specific and might better move there eventually. But certainly no priority and probably best done at a point we have to do anyway some refactoring with that code ...

*note: look at `pre-fpc-lite` & `ccr2.0.0` if you want pre-fpc-lite/fpc 1.0 code with original trusted-ledger and validation code ...)*

Fixes #


bvavala (approved):
Nice!

g2flyer:
travis failed, not yet 100% sure why (i actually had tested locally a `make clobber all` ...) but i noticed there is still tons of other tlcc stuff in `ecc*` to be rooted out. So moved to DRAFT mode for now ...

ecc_enclave/enclave/enclave.cpp
```diff
@@ -22,11 +22,6 @@ extern sgx_ec256_public_t enclave_pk;
 // this is tlcc binding
 sgx_ec256_public_t tlcc_pk = {0};
 
-// TODO: replace with secure key generated in `common/enclave/cc_data.cpp` when we also create
-// cc-encryption/decryption key-pair .. state encryption key; hardcoded for debugging
-sgx_aes_gcm_128bit_key_t state_encryption_key = {
```
bvavala, line 27: I guess you squashed a bit too much. Luckily it does not conflict!

bvavala (approved):

ecc_enclave/enclave/enclave.cpp
g2flyer: Arghh, good catch. turns out i screwed up in rebase and merged my changes into your commit (and also forgot to add some changes i had locally).
=> don't merge yet but wait for force-push below (and corresponding successful travis build)

PR #482, opened 2020-12-08
Enforce single enclave registration in ERCC

bvavala:
(self-explanatory)

This PR depends on #480 .

### DIFF
--- a/ercc/registry/registry.go
+++ b/ercc/registry/registry.go
@@ -53,6 +53,10 @@ func (rs *Contract) QueryListEnclaveCredentials(ctx contractapi.TransactionConte
 	if err != nil {
 		return nil, err
 	}
+	if iter == nil {
+		// return empty list, no error
+		return nil, nil
+	}
 
 	// note that we store serialized credential objects in the chaincode state.
 	// since the returns value(s) of this message is sent to the caller (client), it has to serialized anyway.
@@ -70,7 +74,7 @@ func (rs *Contract) QueryListEnclaveCredentials(ctx contractapi.TransactionConte
 	return allCredentials, nil
 }
 
-func (rs *Contract) QueryEnclaveCredentials(ctx contractapi.TransactionContextInterface, chaincodeId, enclaveId string) (string, error) {
+func (rs *Contract) QueryEnclaveCredentials(ctx contractapi.TransactionContextInterface, chaincodeId string, enclaveId string) (string, error) {
 	key, err := ctx.GetStub().CreateCompositeKey("namespaces/credentials", []string{chaincodeId, enclaveId})
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


g2flyer (commented):
Did you on purpose put it on top of my PR? Makes it a bit harder to review. 

Another more import thing security-wise: besides preventing double registration in ERCC (which can give funcitonal problems), shouldn't we also prevent it in `ecc_enclave`?  Otherwise you can do a trivial DoS by just re-issuing an `init_enclave` and wipe out the old state.  Essentially, `ecall_init` in `common.cpp` should be calleable (without error) only once? ( `interfaces.md` actually had a corresponding FSM, but obviously with the short cut there would be now only two rather the three states which that doc mentioned)

bvavala:
> Did you on purpose put it on top of my PR? Makes it a bit harder to review.

No, just found easier to test. It's one small commit.

> Another more import thing security-wise: besides preventing double registration in ERCC (which can give funcitonal problems), shouldn't we also prevent it in `ecc_enclave`? Otherwise you can do a trivial DoS by just re-issuing an `init_enclave` and wipe out the old state. Essentially, `ecall_init` in `common.cpp` should be calleable (without error) only once? ( `interfaces.md` actually had a corresponding FSM, but obviously with the short cut there would be now only two rather the three states which that doc mentioned)

True, but that's related to ecc_enclave -- so addressable in a separate PR.

Also, the problem and the solution are debatable.

First, using a single enclave (i.e., no replication), it is necessary to assume a benign threat model for the hosting org, in order to prevent state loss. This even applies when a trusted ledger is available.

Second, the FSM was originally designed in the presence of a trusted ledger, and it made sense. However, without that, the FSM is not persistent and the proposer cannot be authenticated. So, the trustworthiness of the org still determines the effectiveness of the FSM inside the enclave.

As a consequence, having any such mechanism _inside_ the enclave does seem to add anything.

However, some state-loss-prevention mechanism could still be meaningful _outside_ the enclave.
The catch is that `initenclave` is stateless, so the FSM is not implementable.
Yet, the org admin can be authenticated.
This authentication seems sufficient, given the first point above,

g2flyer:
> Also, the problem and the solution are debatable.
> 
> First, using a single enclave (i.e., no replication), it is necessary to assume a benign threat model for the hosting org, in order to prevent state loss. This even applies when a trusted ledger is available.
> 
> Second, the FSM was originally designed in the presence of a trusted ledger, and it made sense. However, without that, the FSM is not persistent and the proposer cannot be authenticated. So, the trustworthiness of the org still determines the effectiveness of the FSM inside the enclave.
> 
> As a consequence, having any such mechanism _inside_ the enclave does seem to add anything.
> 
> However, some state-loss-prevention mechanism could still be meaningful _outside_ the enclave.
> The catch is that `initenclave` is stateless, so the FSM is not implementable.
> Yet, the org admin can be authenticated.
> This authentication seems sufficient, given the first point above,

Good point that without TL the local peer is anyway in control and could kill and recreate enclave. And initEnclave is a query, so couldn't have chaincode state.  As soon as we add sealing, of course it would become stateful, though! But as the threat is really some random client, just authorizing it only if it is an admin of the local org as you outline seems probably the way to go (and both corresponding info should be available to chaincode, so should hopefully not complicated to implement ....)

Issue #486, opened 2020-12-08
Client-request/response processing


g2flyer:
**Description**
Encryption/decryption, encoding/decoding and validation of client requests and responses

- [ ] client-sdk
  - [x] protobuf
  - [ ] crypto
- [ ]  peer cli
  - [ ] utility to use client-sdk to do actual work and corresponding inclusion in wrapper script
- [ ] ecc[_enclave]
  - [ ] protobuf
  - [ ] crypto

**Link to feature or bug**
<--!
  Please add link to any applicable feature or bug.
-->

**Link to dependent issues**
<--! 
  Please link to any other work-item this issue depends on.
-->

g2flyer:
ass discussed in #516, move this to FPC 1.1

PR #492, opened 2020-12-09
Peer cli assist for client request/response processing

### LINKED ISSUE #288: Client side support: peer cli
See also sibling features #289 and #290

g2flyer:

Extension of peer cli assist to handle client request/response processing (e.g., encryption/decryption and related protobuf and base64 encoding/decoding).  Modified peer cli wrapper which takes this cli and implemnts the new invoke flows. Additionally some encoding fixes and modified integration tests to work with new responses.

BTW:  we currently somewhat superfluously base64 encode the result in `ecc_enclave/enclave/enclave.cpp:238` before putting it in the protobuf. But seems easiest to drop that when we add support for fabric response object ...

This should close #288 and address the peer cli side of #486 

Note integration test will fail due to missing signature in endorsement.  However, if you comment out the endorsement response test on line 671 in `fabric/bin/peer.sh`, tests which do not rely on state-update will succeed, e.g., all of `integration/echo_test.sh` should succeed.  Note given that we do not have any proper read-write setup, even successful transactions (e.g., by commenting out lines 195-197 in `ecc/chaincode/ecc.go`) will not make the tests succeed (but will allow you to test that the endorsement-verification-response-validation logic in peer.sh is correct ..)

### DIFF
--- a/client_sdk/go/fpc/crypto/crypto.go
+++ b/client_sdk/go/fpc/crypto/crypto.go
@@ -70,14 +70,31 @@ type EncryptionContextImpl struct {
 	chaincodeEncryptionKey []byte
 }
 
-func (e *EncryptionContextImpl) Reveal(responseBytes []byte) ([]byte, error) {
-	response := &protos.ChaincodeResponseMessage{}
-	err := proto.Unmarshal(responseBytes, response)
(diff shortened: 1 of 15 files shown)


bvavala (approved):
lg, builds/works as expected

PR #494, opened 2020-12-09
Convert all log to flogger

g2flyer:

Converted all uses of "log" to "flogging" so we have consistent logging across the board and also logging control via `FABRIC_LOGGIN_SPEC` ...

Was straight-forward conversion. Did also usually convert `log.Print` to `logger.Debug`, which maybe in some cases could have been also handled as Info.  Note as `management.go` and `contract.go` share the name space (and hence can only have one logger variable) i solomonically solved it by putting `logger` var into a separate file `logging.go` :-)

### DIFF
--- a/client_sdk/go/fpc/attestation/transformation.go
+++ b/client_sdk/go/fpc/attestation/transformation.go
@@ -9,18 +9,20 @@ package attestation
 
 import (
 	"fmt"
-	"log"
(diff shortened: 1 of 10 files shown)


bvavala (approved):
lgtm

mbrandenburger (approved):
LGTM!

PR #495, opened 2020-12-12
Make test-net work in HW mode

g2flyer:

PR #483 successfully worked with test-net tests in SIM mode, but not in HW mode.  This PR fixes this.

More specifically, it
- fixes the docker, docker-compose and test-net scripts to work properly with hw mode
- it ensurse that only HW attestations are accepted in HW mode
- it changes the (management side) of the client sdk to get attestation parameters with the same logic as the peer cli (ie.., based on `SGX_MODE` and `SGX_CREDENTIALS_PATH` env variables)

Note that the instructions for the test net have slightly changed!


common/crypto/attestation-api/evidence/verify-evidence.cpp
```diff
@@ -237,11 +237,13 @@ bool verify_evidence(uint8_t* evidence,
     COND2LOGERR(s == NULL, "no attestation type");
     COND2LOGERR(evidence_field == NULL, "no evidence field");
 
+#ifdef SGX_SIM_MODE
```
bvavala, line 240: There must be a security reason for this, but I cannot see it.
The `attestation-api` aims to be a generic interface, which can support several types of attestations.
A `simulated` attestation is just one of the types, which requires no hardware support.
The interface supports it, just as it supports SGX EPID, and additional ones later.

The conditional compile disables support for the `simulated` attestation when compiled with HW-mode in SGX.
However, first, the technology is not meant to be mutually exclusive with SGX in HW mode. In fact, it's independent from it.
Second, both can coexist because the `evidence` data blob differentiates them.

Back to FPC, the evidence data blob enables informed risk taking. Such data provides participants on the channel with all the info they need to make a trust decision. Participants that require high level of privacy might want to check that the evidence is related to an EPID/DCAP attestation, as well as additional reported details; while chaincode developers (on the same channel) might be initially comfortable with a simulated attestation.

Concluding, with the simulated attestation disabled, I see limitations and no benefits. The additional checks that participants may need to do, can be either offloaded to the participants themselves, or specified through a suitable security policy which the Registry enforces. How such a policy would look like is an open question.

bvavala (commented):
Looks good, though I am unable to test it now.

Regarding the attestation part, I think the approach should be different.

PR #498, opened 2020-12-15
ERCC & sample-net fix/update 

g2flyer:
**What this PR does / why we need it**:

Fixes bug in ercc `queryListProvisionedEnclaves` and adds new err function `queryChaincodeEndPoints` (primarily for peer cli assist but can also be convenient for manual debugging).  Also fixed small bug in sample-net and added section with diagnostics commands, in particular, with copy/paste ready commands to query ercc.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

You can test, e.g., as follows:
```
cd $FPC_PATH/integration/test-network/
# Follow README to setup everything, including running test

# exercise ercc commans
cd $FPC_PATH/integration/test-network/fabric-samples/test-network; \
source ./scripts/envVar.sh; \
export PATH=$(readlink -f ../bin):$PATH
export FABRIC_CFG_PATH=$FPC_PATH/integration/test-network/fabric-samples/config
setGlobals 1;
peer chaincode query -C mychannel -n ercc -c '{"Function": "queryChaincodeEndPoints", "Args" : ["echo"]}'
peer chaincode query -C mychannel -n ercc -c '{"Function": "queryListProvisionedEnclaves", "Args" : ["echo"]}'
```

**Does this PR introduce a user-facing changes and/or breaks backward compatability?**:

PR #500, opened 2020-12-16
Fix initEnclave in integration tests

g2flyer:

ecc_mock - ecc merge (#483) left various incompatibilities between integration test (e.g., peer wrapper) and sample network.  As part of resolving this, i also simplified attestation parameter handling, added some more debug output and made `FPC_PATH` the sole shell variable we use, also in script (instead of `FPC_TOP_DIR`)

Should work for all three scenarios
- test net (failing as expected due to missing signature)
- test net with mock_ecc (succeeding completely)
- integration test (failing in invokes, i.e., after `initEnclave`. Complete integration test has to wait for merge of #498 and completiong of #492 ...)


bvavala (approved):
nice simplification.
echo-test tested successfully

PR #501, opened 2020-12-18
support for digital signatures + integration tests work (again)

bvavala:
This PR add:
* new protobufs for better handling digital signatures and rwset
* implements digital signature computation and verification
* implements rwset serialization and replay
* makes integration tests work again

Notes:
* lots of code, mainly in the enclave, needs refactoring and improved error handling
* fabric (state) calls in the enclave are void, but should be modified to return any failure
* Fabric's kv store consistency level has to be investigated, to ensure that FPC chaincodes match expected behavior of regular chaincodes (e.g., do reads always return the latest committed value first? or the most recent local write first? are writes of the same key collapsed into a single one, or result into multiple versions?)

### DIFF
--- a/common/error.h
+++ b/common/error.h
@@ -6,10 +6,6 @@
 
 #pragma once
 
-#ifndef LOG_DEBUG
-#define LOG_DEBUG
-#endif
-
 #define COND2ERR(b)                                            \
     do                                                         \
     {                                                          \
--- a/common/error.h
+++ b/common/error.h
@@ -6,6 +6,10 @@
 
 #pragma once
 
+#ifndef LOG_DEBUG
+#define LOG_DEBUG
+#endif
+
 #define COND2ERR(b)                                            \
     do                                                         \
     {                                                          \
--- a/ecc/chaincode/ecc.go
+++ b/ecc/chaincode/ecc.go
@@ -152,7 +152,7 @@ func (t *EnclaveChaincode) endorse(stub shim.ChaincodeStubInterface) pb.Response
 		return shim.Error(errMsg)
 	}
 
-	responseMsg, err := extractChaincodeResponseMessage(stub)
+	signedResponseMsg, responseMsg, err := extractChaincodeResponseMessages(stub)
 	if err != nil {
 		errMsg := fmt.Sprintf("cannot extract chaincode response message: %s", err.Error())
 		logger.Errorf(errMsg)
@@ -184,18 +184,21 @@ func (t *EnclaveChaincode) endorse(stub shim.ChaincodeStubInterface) pb.Response
 
 	// check cc param.MSPID matches MSPID of endorser (Post-MVP)
 
-	// replay read/writes from kvrwset from enclave (to prepare commitment to ledger) and extract kvrwset for subsequent validation
-	readset, writeset, err := utils.ReplayReadWrites(stub, responseMsg.RwSet)
-	if err != nil {
-		return shim.Error(err.Error())
-	}
-
 	// validate enclave endorsement signature
-	err = utils.Validate(responseMsg, readset, writeset, &attestedData)
+	logger.Debugf("Validating endorsement")
+	err = utils.Validate(signedResponseMsg, &attestedData)
 	if err != nil {
 		return shim.Error(err.Error())
 	}
 
+	// replay read/writes from kvrwset from enclave (to prepare commitment to ledger) and extract kvrwset for subsequent validation
+	logger.Debugf("Replaying rwset")
+	err = utils.ReplayReadWrites(stub, responseMsg.FpcRwSet)
+	if err != nil {
+		return shim.Error(err.Error())
+	}
+
+	logger.Debugf("Endorsement successful")
 	return shim.Success([]byte("OK")) // make sure we have a non-empty return on success so we can distinguish success from failure in cli ...
 }
 
--- a/ecc/chaincode/ecc.go
+++ b/ecc/chaincode/ecc.go
@@ -115,27 +115,27 @@ func (t *EnclaveChaincode) invoke(stub shim.ChaincodeStubInterface) pb.Response 
 		return shim.Error(errMsg)
 	}
 
-	chaincodeResponseMessage, errInvoke := t.enclave.ChaincodeInvoke(stub, chaincodeRequestMessage)
+	signedChaincodeResponseMessage, errInvoke := t.enclave.ChaincodeInvoke(stub, chaincodeRequestMessage)
 	if errInvoke != nil {
 		errMsg = fmt.Sprintf("t.enclave.Invoke failed: %s", errInvoke)
 		logger.Errorf(errMsg)
 		// likely a chaincode error, so we still want response go back ...
 	}
 
-	chaincodeResponseMessageB64 := []byte(base64.StdEncoding.EncodeToString(chaincodeResponseMessage))
-	logger.Debugf("base64-encoded response message: '%s'", chaincodeResponseMessageB64)
+	signedChaincodeResponseMessageB64 := []byte(base64.StdEncoding.EncodeToString(signedChaincodeResponseMessage))
+	logger.Debugf("base64-encoded response message: '%s'", signedChaincodeResponseMessageB64)
 
 	var response pb.Response
 	if errInvoke == nil {
 		response = pb.Response{
 			Status:  shim.OK,
-			Payload: chaincodeResponseMessageB64,
+			Payload: signedChaincodeResponseMessageB64,
 			Message: errMsg,
 		}
 	} else {
 		response = pb.Response{
 			Status:  shim.ERROR,
-			Payload: chaincodeResponseMessageB64,
+			Payload: signedChaincodeResponseMessageB64,
 			Message: errMsg,
 		}
 	}
--- a/ecc/chaincode/enclave/enclave.go
+++ b/ecc/chaincode/enclave/enclave.go
@@ -41,6 +41,7 @@ func NewEnclaveStub() StubInterface {
 }
 
 func (e *EnclaveStub) Init(chaincodeParams, hostParams, attestationParams []byte) ([]byte, error) {
+	// Estimate of the buffer length that is necessary for the credentials. It should be conservative.
 	const credentialsBufferMaxLen = 16 * 1024
 
 	if e.isInitialized {
@@ -106,6 +107,7 @@ func (e *EnclaveStub) GetEnclaveId() (string, error) {
 
 // ChaincodeInvoke calls the enclave for transaction processing
 func (e *EnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, crmProtoBytes []byte) ([]byte, error) {
+	// Estimate of the buffer length where the enclave will write the response.
 	const scresmProtoBytesMaxLen = 1024 * 100 // Let's be really conservative ...
 
 	if !e.isInitialized {
--- a/ecc/chaincode/enclave/enclave.go
+++ b/ecc/chaincode/enclave/enclave.go
@@ -16,7 +16,6 @@ import (
 	"unsafe"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/hyperledger-labs/fabric-private-chaincode/internal/protos"
 	"github.com/hyperledger/fabric-chaincode-go/shim"
 	"golang.org/x/sync/semaphore"
 )
@@ -29,7 +28,6 @@ import (
 import "C"
 
 const enclaveLibFile = "enclave/lib/enclave.signed.so"
-const maxResponseSize = 1024 * 100 // Let's be really conservative ...
 
 type EnclaveStub struct {
 	eid           C.enclave_id_t
@@ -43,6 +41,8 @@ func NewEnclaveStub() StubInterface {
 }
 
 func (e *EnclaveStub) Init(chaincodeParams, hostParams, attestationParams []byte) ([]byte, error) {
+	const credentialsBufferMaxLen = 16 * 1024
+
 	if e.isInitialized {
 		return nil, fmt.Errorf("enclave already initialized")
 	}
@@ -50,7 +50,7 @@ func (e *EnclaveStub) Init(chaincodeParams, hostParams, attestationParams []byte
 	var eid C.enclave_id_t
 
 	// prepare output buffer for credentials
-	credentialsBuffer := C.malloc(maxResponseSize)
+	credentialsBuffer := C.malloc(credentialsBufferMaxLen)
 	defer C.free(credentialsBuffer)
 	credentialsSize := C.uint32_t(0)
 
@@ -70,7 +70,7 @@ func (e *EnclaveStub) Init(chaincodeParams, hostParams, attestationParams []byte
 		(*C.uint8_t)(C.CBytes(hostParams)),
 		C.uint32_t(len(hostParams)),
 		(*C.uint8_t)(credentialsBuffer),
-		C.uint32_t(maxResponseSize),
+		C.uint32_t(credentialsBufferMaxLen),
 		&credentialsSize)
 
 	if ret != 0 {
@@ -106,6 +106,8 @@ func (e *EnclaveStub) GetEnclaveId() (string, error) {
 
 // ChaincodeInvoke calls the enclave for transaction processing
 func (e *EnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, crmProtoBytes []byte) ([]byte, error) {
+	const scresmProtoBytesMaxLen = 1024 * 100 // Let's be really conservative ...
+
 	if !e.isInitialized {
 		return nil, fmt.Errorf("enclave not yet initialized")
 	}
@@ -128,9 +130,9 @@ func (e *EnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, crmProto
 	defer C.free(unsafe.Pointer(signedProposalPtr))
 
 	// prep response
-	cresmProtoBytesLenOut := C.uint32_t(0) // We pass maximal length separately; set to zero so we can detect valid responses
-	cresmProtoBytesPtr := C.malloc(maxResponseSize)
-	defer C.free(cresmProtoBytesPtr)
+	scresmProtoBytesLenOut := C.uint32_t(0) // We pass maximal length separately; set to zero so we can detect valid responses
+	scresmProtoBytesPtr := C.malloc(scresmProtoBytesMaxLen)
+	defer C.free(scresmProtoBytesPtr)
 
 	crmProtoBytesPtr := C.CBytes(crmProtoBytes)
 	defer C.free(unsafe.Pointer(crmProtoBytesPtr))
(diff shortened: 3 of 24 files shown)


internal/utils/validation.go
```diff
@@ -8,95 +8,129 @@ SPDX-License-Identifier: Apache-2.0
...
+			k := TransformToFPCKey(rwset.Reads[i].Key)
+			v, err := stub.GetState(k)
+			if err != nil {
+				return fmt.Errorf("error (%s) reading key %s", err, k)
+			}
+
+			logger.Debugf("read key %s value(hex) %s", k, hex.EncodeToString(v))
+
+			// compute value hash
+			h := sha256.New()
```
g2flyer, line 64: For consistency, shouldn't we use here also the pdo hash function (as used in generation?) You already export it through cgo?

internal/utils/validation.go
```diff
@@ -8,95 +8,129 @@ SPDX-License-Identifier: Apache-2.0
...
+	signaturePtr := C.CBytes(signedResponseMessage.Signature)
+	defer C.free(signaturePtr)
 
-	valid := ecdsa.VerifyASN1(pub.(*ecdsa.PublicKey), hash[:], responseMsg.Signature)
-	if !valid {
-		return fmt.Errorf("signature invalid")
+	ret := C.verify_signature((*C.uint8_t)(enclaveVkPtr), C.uint32_t(len(attestedData.EnclaveVk)), (*C.uint8_t)(responseMessagePtr), C.uint32_t(len(signedResponseMessage.ChaincodeResponseMessage)), (*C.uint8_t)(signaturePtr), C.uint32_t(len(signedResponseMessage.Signature)))
+	if ret == false {
+		return fmt.Errorf("enclave signature verification failed")
 	}
```
g2flyer, line 141: Hmm, would be nicer to abstract into a separate go function, e.g., in `internal/utils/crypto.go` (where now also the `ComputeHash` function is obsolete).  You will need this function also in mock and it could centralize all corresponding cgo stuff (and eventually also encryption) in a single and shareble file?

fabric/bin/peer.sh
```diff
@@ -666,7 +666,7 @@ handle_chaincode_call() {
 	if [ ! -z ${DO_WAIT_TIMEOUT+x} ]; then
 	    opt_arg="--waitForEvent ${opt_arg}"
 	fi
-	try_r $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke "${OTHER_ARGS[@]}" ${opt_arg} -C ${CHAN_ID} -n ${CC_ID} -c '{"Args":["__endorse", "'${encrypted_response}'"]}'
+	try_r $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke "${OTHER_ARGS[@]}" ${opt_arg} -C ${CHAN_ID} -n ${CC_ID} -c '{"Args":["__endorse", "'${encrypted_response}'"]}' --waitForEvent
```
g2flyer, line 669: `opt_arg` (should) handle that if you passed it when invoking the cli, and if not, then we should also not do it here ...
```suggestion
	try_r $RUN ${FABRIC_BIN_DIR}/peer chaincode invoke "${OTHER_ARGS[@]}" ${opt_arg} -C ${CHAN_ID} -n ${CC_ID} -c '{"Args":["__endorse", "'${encrypted_response}'"]}'
```

ecc_enclave/enclave/enclave.cpp
```diff
@@ -229,11 +74,16 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
...
 
-    ret = invoke(response, response_len, &response_len_out, &ctx);
-    COND2ERR(ret != 0);
+    invoke_ret = invoke(response, response_len, &response_len_out, &ctx);
+    // invoke_ret is not checked
+
+    // TODO double check or rethink if it is appropriate for a chaincode
+    // to return an error and still forward the response
+    // in particular: should the enclave sign a response? and the rwset? could the tx be committed
+    // though it failed?
```
g2flyer, line 86: Invoke_ret should be handled when we add the fabric response object (in status field) instead of just using the response byte array, At that time we might also add additional out params to invoke so the chaincode can (as in standard fabric) also pass an error message back ...

ecc_enclave/enclave/enclave.cpp
```diff
@@ -248,31 +98,105 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
...
+            crm.encrypted_response = (pb_bytes_array_t*)pb_realloc(
+                crm.encrypted_response, PB_BYTES_ARRAY_T_ALLOCSIZE(b64_response.length()));
+            COND2LOGERR(crm.encrypted_response == NULL, "cannot allocate encrypted message");
+            crm.encrypted_response->size = b64_response.length();
+            ret = memcpy_s(crm.encrypted_response->bytes, crm.encrypted_response->size,
+                b64_response.c_str(), b64_response.length());
+            COND2LOGERR(ret != 0, "cannot encode field");
+        }
+
+        {  // serialize enclave id
```
g2flyer, line 112: Isn't the serialize misleading here and for rwset below?  What you create here is first the (non-serialized/in-memory) crm protobuf object and the serialize is really only what you call the encode step?  the proposal is somewhat a special case as you re-use the already serialized version (so the comment "serialize" is also misleading, just for different reasons?)

seems for signature the comments are much clearer (with fill). Probably best to use same terminology (and sequence, i.e., first object creation, then encoding) here as below (with additional explicitly pointing out that for proposal you reuse already serialized version and maybe just moving it right in front the encoding part?)

ecc_enclave/enclave/shim.cpp
```diff
@@ -64,6 +68,8 @@ void get_state(
     COND2LOGERR(encoded_cipher_len != encoded_cipher_s.size(), "Unexpected string length");
 
     // base64 decode
+    // TODO: double check if Fabric handles values as strings; if it handles binary values, remove
```
g2flyer, line 71: fabric has only byte version of getstate/putstate in [shim](https://pkg.go.dev/github.com/hyperledger/fabric-chaincode-go/shim). We might have though some debug statement which might rely on it being a string

btw: i notice now that  also documents a `DelState` function.  Can't recall have seen that in the past but matches the `deleted` flag from the protobuf.

ecc_enclave/enclave/shim.cpp
```diff
@@ -103,17 +109,49 @@ void get_public_state(
...
+    ByteArray value_hash = pdo::crypto::ComputeMessageHash(ByteArray(val, val + *val_len));
+    auto it = ctx->read_set.find(key_s);
+    if (it == ctx->read_set.end())
+    {
+        // key not found, insert it
+        ctx->read_set.insert({std::string(key), value_hash});
+    }
+    else
+    {
+        // key found, ensure value is the same, or fail
```
g2flyer, line 124: We should mention here a comment on our assumptions/experiments on (surprising) fabric behavior and maybe a TODO to double-check with fabric maintainers ...  (i.e., along the lines of the text you have in the PR itself)

Note that auction_test actually fails for two tests which might be related to this (or below write assumption)?

g2flyer (changes requested):
Looks overall very good.  Build requires a simple patch (see below) but then works. However, for me, in both SIM and HW mode, the auction_test fails for two cases (echo and deployment test succeed). Maybe related to the state consistency assumptions? (Looking at auction_test it could also have been due to attempting concurency which actually would be a problem for DRAW detection when reordering happens. But you forced a --waitForEvent inside peer.cli, so this should not be the issue?0

A number of inline comments. As mentioned there (but maybe a bit hidden). `mock_enclave` also has to be changed with the new message formats and pdo signature
```
From b0c16f4370607a027edc433b24e3b5e056f3c06d Mon Sep 17 00:00:00 2001
From: michael steiner <michael.steiner@intel.com>
Date: Fri, 18 Dec 2020 11:28:22 -0800
Subject: [PATCH] build fix

Signed-off-by: michael steiner <michael.steiner@intel.com>
---
 Makefile | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/Makefile b/Makefile
index 43753d5..50fd55d 100644
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,8 @@ include $(TOP)/build.mk

 SUB_DIRS = protos common internal ercc ecc_enclave ecc fabric client_sdk examples utils integration # docs

-FPC_SDK_DEP_DIRS = protos utils/fabric common ecc_enclave ecc
-FPC_PEER_DEP_DIRS = protos common ercc fabric ecc_enclave ecc
+FPC_SDK_DEP_DIRS = protos common ecc_enclave ecc utils/fabric
+FPC_PEER_DEP_DIRS = protos common ercc ecc_enclave ecc fabric
 # FPC_PEER_DEP_DIRS has to include protos, ecc, ecc_enclave, common and ercc only if we run chaincode in external builder directly on host and not indirectly via docker
 FPC_PEER_CLI_WRAPPER_DEP_DIRS = utils/fabric

--
2.29.2
```

common/error.h
```diff
@@ -6,6 +6,10 @@
 
 #pragma once
 
+#ifndef LOG_DEBUG
+#define LOG_DEBUG
+#endif
+
```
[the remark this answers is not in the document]
bvavala: right, fixed

g2flyer:
Ah, notice that you also just found the build issue (note my commit tries to make the order of directories overall consistent across the different SUB_DIR definitions as these were also slightly different) and looking at the code i also realized the issue with auction test was composite keys ...

ecc/chaincode/ecc.go
```diff
@@ -152,7 +152,7 @@ func (t *EnclaveChaincode) endorse(stub shim.ChaincodeStubInterface) pb.Response
 		return shim.Error(errMsg)
 	}
 
-	responseMsg, err := extractChaincodeResponseMessage(stub)
```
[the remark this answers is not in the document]
bvavala: I guess you mean in `invoke`. makes sense

protos/fpc/fpc.proto
```diff
@@ -108,20 +108,32 @@ message CleartextChaincodeResponse {
...
+// FPCKVSet augments the Fabric kvrwset.KVRWSet protobuf to include the hash of the value of each read.
+// Specifically, read_value_hashes[i] is the hash of the value associated to rw_set.reads[i].key
+message FPCKVSet {  
+    kvrwset.KVRWSet rw_set = 1;
+    repeated bytes read_value_hashes = 2;
+}
+
 message ChaincodeResponseMessage {
     // an aes128-gcm encryption of the serialization of CleartextChaincodeResponse
     bytes encrypted_response = 1;
```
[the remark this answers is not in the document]
bvavala: let's begin from a working implementation, I have no strong opinion on that

internal/utils/validation.go
bvavala: yes, that export was actually for testing.
I can address this in a future pr. For now it's safe, since with a different hash function would fail

internal/utils/validation.go
bvavala: There is a lot of refactoring to be done. I have opened 501 to track the tasks, and maybe proceed in parallel

bvavala:
> Ah, notice that you also just found the build issue (note my commit tries to make the order of directories overall consistent across the different SUB_DIR definitions as these were also slightly different) and looking at the code i also realized the issue with auction test was composite keys ...

We really should support the fabric shim functions for the composite keys. FPC is now redefining the machinery and translating it for Fabric.

ecc_enclave/enclave/enclave.cpp
bvavala: seems appropriate, we have the todo there.

ecc_enclave/enclave/shim.cpp
bvavala: but why would we convert these to string? I had noticed that, and it's unnecessary.

ecc_enclave/enclave/shim.cpp
bvavala: > We should mention here a comment on our assumptions/experiments on (surprising) fabric behavior and maybe a TODO to double-check with fabric maintainers ... (i.e., along the lines of the text you have in the PR itself)

added a comment

> Note that auction_test actually fails for two tests which might be related to this (or below write assumption)?

The problem was the handling of the composite keys in the FPC go shim :(

ecc/chaincode/enclave/enclave.go
```diff
@@ -43,14 +41,16 @@ func NewEnclaveStub() StubInterface {
 }
 
 func (e *EnclaveStub) Init(chaincodeParams, hostParams, attestationParams []byte) ([]byte, error) {
+	const credentialsBufferMaxLen = 16 * 1024
```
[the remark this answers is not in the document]
bvavala: added comment

ecc/chaincode/enclave/enclave.go
```diff
@@ -106,6 +106,8 @@ func (e *EnclaveStub) GetEnclaveId() (string, error) {
 
 // ChaincodeInvoke calls the enclave for transaction processing
 func (e *EnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, crmProtoBytes []byte) ([]byte, error) {
+	const scresmProtoBytesMaxLen = 1024 * 100 // Let's be really conservative ...
```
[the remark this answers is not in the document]
bvavala: added comment

ecc/chaincode/ecc.go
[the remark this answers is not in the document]
bvavala: done

internal/utils/validation.go
bvavala: added comment

fabric/bin/peer.sh
bvavala: done -- hopefully the auction still works

ecc_enclave/enclave/shim.cpp
bvavala: added comment

ecc/chaincode/enclave/enclave.go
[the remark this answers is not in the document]
g2flyer: hmm, the comment doesn't really mention on _how_ you got to that number. That it was an estimate i already had figured out before :-)

ecc/chaincode/ecc.go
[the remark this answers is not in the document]
g2flyer: yep, that's what i meant ...

fabric/bin/peer.sh
g2flyer: as mentioned in the main comment.  I think the logic in `auction_test.sh` of omitting --waitForEvent is actually flawed, but that should be fixed there not in the peer cli

g2flyer:
> > Ah, notice that you also just found the build issue (note my commit tries to make the order of directories overall consistent across the different SUB_DIR definitions as these were also slightly different) and looking at the code i also realized the issue with auction test was composite keys ...
> 
> We really should support the fabric shim functions for the composite keys. FPC is now redefining the machinery and translating it for Fabric.

Well, shim.h does define composite keys somewhat consistent with go shim.  Just that the implementaion merges it in a singe ecall for get/put.  If you look at the full FPC design and the related ECC<->TLCC channel with the corresponding verification function, i already split that back.

That said, we anyway don't support them for FPC lite right now.  We could, of course, by demultiplexing them onto a single key, similar to doing so to support multiple keys.  Not sure whether from a priority perspective it is on the top of the list though?  I.e., given that an application can also handle it reasonably well at the application level (and us implementing in shim would still loose the data-parallel aspect of fabric),  i would wait until there is push-back.  Seems to me much less fundamental than support of seal (otherwise you simply won't survive a peer restart) or multi-enclave (robustness but also better security/fairness), neither of which an application writer can do anything at the app level

g2flyer (approved):
builds and runs now smoothly.  I'm fine also with doing mock_enclave in a separate pr, in particular as (a) already in the current state we get again a working integration test and (b) doing this change for mock_enclave also bring in all the stuff mentioned in #502, so better done separately.

bvavala:
> Well, shim.h does define composite keys somewhat consistent with go shim. Just that the implementaion merges it in a singe ecall for get/put. If you look at the full FPC design and the related ECC<->TLCC channel with the corresponding verification function, i already split that back.

That's correct, but just part of the story.
If you also look at `shim.go`, particularly [here ](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/50a54ea74aedb89c691672de8322e71f1b6bde7c/ecc/chaincode/enclave/shim.go#L142) and [here](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/50a54ea74aedb89c691672de8322e71f1b6bde7c/ecc/chaincode/enclave/shim.go#L169), you'll see that the code embeds the recognition of composite keys (`IsFPCCompositeKey`) and it performs a translation by splitting them accordingly (`SplitFPCCompositeKey`).
Back to the auction example, an FPC composite key is `.somePrefix.MyAuction1.`, which is translated to the Fabric composite key `somePrefixMyAuction1`.

g2flyer:
> That's correct, but just part of the story.
> If you also look at `shim.go`, particularly [here ](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/50a54ea74aedb89c691672de8322e71f1b6bde7c/ecc/chaincode/enclave/shim.go#L142) and [here](https://github.com/hyperledger-labs/fabric-private-chaincode/blob/50a54ea74aedb89c691672de8322e71f1b6bde7c/ecc/chaincode/enclave/shim.go#L169), you'll see that the code embeds the recognition of composite keys (`IsFPCCompositeKey`) and it performs a translation by splitting them accordingly (`SplitFPCCompositeKey`).
> Back to the auction example, an FPC composite key is `.somePrefix.MyAuction1.`, which is translated to the Fabric composite key `somePrefixMyAuction1`.

Not clear to me why why we added the '.' separator, it should also work without with the pre-fix match? but also not clear whether it has any impact on the developer? Looking at [fabric go stub](https://github.com/hyperledger/fabric-chaincode-go/blob/master/shim/stub.go) it seems that they (a) keep composite keys in a completely different namespace of non-composite-keys and (b) use MinUnicodeRuneValue as separator.  Now the way we use the stub, we do inherit the namespace separation, i.e., (a) and as our separator is valid only during the transfer between enclave and go, it shouldn't really matter? 

The main problem though would only be if the developer would use a non-composite key which uses the dot, thinking it would be separate?

https://github.com/hyperledger/fabric-chaincode-go/blob/master/shim/stub.go

bvavala:
> Not clear to me why why we added the '.' separator, it should also work without with the pre-fix match?

I speculate that originally the separator was something like `\00`, so this was translated to a `.`.
Now, I agree, it should not be an issue.

> The main problem though would only be if the developer would use a non-composite key which uses the dot, thinking it would be separate?

Yes, more generally: a developer must know about that (i) to use it, (ii) to avoid mistakes when choosing key strings, (iii) to avoid using Fabric's composite keys.

g2flyer:
> > The main problem though would only be if the developer would use a non-composite key which uses the dot, thinking it would be separate?

Actually, this paragraph (and subsequent link) were left overs which i forgot to clean up and the paragraph above actually argues, based on the fabric go implementation  that ...

> Yes, more generally: a developer must know about that (i) to use it, (ii) to avoid mistakes when choosing key strings, (iii) to avoid using Fabric's composite keys.

... this is actually not a concern

PR #505, opened 2020-12-19
Make client go sdk work with new response message

g2flyer:

Fixes encoding bug with new messages in client go sdk

Fixing bug was less effort than creating an issue :-)


bvavala (approved):
tested

Issue #370, opened 2020-06-10
Keep Uptodate with latest Fabric Releases


g2flyer:
**Is your feature request related to a problem? Please describe.**

Standing issue to follow and keep in pace with fabric releases

**Describe the solution you'd like**
n/a

**Describe alternatives you've considered**
n/a

**Additional context**

g2flyer:
This is handled only for now, i was thinking of keeping this as a standing issue.  E.g., we might also have to update to 2.2! So re-opened it ...

mbrandenburger:
> This is handled only for now, i was thinking of keeping this as a standing issue. E.g., we might also have to update to 2.2! So re-opened it ...

That's ok - but can be moved to MVP now.

g2flyer:
> That's ok - but can be moved to MVP now.

I was thinking about it but left it for the unlikely case that we would be somewhat associated with 2.2 (in which case we would better upgrade).  But i guess we can move it back here in the unlikely case that this becomes relevant ...

PR #508, opened 2020-12-23
Bump FABRIC_VERSION to 2.3.0

g2flyer:

To be in sync with latest Fabric version  from v2.2.0 to v2.3.0 and also bumped also some other related dependenceis:
- update of `go.mod`
  - fabric-sdk-go => v1.0.0-rc1
  - fabric-contract-api-go => v1.1.1
  - fabric-chaincode-go => v0.0.0-20201119163726-f8ef75b17719
- bumped sample-config (due to couchdb dependencies) to current `master`.
         (unfortunately, there is no tagged version i could use; that said, we also previously used just some "random" commit from master ...)

Notes:
- What's new [v2.2.1]() & [v2.3.0](https://hyperledger-fabric.readthedocs.io/en/release-2.3/whatsnew.html#what-s-new-in-hyperledger-fabric-v2-3)
- Release notes [v2.2.1](https://github.com/hyperledger/fabric/releases/tag/v2.2.1) & [v2.3.0](https://github.com/hyperledger/fabric/releases/tag/v2.3.0):
- some noteworthy changes:
  - gives now option for Orderer directly on application channel rather than system channel
  - adds ledger snapshots (interesting for FPC from a long-term perspective)
  - escc & vscc 'enable' removed from `sampleconfig/core.yaml` The plugins are still defined but i guess this is a first step to purge them?
  - Go got bumped to 1.14.12 (but is still 1.15! BTW: after replacing ecdsa sigs with pdo crypto, we actually wouldn't really need the hard 1.15 dependency? Not that we should proactively move back, but probably good to keep in mind ...)
 - CouchDB 3.1.1 is now the tested CouchDB version.

Next "iteration" of issue #370 .

to prepare testing after checkout
```
git submodules update --init
(cd ../../hyperledger/fabric; git fetch; git reset --hard v2.3.0)
make clobber
```


bvavala (approved):
lg

PR #509, opened 2020-12-23
Bump to SGX SDK 2.12 and mention also Ubuntu 20.04

g2flyer:

Make sure for release that we support latest SGX version; also enables us to provide support for Ubuntu 20.04 (see notes in `README.md` for some small `config.override.mk` change needed.)

Notes: 
- It still should work on a host having older SGX SDK, e.g., 2.11, just not on Ubuntu 20.04. So not hurry to upgrade your various test machines.
- there is also a newer version of sgxssl ( lin2.11_1.1.1.g vs currently used lin2.10_1.1.1.g). As this version changed (pretty much undocumented short of code) pthread stuff and breaks pdo crypto, i didn't bump that version.

Next iteration of #411 


bvavala (approved):
lg.
Importantly, travis/docker tests with the 2.12 SDK, but locally the 2.10 SDK still works.

Issue #504, opened 2020-12-18
Validation that invoke arguments are correct is missing


g2flyer:
To safe us in parsing fabric proposal in enclave/c, we currently pass also the request-message to the enclave and take arguments from there, not proposal.  However, we do only include proposal into signature which we later verify in enclave without nothing verifying the consistency of the two. So we either - should include also request message into (to be signed) response message and have __endorse verify the proposal and request message in signature are consistent, i.e., the latter is contained in the former, or
- we extract inside enclave the request from proposal (making also passing of proposal unnecessary)

(Related to #460)

bvavala:
Elaborating on the two solutions (extracting request from proposal; include request in response, sign and verify later) above, there are tradeoffs.

The former results in fewer bytes passed between Go and enclave space, and it requires some nanopb/cpp coding inside the enclave to handle the signed proposal. Essentially this is what Fabric does in the [shim](https://github.com/hyperledger/fabric-chaincode-go/blob/master/shim/stub.go) to set it up.

The latter incurs in additional bytes passed to the enclave (twice the request, since this is passed directly and it is also included in the signed proposal). However, it does not require any extraction from the signed proposal, thereby saving nanopb/cpp code for handling Fabric-internal details. Finally, the hash of the request must additionally be included in the response, and checked later.

Issue #478, opened 2020-12-03
PDO Crypto RSA pk-encryption with 3072 bit keys and unbound payload size


g2flyer:
**Is your feature request related to a problem? Please describe.**

The current pdo crypto library we use does only 2048-bit RSA for pk-encryption and also limits the message length to whatever fits into a OAEP-padded RSA encryption (~200bytes).   The former doesn't match the NIST recommendation anymore for 128-bit security and would require (at least) 3072 bits.   As request size is aproprio not bound, it would also be highly desirable to support hybrid-public-key encryption for RSA which removes the limit

**Describe the solution you'd like**
- change key size from 2048 to 3072
- either change existing EncryptMessage or add new EncryptLongMessage or alike which does hybrid encryption. (E.g., RSA-KEM as key-encapsulation algorithm and AES-256-GCM with nonce 0 as data encapsulation mechanisms; RSA-OAEP with random key and AES-256-GCM should also work and might be simpler to implement with openssl)

**Describe alternatives you've considered**
an alternative to match signatures would also be to go to EC-based crypto and DHIES. Alas, openssl _still_ does not support it, although there are some third-party packages doing it on top of openssl: 
https://github.com/chixsh/openssl-pkey-ec-ies

**Additional context**

mbrandenburger:
@bvavala

bvavala:
> @bvavala What does it take to bump RSA key size to 3072? Just changing a [single line in PDO](https://github.com/hyperledger-labs/private-data-objects/blob/master/common/crypto/pkenc.h#L26) - and see if it breaks FPC? [This](https://github.com/hyperledger/fabric-private-chaincode/blob/main/internal/crypto/crypto.go#L28) looks like a good candidate to fail.
> WDYT?

Possibly yes. We should try it locally, open a PR in PDO and update FPC.
I see an opportunity here for you to do your first commit in PDO :)

mbrandenburger:
> I see an opportunity here for you to do your first commit in PDO :)

Ok, will try! :P

mbrandenburger:
Long message asymmetric encryption was introduced with #584.

mbrandenburger:
RSA 3072 bit keys are introduced with #605. Once #605 is merged this issue can be closed.

mbrandenburger:
Completed with #604

PR #516, opened 2020-12-26
Enable message encryption and chaincode request message consistency check

### LINKED ISSUE #412: Design common attestation / crypto API

bvavala:
(self-explanatory)

This PR addresses #504 (consistency check), part of #486 (message encryption), and maintains the limitations described in #478 (message size).

Also, see comment below, closes #412, and #486 can be moved to next release

### DIFF
--- a/client_sdk/go/fpc/crypto/crypto.go
+++ b/client_sdk/go/fpc/crypto/crypto.go
@@ -8,6 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 package crypto
 
 import (
+	"crypto/rand"
 	"encoding/base64"
 
 	"github.com/hyperledger-labs/fabric-private-chaincode/internal/protos"
@@ -29,16 +30,65 @@ import "C"
 
 var logger = flogging.MustGetLogger("fpc-client-crypto")
 
-func encrypt(input []byte, encryptionKey []byte) ([]byte, error) {
-	return input, nil
+func keyGen() ([]byte, error) {
+	// generate random symmetric key of 16 bytes
+	// the length is required by the pdo crypto library
+	key := make([]byte, 16)
+	n, err := rand.Read(key)
+	if n != len(key) || err != nil {
+		return nil, err
+	}
+
+	return key, nil
 }
 
-func keyGen() ([]byte, error) {
-	return []byte("fake key"), nil
+func encrypt(input []byte, encryptionKey []byte) ([]byte, error) {
+	//This is an RSA encryption performed with the pdo crypto library
+	//Importantly, the library uses 2048bit RSA keys, so the input size has to be ~200bytes
+	//TODO-1: bump up the key length to 3072
+	//TODO-2: extend procedure for large input sizes
+
+	inputMessagePtr := C.CBytes(input)
+	defer C.free(inputMessagePtr)
+
+	encryptionKeyPtr := C.CBytes(encryptionKey)
+	defer C.free(encryptionKeyPtr)
+
+	//pdo crypto uses 2048bit (256bytes) RSA keys, the encrypted message size buffer is set accordingly
+	const encryptedMessageSize = 256
+	encryptedMessagePtr := C.malloc(encryptedMessageSize)
+	defer C.free(encryptedMessagePtr)
+
+	encryptedMessageActualSize := C.uint32_t(0)
+
+	ret := C.pk_encrypt_message((*C.uint8_t)(encryptionKeyPtr), C.uint32_t(len(encryptionKey)), (*C.uint8_t)(inputMessagePtr), C.uint32_t(len(input)), (*C.uint8_t)(encryptedMessagePtr), C.uint32_t(encryptedMessageSize), &encryptedMessageActualSize)
+	if ret == false {
+		return nil, fmt.Errorf("encryption failed")
+	}
+
+	return C.GoBytes(encryptedMessagePtr, C.int(encryptedMessageActualSize)), nil
 }
 
 func decrypt(encryptedResponse []byte, resultEncryptionKey []byte) ([]byte, error) {
-	return encryptedResponse, nil
+	encryptedResponsePtr := C.CBytes(encryptedResponse)
+	defer C.free(encryptedResponsePtr)
+
+	resultEncryptionKeyPtr := C.CBytes(resultEncryptionKey)
+	defer C.free(resultEncryptionKeyPtr)
+
+	// the (decrypted) response size is estimated to be <= the encrypted response size
+	responseSize := len(encryptedResponse)
+	responsePtr := C.malloc(C.ulong(responseSize))
+	defer C.free(responsePtr)
+
+	responseActualSize := C.uint32_t(0)
+
+	ret := C.decrypt_message((*C.uint8_t)(resultEncryptionKeyPtr), C.uint32_t(len(resultEncryptionKey)), (*C.uint8_t)(encryptedResponsePtr), C.uint32_t(len(encryptedResponse)), (*C.uint8_t)(responsePtr), C.uint32_t(responseSize), &responseActualSize)
+	if ret == false {
+		return nil, fmt.Errorf("decryption failed")
+	}
+
+	return C.GoBytes(responsePtr, C.int(responseActualSize)), nil
 }
 
 type EncryptionProvider interface {
@@ -57,6 +107,11 @@ func (e EncryptionProviderImpl) NewEncryptionContext() (EncryptionContext, error
 	}
 
 	ccEncryptionKey, err := e.GetCcEncryptionKey()
+	if err != nil {
+		return nil, err
+	}
+	//decode key
+	ccEncryptionKey, err = base64.StdEncoding.DecodeString(string(ccEncryptionKey))
 	if err != nil {
 		return nil, err
 	}
--- a/client_sdk/go/fpc/crypto/crypto.go
+++ b/client_sdk/go/fpc/crypto/crypto.go
@@ -31,9 +31,9 @@ import "C"
 var logger = flogging.MustGetLogger("fpc-client-crypto")
 
 func keyGen() ([]byte, error) {
-	// generate random symmetric key of 16 bytes
-	// the length is required by the pdo crypto library
-	key := make([]byte, 16)
+	// the specified key length is required by the pdo crypto library
+	keyLength := C.SYM_KEY_LEN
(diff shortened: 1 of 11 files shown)


client_sdk/go/fpc/crypto/crypto.go
```diff
@@ -29,16 +30,65 @@ import "C"
...
 }
 
-func keyGen() ([]byte, error) {
-	return []byte("fake key"), nil
+func encrypt(input []byte, encryptionKey []byte) ([]byte, error) {
+	//This is an RSA encryption performed with the pdo crypto library
+	//Importantly, the library uses 2048bit RSA keys, so the input size has to be ~200bytes
+	//TODO-1: bump up the key length to 3072
+	//TODO-2: extend procedure for large input sizes
+
```
g2flyer, line 50: Some additional background/motivation
```suggestion
	//Importantly, the library uses 2048bit RSA keys & OAEP encoding, so the input size can be at most ~200bytes
	//TODO-1: bump up the key length to 3072 to match NIST strength
	//TODO-2: extend procedure for large input sizes (via hybrid encryption)

```

client_sdk/go/fpc/crypto/crypto.go
```diff
@@ -29,16 +30,65 @@ import "C"
...
+	inputMessagePtr := C.CBytes(input)
+	defer C.free(inputMessagePtr)
+
+	encryptionKeyPtr := C.CBytes(encryptionKey)
+	defer C.free(encryptionKeyPtr)
+
+	//pdo crypto uses 2048bit (256bytes) RSA keys, the encrypted message size buffer is set accordingly
+	const encryptedMessageSize = 256
```
g2flyer, line 58: See above regarding use of const from PDO C code. Given that we also have a const for max message size, maybe also a good idea to test here and give a good corresponding log-message (enclave logs might not be built-in, in which case error tracing will be a bit more of a challenge ...)

common/enclave/cc_data.h
```diff
@@ -45,6 +45,9 @@ class cc_data
     ByteArray get_state_encryption_key();
     std::string get_enclave_id();
     bool sign_message(const ByteArray& message, ByteArray& signature) const;
+    bool decrypt_cc_message(const ByteArray& encrypted_message, ByteArray& message) const;
```
g2flyer, line 48: the `_cc_` seems a bit inconsistent given that we don't do it for sign and encrypt ...

ecc_enclave/enclave/enclave.cpp
```diff
@@ -99,13 +114,18 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
         // TODO: encrypt fabric Response object
         crm = {};
 
+        {  // encrypt response
+            encrypted_response = pdo::crypto::skenc::EncryptMessage(return_encryption_key,
+                ByteArray(b64_response.c_str(), b64_response.c_str() + b64_response.length()));
+        }
+
```
g2flyer, line 121: You have an `encrypt_message` in cc-data which would be nicer to use here?  Also would make error checking consistent with above? (right now an error would cause an exception which i'm not sure is even caught be us?)

ecc_enclave/enclave/enclave.cpp
```diff
@@ -134,6 +154,26 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
             crm.has_proposal = true;
         }
 
+        {  // fill chaincode request message hash
+            // hash request
+            ByteArray ba_cc_request_message(
+                cc_request_message_bytes, cc_request_message_bytes + cc_request_message_bytes_len);
+            ByteArray ba_cc_request_message_hash =
+                pdo::crypto::ComputeMessageHash(ba_cc_request_message);
```
g2flyer, line 162: ~this hash looks rather redundant?~

Oh, forgot that we do also send this as part of message to make protobuf easier (even though that of course is also redunant. So hash keeps message smaller. ..

internal/utils/utils.go
```diff
@@ -83,3 +87,93 @@ func ValidateEndpoint(endpoint string) error {
 
 	return nil
 }
+
+func GetChaincodeRequestMessageFromSignedProposal(signedProposal *pb.SignedProposal) (crmProtoBytes []byte, e error) {
```
g2flyer, line 91: would seem more natural to have this in `internal/utils/proto.go`?

g2flyer (approved):
Looks very good and should make complete now FPC 1.0 from an end-to-end security perspective!  Also works nicely in HW mode for both (automated) integration test and (manual) test-net test.   A few comments/observations/suggestions, though; none of them a show-stopper. So take it or leave it.

BTW: i think this should also close #412? Also #486, i guess we could move to FPC 1.1?

common/enclave/cc_data.h
bvavala: I kind of agree.
Thing is that `encrypt_message` uses just symmetric crypto and passes a key, while `decrypt_cc_message` uses (for now just) asymmetric crypto.
So the "inconsistency" aims actually at differentiating this aspect.

client_sdk/go/fpc/crypto/crypto.go
```diff
@@ -29,16 +30,65 @@ import "C"
 
 var logger = flogging.MustGetLogger("fpc-client-crypto")
 
-func encrypt(input []byte, encryptionKey []byte) ([]byte, error) {
-	return input, nil
+func keyGen() ([]byte, error) {
+	// generate random symmetric key of 16 bytes
+	// the length is required by the pdo crypto library
+	key := make([]byte, 16)
```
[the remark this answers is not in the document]
bvavala: Reference... no, because it's buried in Cpp code.
However, yes, I can make "wrapper" constants referring to those.
Done.

client_sdk/go/fpc/crypto/crypto.go
bvavala: copy-pasted.

client_sdk/go/fpc/crypto/crypto.go
bvavala: same answer as above, and added test

ecc_enclave/enclave/enclave.cpp
bvavala: good catch, that function was meant exactly for that

internal/utils/utils.go
bvavala: no strong opinion, but I have moved it

bvavala:
> Looks very good and should make complete now FPC 1.0 from an end-to-end security perspective!

Yes, finally!! 

> BTW: i think this should also close #412? Also #486, i guess we could move to FPC 1.1?

Good observation, yes and yes. Added to the PR comment.

bvavala:
FYI travis fails due to some reached quotas enforced by docker. Given where it failed, I would call it successful.

client_sdk/go/fpc/crypto/crypto.go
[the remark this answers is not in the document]
g2flyer: thanks, should be more robust now and also nicer to read

common/enclave/cc_data.h
g2flyer: not that the name does reflect this distinction? (for that i would have expected something like `_sym_` and `_pk_` or alike).  BTW: also realize now that given that you pass the key (rather taking it from object like for decrypt), it's also less obvious why it is even a method of that object?  Arguably, one could also have set the return key as object field, in which case arguably the sym vs pk distinction at this level of abstraction is actually completely irrelevant?  All that said, the current naming certainly shouldn't lead to any confusion, so no real issue by leaving it as-is, just a bit nitpicking ;-)

g2flyer:
> FYI travis fails due to some reached quotas enforced by docker. Given where it failed, I would call it successful.

This dockerhub limit i've also encountered (regularly) on my local tests. There i solved it by doing a `docker login` with my dockerhub account. Not sure how you solve it for travis. It seems to work on and off, haven't seen a concrete pattern yet.  But given that we might have to move to azure pipelines or alike, probably not worth spending much time on it.

g2flyer (approved):
new version also works fine in HW mode and changes look good

g2flyer:
> new version also works fine in HW mode and changes look good

BTW: as you labelled this as security-related issue, should maybe @mbrandenburger also have a look for a second pair? (Not sure when he is back).  Otherwise, i'll merge ...

bvavala:
> BTW: as you labelled this as security-related issue, should maybe @mbrandenburger also have a look for a second pair? (Not sure when he is back). Otherwise, i'll merge ...

I would prefer to merge this, as well the upcoming one on mock_enclave, and do additional revisions/modifications post-facto.
Also, likely I won't be available to push fixes, so either you or Marcus may want to proceed in a separate PR.

g2flyer:
> I would prefer to merge this, 

ok, i'll merge 

> as well the upcoming one on mock_enclave, and do additional revisions/modifications post-facto.
> Also, likely I won't be available to push fixes, so either you or Marcus may want to proceed in a separate PR.

As mentioned on slack, i'm off work this and next week, so will do unlikely much until Jan 11th, so will do unlikely much beforehand

PR #520, opened 2021-01-14
Add integration test using client sdk

### LINKED ISSUE #489: Go client sdk-based test for integration tests
**Description**

Version of equivalent of (main part) of `integration/echo_test.sh` and/or `integration/auction_test.sh` based on a  Go Application so travis would exercise also automatically all of client-sdk (part of client-sdk should already be done as part of cli-integration in issue #486.

Setup & Chaincode creation might be best used the script-based version currently used in theses scripts as orthogonal to the actual chaincode application. The chaincode creation could eventually be supplemented by a fully-programmatic version, e.g., as part of issue #488 

Probably best done together with #490 given the large overlap?

**Link to feature or bug**

**Link to dependent issues**

mbrandenburger:
- Port existing auction integration test using the FPC client go sdk

Fixes #489

The is triggered by running `make test` in `integration/`

```

### CODE REVIEW

**Comment by mbrandenburger** [comment]
currently fails with the following error:
```
      Unexpected error:
          <*errors.errorString | 0xc0005382e0>: {
              s: "Failed to get network: Failed to create new channel client: event service creation failed: could not get chConfig cache reference: QueryBlockConfig failed: no channel peers configured for channel [mychannel]",
          }
          Failed to get network: Failed to create new channel client: event service creation failed: could not get chConfig cache reference: QueryBlockConfig failed: no channel peers configured for channel [mychannel]
      occurred
```

**Comment by mbrandenburger** [comment]
I fixed the problems with the client SDK connecting to the integration-test network. Please have a look.

File: fabric/bin/lib/common_ledger.sh [hunk 1]
```diff
@@ -39,7 +39,7 @@ ledger_precond_check() {
 define_common_vars() {
     # use our wrapper for commands, so provide convience functions ..
     # note: as we might have defined FABRIC_CFG_PATH in this script, we have the pass it along!
-    PEER_CMD="env FABRIC_CFG_PATH=${FABRIC_CFG_PATH} ${FABRIC_SCRIPTDIR}/peer.sh"
+    PEER_CMD="env FABRIC_CFG_PATH=${FABRIC_CFG_PATH} CORE_PEER_GOSSIP_EXTERNALENDPOINT=127.0.0.1:7051 ${FABRIC_SCRIPTDIR}/peer.sh"
```
g2flyer, line 42: strange that you would need it for client-sdk. From core.yml this seems to be required for org-external inter-peer communication which doesn't seem to be case here?    Independent of that, might be better to define this also in `core.yaml` (including comment why needed) so we have all these FPC integration-test specific stuff in a mostly single place?

integration/Makefile
```diff
@@ -17,3 +17,8 @@ echo_test:
 
 deployment_test:
 	./deployment_test.sh
+
+client_sdk_test:
+	cd client_sdk/go/auction && \
```
g2flyer, line 22: If you create this deep dir structure, wouldn't then for consistency reasons the old scripts have to move to `client_sdk/cli` or alike?

integration/client_sdk/go/auction/auction_test.go
```diff
@@ -0,0 +1,181 @@
...
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(network).ShouldNot(BeNil())
+
+		//admin := fpc.GetManagementAPI(network, ccID)
+		//
+		//// Setup Chaincode Enclave
+		//attestationParams := []string{"some params"}
+		//err = admin.InitEnclave("peer0.peer1.example.com:7051", attestationParams...)
+		//Expect(err).ShouldNot(HaveOccurred())
+
```
g2flyer, line 44: why keep this as comment?

```suggestion
```

integration/client_sdk/go/auction/auction_test.go
```diff
@@ -0,0 +1,181 @@
...
+		contract = fpc.GetContract(network, ccID)
+		Expect(contract).ShouldNot(BeNil())
+
+		res, err := contract.SubmitTransaction("init", "MyAuctionHouse")
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(res).Should(Equal([]byte("OK")))
+	})
+
+	Context("Scenario 1", func() {
+		When("close non existing auction.", func() {
```
g2flyer, line 56: The scenarios below test mostly the auction chaincode, which is also tested in the  cli-based auction test, but not the client-sdk itself.  Might maybe be good to add also some scenario of testing client-sdk itself?  

Some additional scenarios which come in mind:
- a query (below are all invokes?)
- a query/tx on a non-existing chaincode
- a query/tx on an installed but not initialized (and hence not registered) chaincode

g2flyer (commented):
Overall looks good.  A few suggestions, with the main one maybe adding some scenarios which complement rather than replicate the old auction test (testing the chaincode).  Also, at least for me linter fails as it complains about some `gofmt` errors in fabric-samples.  Strange why that happens only now, though. In any case, we should prune fabric-samples from the linter checks (ie. an additional regexp entry in the exclusion list in `scripts/function.sh` ...

integration/client_sdk/go/auction/auction_test.sh
```diff
@@ -0,0 +1,80 @@
...
+
+CC_ID=auction_test
+CC_PATH=${FPC_PATH}/examples/auction/_build/lib/
+CC_LANG=fpc-c
+CC_VER="$(cat ${CC_PATH}/mrenclave)"
+CC_SEQ="1"
+CC_EP="OR('SampleOrg.member')" # note that we use .member as NodeOUs is disabled with the crypto material used in the integration tests.
+
+NUM_FAILURES=0
+NUM_TESTS=0
```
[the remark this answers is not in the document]
mbrandenburger: removed

fabric/bin/lib/common_ledger.sh
mbrandenburger: it seems the client sdk relies on the discovery service. Without this external endpoint setting, it seems that the client cannot discover the peer. I moved this configuration from `common_ledger.sh` to `core.yaml` in `integration/config.yaml` as suggested

integration/Makefile
mbrandenburger: let's keep the CLI tests where they are, at least for now. I moved to go tests now to `client_sdk/go` and the corresponding shell script to `integration` as this also uses CLI.

integration/client_sdk/go/auction/auction_test.go
mbrandenburger: removed

integration/client_sdk/go/auction/auction_test.go
mbrandenburger: I added test for the suggested scenarios.

mbrandenburger:
> Also, at least for me linter fails as it complains about some `gofmt` errors in fabric-samples. Strange why that happens only now, though. In any case, we should prune fabric-samples from the linter checks (ie. an additional regexp entry in the exclusion list in `scripts/function.sh` ...

we are sym linking the function script from the fabric repo to ours, so we cannot easily add something to the exclusion list. Maybe just make a PR on fabric samples and lint there? :D

g2flyer:
> we are sym linking the function script from the fabric repo to ours, so we cannot easily add something to the exclusion list. Maybe just make a PR on fabric samples and lint there? :D

We already replaced symlinks for `check_license.sh`, so we might as well do also for `golinter.sh`/functions.sh`?  While not ideal, quickest way to get functioning CI back?

g2flyer (commented):
Overall changes look good. Build & tests also work in both SIM and HW mode.   Note though that now linter fails because of _our code_.  As mentioned also above, i actually revert to my opinion that fixing linter already as part of this PR would be the more prudent way.

mbrandenburger:
@g2flyer I've created PR #524 to address the linting issue with fabric samples.
I also fixed the linting issue in `auction_test.go`, so once PR #524 this here should go through smoothly.

bvavala (approved):
lgtm

PR #522, opened 2021-01-19
Add lifecycle support to FPC client SDK

### LINKED ISSUE #488: Revisit Go Admin SDK
**Description**

Go SDK did add now programmatic support for Lifecycle v2 in https://godoc.org/github.com/hyperledger/fabric-sdk-go/pkg/client/resmgmt.  Hence we might 
- [ ] revisit our current API in light of above to be consistent (as done for non-admin client sdk)
- [ ] revisit our test/sample program to do also chaincode creation programmatically rather than by cli (i.e., such that we have two ways to create, either fully programmatic or fully peer clie based, as currently done in integration-tests)

**Link to feature or bug**

**Link to dependent issues**

mbrandenburger:

- Implement LifecycleInitEnclave on top of resmgmt package
- Implement NewCCPackage
- Add SGX attestation paramer API
- Integrate LifecycleInitEnclave and NewCCPackage in integration test
- Rename client_sdk/go/test to client_sdk/go/sample
- Refactor client sdk structure to be aligned with fabric client go sdk
- Update godoc and references

Fixes #488

TODOs:

- [x] Implement fpc chaincode packaging
- [x] Implement LifecycleInitEnclave
- [x] Test with integration test
- [x] Better godoc
- [x] cross-check godoc references in .md files

Remaining todos are covered in a separate PR.
- [ ] Test with test-network
- [ ] Pass `resmgmt.RequestOptions` to underlying channel client

```

### CODE REVIEW

File: integration/client_sdk/go/utils.go [hunk 1]
```diff
@@ -97,3 +115,158 @@ func setupNetwork(channel string) (*gateway.Network, error) {
...
+
+	initTxId, err := client.LifecycleInitEnclave(nw.ChannelID, initReq,
+		resmgmt.WithRetry(retry.DefaultResMgmtOpts),
+		resmgmt.WithTargetEndpoints(nw.Peers...), // peers that are responsible for enclave registration
+		resmgmt.WithOrdererEndpoint(nw.Orderers[0]),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to init enclave: %v", err)
+	}
+	logger.Infof("%s successfully initialized enclave: %v", cc.Id, initTxId)
```
mbrandenburger, line 269: Please see this as an example how `LifecycleInitEnclave` is supposed to use.

integration/client_sdk/go/utils.go
```diff
@@ -97,3 +115,158 @@ func setupNetwork(channel string) (*gateway.Network, error) {
...
+	desc := &fpc.Descriptor{
+		Path:  cc.Path,
+		Type:  cc.Lang,
+		Label: cc.Id,
+	}
+	ccPkg, err := fpc.NewCCPackage(desc)
+	if err != nil {
+		return fmt.Errorf("failed to create new chaincode package: %v", err)
+	}
+	logger.Infof("%s successfully packaged", cc.Id)
```
mbrandenburger, line 181: This is how packaging is supposed to be used.

integration/client_sdk/go/utils.go
g2flyer: looks nice and uniform to the existing lifecycle events

integration/client_sdk/go/utils.go
g2flyer: Main suggestion i would have regarding naming:
- instead `Descriptor` call it `...NewCCPackageRequest`
- prefix request type and operation with `Lifecycle`
This would make it even more uniform with the other operations.

The other thought i would have is whether the sgx-mode we might expose also in the request (maybe still with the env-var based derivation if it is not explicitly set in request)?

On a related note, don't forget to add godoc comments!

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,204 @@
...
+	"strings"
+
+	"github.com/hyperledger/fabric/core/chaincode/persistence"
+	"github.com/hyperledger/fabric/core/chaincode/platforms/util"
+	"github.com/pkg/errors"
+)
+
+/*
+Notice: This file is has been modified for Fabric Private Chaincode.
+*/
```
g2flyer, line 21: A bit a confusing comment: I guess you copy/pasted it from somewhere before starting modification? If so, you should say where you got if from ...

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Descriptor holds the package data
+type Descriptor struct {
+	Path  string
+	Type  string
+	Label string
+}
+
+// ReadMrenclave returns mrenclave for the given chaincode at ccPath
+func ReadMrenclave(ccPath string) (string, error) {
```
g2flyer, line 56: Is this (and below `Validate`) intentionally a public function?  For a public API i think we should minimize what is exposed ...

integration/client_sdk/go/auction_test.go
```diff
@@ -22,24 +23,48 @@ func TestGoClientSDK(t *testing.T) {
...
+
+	//// get admin client (auction)
+	//admin := fpc.GetManagementAPI(network, ccID)
+	//
+	//// Setup Auction Chaincode Enclave
+	//attestationParams := []string{"some params"}
+	//err = admin.InitEnclave("127.0.0.1:7051", attestationParams...)
+	//Expect(err).ShouldNot(HaveOccurred())
+	//logger.Infof("%s successfully performed initEnclave", ccID)
+
```
g2flyer, line 59:
```suggestion
```

bvavala (commented):
Nice handling of lifecycle from the "FPC client"

client_sdk/go/fpc/packager.go
mbrandenburger: the packaging code is based on `fabric-sdk-go/pkg/fab/ccpackager/lifecycle/packager.go `

client_sdk/go/fpc/packager.go
mbrandenburger: `ReadMrenclave` should be a public function as it is used by the sdk users to get the mrenclave for the other lifecycle operations.

`Validate` can be private. In the go sdk it is also public but only used within its package.

integration/client_sdk/go/utils.go
mbrandenburger: Regarding naming, I was following the original naming the go sdk - I would keep the names as they are to keep the relationship close to the go sdk. Same for moving NewCCPackage to the client.

> The other thought i would have is whether the sgx-mode we might expose also in the request (maybe still with the env-var based derivation if it is not explicitly set in request)?

yes, good idea.

integration/client_sdk/go/utils.go
```diff
@@ -97,3 +115,158 @@ func setupNetwork(channel string) (*gateway.Network, error) {
...
+	if err != nil {
+		return err
+	}
+
+	ccDetails := &testutils.ChaincodeDetails{
+		Id:           ccID,
+		Path:         ccPath,
+		Version:      mrenclave,
+		Lang:         fpc.ChaincodeLang,
+		Policy:       policydsl.SignedByAnyMember([]string{"SampleOrg"}),
```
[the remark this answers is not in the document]
mbrandenburger: Oh you made a good point here. Since this code is integration test only and the current setting the actual policy picked here is not so important (functional), however, as this integration test may also function as documentation and as an example how to use FPC this *anyMember* policy is certainly not good. ACK.

Here in this test we can just unset the policy and thereby use the default implicit policy. But I see a problem here that we do not document this in code properly.  We could provide a `fpcpolicy.MajorityOfMembers(members []string),` function that we use throughout our tests. Thereby by reading our code it is clear that we always use a "strong" policy.

integration/client_sdk/go/utils.go
g2flyer: > Regarding naming, I was following the original naming the go sdk - I would keep the names as they are to keep the relationship close to the go sdk. Same for moving NewCCPackage to the client.

Oops, forgot about the packager package in standard go and wrongly thought there wouldn't be any packaging in the go framework.  Indeed, your proposal does matches  the various forms from standard sdk.

client_sdk/go/fpc/management.go
```diff
@@ -71,58 +65,40 @@ func (c *managementState) InitEnclave(peerEndpoint string, attestationParams ...
 		return err
 	}
 
-	// Set attestation paramaters
-	type Params struct {
-		AttestationType string `json:"attestation_type"`
-		HexSpid         string `json:"hex_spid"`
-		SigRL           string `json:"sig_rl"`
+	// TODO parse attestationParams
+	// Currently we ignore the function inputs and read the attestation params from environment, needs to be fixed
```
g2flyer, line 69: I think the best way would be to define the attestationParams as optional (e.g., by having a separate attestationParam-less version of `InitEnclave`, with this one calling `CreateAttestationParamsFromEnvironment` and then calling the version with the explicit arg?)

client_sdk/go/fpc/packager.go
g2flyer: ah, ok.

integration/client_sdk/go/utils.go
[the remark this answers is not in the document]
g2flyer: Hmm, i don't think you can implement in a robust way with current fabric a MajorityOfMembers. Per-chaincode fabric allows only signature policies; majority can only be defined in configtx.yaml and alike.   Note as (number of) orgs might change over time, you won't be able to define a robust majority based on signature policies!
In any case, this is the same instance of the issue we discussed last week for the integration-test cli and i thought the conclusion was that we just add a corresponding comment?

client_sdk/go/fpc/management.go
g2flyer: Actually, i just realize that i'm a bit confused. Isn't this function essentially now a private one as folks should use the LifecycleInitEnclave one?  Having it public would seem somewhat confusing?

And from that perspective, i think also above suggestion doesn't make sense.  Seems to me we actually should handle that via `LifecycleInitEnclaveRequest ` and handle it similar to `SGX_MODE`. There is a field which when defined is used from the request and if not defined we read it from the environment?

mbrandenburger: well, I guess the `InitEnclave` command in management.go will disappear and is replaced with the `LifecycleInitEnclave` command. 

With the new `LifecycleInitEnclave` the attestation params are passed via `LifecycleInitEnclaveRequest`. So the user is responsible to gather the attestation params. For convenience we can provide methods like ...

```go
func CreateAttestationParamsFromEnvironment() (*AttestationParams, error) {}
func CreateAttestationParamsFromCredentialsPath(sgxCredentialsPath string) (*AttestationParams, error) {}
...
```

integration/client_sdk/go/utils.go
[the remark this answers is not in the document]
mbrandenburger: yes, you are right. In our discussion yesterday with Bruno we also agreed that my proposed `MajorityOfMembers` function is not strong enough for dynamic membership on a channel. I added some more comments on this function and devoted it for testing only.

client_sdk/go/fpc/management.go
g2flyer: > well, I guess the `InitEnclave` command in management.go will disappear and is replaced with the `LifecycleInitEnclave` command.

I thought you actually still call `InitEnclave` from `LifecycleInitEnclave` in which case it couldn't really disappear?  That said, i don't really care about either disappearing or being private as long as it wouldn't be visible in the public interface/godoc anymore :-)

> With the new `LifecycleInitEnclave` the attestation params are passed via `LifecycleInitEnclaveRequest`. So the user is responsible to gather the attestation params. For convenience we can provide methods like ...
> 
> ```go
> func CreateAttestationParamsFromEnvironment() (*AttestationParams, error) {}
> func CreateAttestationParamsFromCredentialsPath(sgxCredentialsPath string) (*AttestationParams, error) {}
> ...
> ```

I thought you already had some public functions to get them (parts of it) from environment?   In any case, my proposal to them optional parameter which if absent would be automatically filled from environment was just a thought; i'm fine to always require them in the request object ...

mbrandenburger: > I thought you actually still call InitEnclave from LifecycleInitEnclave in which case it couldn't really disappear? 

no, I am not calling the `initEnclave` from mangement.go as this is based on the gateway API which does not interact nicely with the "other" sdk clients such as channel or resmgmt client.

> That said, i don't really care about either disappearing or being private as long as it wouldn't be visible in the public interface/godoc anymore :-)
Sure, I still need to do that.

integration/client_sdk/go/utils.go
```diff
@@ -97,3 +123,187 @@ func setupNetwork(channel string) (*gateway.Network, error) {
...
+	if err != nil {
+		return err
+	}
+
+	ccDetails := &testutils.ChaincodeDetails{
+		Id:           ccID,
+		Path:         ccPath,
+		Version:      mrenclave,
+		Lang:         fpc.ChaincodeLang,
+		Policy:       fpcpolicy.MajorityOfMembers(consortium),
```
g2flyer, line 153: As mentioned on the call, from a blind copy&paste perspective i think this is even more brittle than the old single org policy. So a forward comment would be useful.  A url to the godoc would be even better but i was too lazy to figure that out :-)
```suggestion
		Policy:       fpcpolicy.MajorityOfMembers(consortium), // Note this assumes a static network, see comments on this function
```
```suggestion
		Policy:       fpcpolicy.MajorityOfMembers(consortium),
```

mbrandenburger: oh didn't see this suggestion before I pushed an update with a similar comment :)

g2flyer: > oh didn't see this suggestion before I pushed an update with a similar comment :)

my comment hit github probably just shortly before you pushed.  That said, there are two aspects i like about mine: having it on the same line as the function is a bit more robust from copy/paste perspective and having the forward reference to the (more extensive) godoc comment on the function itself.  Arguably, as long as you have a static network and majority is what you want/need this is robust, so we do not really need the "testing" disclaimer?

mbrandenburger:
Note this PR currently contains the content of #529. Once it is merged, this PR should be rebased.

client_sdk/go/fpc/platform/params.go
```diff
@@ -0,0 +1,126 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package platform
```
g2flyer, line 7: Do we need a new package?  In particular, right now it seems purely attestation specific, so platform is not an intuitive name?  In particular as you abstract attestation, i.e., it's not that you define a SGX-EPID platform or alike?   This question mainly arose as i saw all these public functions below which i felt the developer wouldn't really have a need to see. But with a separate package i guess you had no choice ...

client_sdk/go/fpc/platform/params_test.go
```diff
@@ -0,0 +1,7 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package platform
```
g2flyer, line 7: I guess a `// TODO: implement me ...` is missing here :-)

client_sdk/go/fpc/policy/policy.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/hyperledger/fabric-protos-go/msp"
+	"github.com/hyperledger/fabric/common/policydsl"
+)
+
+// MajorityOfMembers returns an endorsement policy that requires the signatures
+// by a majority of the given members.
+//
+// Note that this endorsement policy is not flexible enough to support membership
+// changes. A more robust majority-based endorsement policy can be provided
+// through the channels default endorsement policy.
```
g2flyer, line 20: I think this should be a stronger message given the potential insecurity!   

BTW: i hope we are actually also allowing _not_ specyfing a policy so the default policy applies (another advantage, besides the inherent robustness, of my proposal to drop this and just use in both test-cases the default policy ... ;-))

```suggestion
//
// IMPORTANT: Note that this endorsement policy is not flexible enough to support membership changes and _might become unintendently *insecure* when membership increases. 
// A more robust majority-based endorsement policy can be defined 
// by not specifying an explicit policy and hence applying the channels default endorsement policy (which is majority in the default fabric network configurations).
```

client_sdk/go/fpc/resmgmt/lifecycleclient_test.go
```diff
@@ -0,0 +1,7 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package resmgmt
```
g2flyer, line 7: I guess a '// TODO: Implement me' is missing here (or we could have left out the file completely?)

client_sdk/go/samples/admin/main.go
```diff
@@ -0,0 +1,347 @@
+/*
```
g2flyer, line 1: I guess this code is eventually going to replace the `setup` code from below (`utils.go` if i remember correctly)?  If so i would keep the testing code in integration tests and not trying to add samples already in this PR.

This also would remove the need for the `testing` package which too me doesn't seem that motivated as a package. Also if this really is a sample code, then it shouldn't rely on internal packages and anything samples?  If it is really testing (and as mentioned above it should hence be moved to `integration/....` then there is no need for a global testing package ...

g2flyer (commented):
a few more comments ...

mbrandenburger:
@bvavala @g2flyer please have a look. I squashed my previous fixups together.
Note that build fails due to some docker issue. No idea why this is happening now.

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,238 @@
...
+const (
+	codePackageName      = "code.tar.gz"
+	metadataPackageName  = "metadata.json"
+	mrenclaveFileName    = "mrenclave"
+	enclaveBinaryName    = "enclave.signed.so"
+	gzipCompressionLevel = gzip.DefaultCompression
+	ChaincodeLang        = "fpc-c"
+)
+
+// NewCCPackage creates a fpc chaincode package
```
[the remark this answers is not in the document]
mbrandenburger: See updated implementation. Now, `NewCaaSPackage` is also integrated in `NewCCPackage`. Depending on the package type set in the desc object the right implementation is picked.

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if err := persistence.ValidateLabel(p.Label); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type CaaSDescriptor struct {
+	Name                string
```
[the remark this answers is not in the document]
mbrandenburger: this is removed

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,238 @@
...
+	PeerAddress         string
+}
+
+func (p *CaaSDescriptor) validate() error {
+	// TODO implement me
+	return nil
+}
+
+// PackageMetadata holds the path and type for a chaincode package
+type PackageMetadata struct {
```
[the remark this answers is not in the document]
mbrandenburger: not really. moved to a local type

client_sdk/go/fpc/packager.go
```diff
@@ -0,0 +1,238 @@
...
+type Descriptor struct {
+	Path    string
+	Type    string
+	Label   string
+	SGXMode string
+}
+
+// validate validates the package descriptor
+func (p *Descriptor) validate() error {
+	if p.Path == "" {
```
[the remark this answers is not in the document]
mbrandenburger: would be nice; but this would require a bit more refactoring here. We can add a `TODO implement enclave file(s) exist(s) check`

client_sdk/go/fpc/platform/params.go
mbrandenburger: ok platform is not ideal. I moved the attestation stuff to a `SGX` package.

client_sdk/go/fpc/resmgmt/lifecycleclient_test.go
mbrandenburger: Added the TODO. I think having this file with a TODO produces more pressure also to write this test code!

client_sdk/go/samples/admin/main.go
mbrandenburger: I removed the dependency to `internal/testing`

client_sdk/go/pkg/fab/ccpackager/packager.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	return strings.TrimSuffix(string(mrenclave), "\n"), nil
+}
+
+// Descriptor holds the package data. FPC supports two types of packages, ChaincodeType and CaaSType.
+// For normal chaincode deployments, the package type ChaincodeType is used. It requires to define Type, Label, Path, and SGXMode.
+// Alternatively, for deployments as Chaincode as a Service (CaaS), the package type CaaSType is used.
+// It requires to define Type, Label, Path, and CaaSEndpoint. Optionally, CaaSTimeout and CaaSUseTLS can be set.
+type Descriptor struct {
```
g2flyer, line 86: Curious why you merged the Descriptor. I guess you document in top-level comment which type needs which descriptor.  However, wouldn't it have been a cleaner way to define a base type with the common fields and two extension types, one for each deployment type?  In that case, you actually wouldn't even have to explicitly add the `Type` field as it would be defined by the type?

client_sdk/go/fpc/packager.go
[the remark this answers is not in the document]
g2flyer: you added documented but i'm still not sure folks immediately will connect it with the two different external builder setups?

client_sdk/go/fpc/packager.go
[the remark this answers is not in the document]
g2flyer: ok (but `TODO` doesn't exist yet?)

g2flyer (approved):
The change addressed most of my comments (there are though a few you haven't replied, some of them potentially hidden in the "hidden conversations").  There is a surprising amount of renaming and restructuring  but i guess you tried to mimic more what fabric-go-sdk does?  I also have the feeling we are a bit excessive in the number of packages but that it is obviously a matter of taste.   I have also a few comments/suggestions directly on the files, but they are take-it-or-leave-it as we are already in a good state as-is and hence i already approve.  

Note though besides reading the code on github, I did only perfunctorily check the go docs (via `go doc -all ./client_sdk/go/pkg/client/resmgmt/` and alike ...) and did zero testing.  Also no immediate idea with the github action build failed the way it did. Maybe some temporary network issue or alike with dockerhub?

bvavala (approved):

client_sdk/go/fpc/packager.go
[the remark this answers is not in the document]
mbrandenburger: We can improve this with the followup PR where we will use this CaasPackage with the test-network.

client_sdk/go/fpc/packager.go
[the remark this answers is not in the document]
mbrandenburger: added

client_sdk/go/pkg/fab/ccpackager/packager.go
mbrandenburger: oh true, I agree that the merged Descriptor is not the cleanest solution. Let me revisit this in the followup PR. My main motivation was just have a single `NewCCPackage` function in our API. But since the CaaSPackage might also work for non FPC chaincode, in particluar, it is also used for packaging ERCC, having a separate function for CaaS might be better.

client_sdk/go/pkg/client/resmgmt/lifecycleclient.go
```diff
@@ -0,0 +1,178 @@
+/*
```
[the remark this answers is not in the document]
mbrandenburger: good point, I added the logging statements to LifecycleInitEnclave. There is no logging in packager at the moment.

integration/client_sdk/go/utils.go
```diff
@@ -1,16 +1,81 @@
+/*
```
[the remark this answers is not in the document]
mbrandenburger: well, the admin part is also "covered" (or used) by the integration test for the client_sdk. But is true that we do not check different scenarios as we do with the gateway api. Let me add a TODO

client_sdk/go/pkg/fab/policy/policy.go
```diff
@@ -0,0 +1,40 @@
...
+// MajorityOfMembers returns an endorsement policy that requires the signatures
+// by a majority of the given members.
+//
+// IMPORTANT: Note that this endorsement policy is not flexible enough to support
+// membership changes and _might become *insecure* when membership increases.
+// A more robust majority-based endorsement policy can be defined by not
+// specifying an explicit policy and hence applying the channels default endorsement
+// policy (which is "majority" in the default fabric network configurations).
+// It is recommended to use this function in testing environments only where a
+// static channel membership. For other deployments please use a
```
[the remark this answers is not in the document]
mbrandenburger: removed

mbrandenburger:
Added another fixup to integrate last round of valuable comments from @g2flyer. Please check, squash, and merge.

integration/client_sdk/go/utils.go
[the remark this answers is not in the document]
g2flyer: Definitely didn't think it should be done in this PR, so TODO is the right thing ..

client_sdk/go/pkg/fab/ccpackager/packager.go
g2flyer: But in case you want CaaS more general, then i think the old approach with different Descriptor types was better as the descriptor is not really general?  Anyway, something for some future discussion ...

PR #524, opened 2021-01-20
Fix golinter filter

mbrandenburger:
- Move golinter scripts from fabric repo to /scripts
- Exclude integration/test-network/fabric-samples from linter

Fixes #

```

### CODE REVIEW

**Review by g2flyer — COMMENTED** [review]
make fials, but that seems because `internal/utils/validation.go` is not compliant.  See also comment below what i'm not 100% sure we actually filter exactly what we want ...

Somehow, github doesn't allow me to add inline comments, i always get an error "Line must be part of the diff".
So let me give it here with references to files/line

scripts/functions.sh and scripts/golinter.sh:
we should add a header here that we copied it from fabric, similar to what we have done for check_license.sh


scripts/functions.sh, line 37:
Unlikely to be an issue that we will have other .*integration/test-network/fabric-samples dir but to be more precise you would have to prefix it also with a ^ ?.  Also don't you have to escape the /s?  (Strangely, they are handled differently for build and other dirs. Not sure exactly why ...  Did you positively check that we excluded fabric-samples due to this rule or it might simply not have been listed through other parts of the filter (remember, in general the fabric-samples did not get tested other than in your other PR!)


scripts/functions.sh, line 41:
curious, this is an unexpected change from the fabric version:
```
-    files=($@)
+    read -ra files <<<"$@"
```
what was the reason for the change?

mbrandenburger:
> scripts/functions.sh and scripts/golinter.sh:
> we should add a header here that we copied it from fabric, similar to what we have done for check_license.sh

good idea! added this

mbrandenburger:
> make fials, but that seems because `internal/utils/validation.go` is not compliant. See also comment below what i'm not 100% sure we actually filter exactly what we want ...

wired - but here goimports complains.

mbrandenburger:
> scripts/functions.sh, line 37:
> Unlikely to be an issue that we will have other .*integration/test-network/fabric-samples dir but to be more precise you would have to prefix it also with a ^ ?. Also don't you have to escape the /s? (Strangely, they are handled differently for build and other dirs. Not sure exactly why ... Did you positively check that we excluded fabric-samples due to this rule or it might simply not have been listed through other parts of the filter (remember, in general the fabric-samples did not get tested other than in your other PR!)

ok - here the full story. I had to debug a bit in order to find out what is happening with this golinter.
Let me explain how (I understand) it works. First, the scripts searches for go packages using `go list`. The output is reduced to only the first level of directories. For example, the `attestation` package in `ercc/attestation` is simply reduced to `ercc`. The list (or array what ever type that is in bash) is given to `gofmt` for checking. This commands outputs a list (I think as a string) of all go files found in the directories identified before. Finally, the `gofmt` output is given to the filter function, which needs to transform the `gofmt` output to an array using `read`. The array items are checked for generated code and then returned using `echo`.

During testing I made two observations:
- First, with only the `fabric-samples` and no other go files in `integration/`, this folder is not identified by `go list` and thus not checked with `gofmt` at all. I assume this is the reason why we haven't seen the linter errors with fabric-samples before. I guess this `go list` step is just to exclude code, such as `vendor`, from gofmt. You can double check by just running `gofmt -s -l ./..` from `$FPC_PATH` and see that this indeed detects the go files in fabric-samples whereas the "traditional" `golinter.sh` does not detect them.
- Second, transforming the `gofmt` output with `read` does not work as expected within our dev environment. For some reason only the first line of the `gofmt` output was processed so the resulting `files` array contains only a single entry. Not sure why this is not working correctly, I also tried `readarray` instead but only `()` produces the expected array. Note that in the travis test it failed only with a single file in fabric-samples, in particular, with `integration/test-network/fabric-samples/asset-transfer-basic/application-go/assetTransfer.go`. Double checking with `gofmt -s -l ./..` shows that `golinter.sh` does not "detect" (or output) everything it should.

I updated `golinter.sh` and now invoking `gofmt` on the entire repo instead of performing `go list` before. (Note that the script already does it for go vet as well). Then we apply the filter (which works with exclusion of fabric-samples as I added it to the list without any extra escaping.

I also fixed the goimports in `internal/utils/validation.go`

g2flyer (approved):
lgtm

PR #525, opened 2021-01-21
Replace Travis with Github Action

### LINKED ISSUE #476: CI fix/replacement
**Describe the bug**

Do to some recent changes in travis by moving to a credit-based accounting system, FPCs CI doesn't work anymore as hyperledger-labs has exhausted the credits.  

There doesn't seem to be an obvious way to replenish the credits for `travis-ci.com` and `travis-ci.org`  where you still can trigger free manual builds for OSS projects, is going away at end of the year, so trying to repurpose that for our CI doesn't seem worth spending time investigating.

Given that fabric seems to use Azure Pipeline, we might have to transition to that. Technically, it probably isn't too hard judging from [Migrate from Travis to Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/migrate/from-travis?view=azure-devops) but less clear is whether there is a way to get a (for us as project) free service, e.g., by piggybacking on some existing subscriptions or alike.

mbrandenburger:

This PR introduces Github Actions as a replacement for our Travis based automated testing. As we may move forward to use Azure Pipelines in the future, using Github Actions might be only a temporary solution to fix the current issues with Travis.

- Adding Github Action script that triggers our docker based test (based on our travis script)
- Removing Travis script
- Adding Github settings config that allows to disable the "Create a merge commit" Button for PRs :)

Fixes #476 

```


g2flyer (approved):
Nice!

PS: looking at it and noticing that we run as ubuntu-latest, it occurred to me that eventually we might run a second task running the tests also with `DOCKER_BUILD_OPTS=--build-arg UBUNTU_VERSION=20.04 --build-arg UBUNTU_NAME=focal` to not only test for (default) 18.04 but also 20.04 (which should work, but i only occassionally manually tested but is currently not CI-tested).  But definitely something for a future version ...

PR #528, opened 2021-01-29
Some docker tweaks

g2flyer:

The support for addionation docker packages was a bit buggy for base-rt image and non-uniform (and buggy) for dev container. Both should be address in PR.  Also not exporting docker options from host could result in some problems (rsp manual tweaking), this is now automated in this PR.

No existing issue for this (fixing was easier/quicker than creating issue :-)

To test define any (combination) of `DOCKER_BASE_RT_IMAGE_APT_ADD_PKGS`, `DOCKER_BASE_DEV_IMAGE_APT_ADD_PKGS` and/or `DOCKER_DEV_IMAGE_APT_ADD_PKGS` in your `${FPC_PATH}/config.override.mk` and you should see the packages hopefully now in the various containers after a `make -C utils/docker`.  Also a `make -C utils/docker run` should show your '~/.docker` inside the container and hence settings such as proxies should automatically work also inside the dev container ...

```

### DIFF
--- a/utils/docker/dev_peer_cc-builder/Dockerfile
+++ b/utils/docker/dev_peer_cc-builder/Dockerfile
@@ -18,8 +18,8 @@
 #  - fpc branch/tag/commit      FPC_REPO_BRANCH_TAG_OR_COMMIT
 #  - git user:                  GIT_USER_NAME
 #  - git user's email:          GIT_USER_EMAIL
-#  - sgxmode:                   SGX_MODE
-#  - additional dev apt pkgs:   APT_ADD_PKGS
+#  - sgx mode:                  SGX_MODE
+#  - additional apt pkgs:       APT_ADD_PKGS
 
 # Note on SGX_MODE:
 # In a docker build environment, we build but cannot _run_ with SGX_MODE=HW!.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by bvavala — APPROVED** [review]

File: utils/docker/dev_peer_cc-builder/Dockerfile [hunk 1]
```diff
@@ -19,7 +19,7 @@
 #  - git user:                  GIT_USER_NAME
 #  - git user's email:          GIT_USER_EMAIL
 #  - sgxmode:                   SGX_MODE
-#  - additional dev apt pkgs:   APT_DEV_ADD_PKGS
+#  - additional dev apt pkgs:   APT_ADD_PKGS
```
mbrandenburger, line 22:
```suggestion
#  - additional apt pkgs:   APT_ADD_PKGS
```

mbrandenburger (commented):
Can you explain why we mount `~/.docker` now by default?

g2flyer:
> Can you explain why we mount `~/.docker` now by de

Not sure what more than "exporting docker options from host " and "your '~/.docker` inside the container and hence settings such as proxies should automatically work also inside the dev container " you want to know.  But i add also a [doc reference](https://docs.docker.com/engine/reference/commandline/cli/#configjson-properties) in the mix ...

mbrandenburger (approved):

PR #529, opened 2021-02-01
Fix test-network inside dev container

mbrandenburger:
**What this PR does / why we need it**:

This PR addresses the issues identified in #527 

**Which issue(s) this PR fixes**:

Fixes #527 and fixes #519

**Special notes for your reviewer**:

**Does this PR introduce a user-facing changes and/or breaks backward compatability?**:

```

PR #532, opened 2021-02-03
Enable CI for Ubuntu 18.04 and 20.04 + fix READMEs

bvavala:


README.md
```diff
@@ -304,9 +302,15 @@ Make sure that you have the following required dependencies installed:
 
 * [Go](https://golang.org/) 1.15.4 or higher
 
-* Docker 18.x and docker-compose (1.25.x or higher)
+* Docker 18.09 (or higher) and docker-compose 1.25.x (or higher)
```
g2flyer, line 305: I think it would be useful to point out here that 18.04 does not have a good docker compose, copy/pasting from our old proxy text:
```suggestion
* Docker 18.09 (or higher) and docker-compose 1.25.x (or higher)
  Note that version from Ubuntu 18.04 is not recent enough!  To upgrade, install a recent version following the instructions from [docker.com](https://docs.docker.com/compose/install/), e.g., for version 1.25.4 execute	
  ```bash	
  sudo curl -L "https://github.com/docker/compose/releases/download/1.25.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose	
  sudo chmod +x /usr/local/bin/docker-compose	
  ```
```

README.md
```diff
@@ -342,7 +346,7 @@ official [documentation](https://github.com/intel/linux-sgx). Please make sure t
 SDK version as denoted above in the list of requirements.
 
 For SGX SSL, just follow the instructions on the [corresponding
-github page](https://github.com/intel/intel-sgx-Sal). In case you are
```
g2flyer, line 345: How the heck did this typo came in: from git blame it seems it was me who somehow screwed up, no idea how. Anyway, thanks for fixing ...

g2flyer (changes requested):
The README changes look good, with one change suggestion.

That said, the PR currently doesn't really achieves what you want! it _starts_ the build from the two different OS but the actual build/test magic is done inside the container which in either case will be now 20.04. To do this for the two different OS, you have do make sure the docker `--build-args` are set for both `UBUNTU_VERSION` and `UBUNTU_NAME` are set for the two OS version _inside_ the container when building. To force makefiles to do so, you pass the args to make via the `DOCKER_BUILD_OPTS` variable. Note that the earlier version of #529 actually had that in the README and it should be still the visible in the comment section somewhere.  So to make it really build these two version you need a matrix in github action around these variables and pass it to `make` ....

BTW: just noticed, there is also a copy/replace bug in the top comment section of `utils/docker/base-rt/Dockerfile`: the second `UBUNTU_VERSION` should be `UBUNTU_NAME` to match the description.

.github/workflows/test.yaml
```diff
@@ -16,4 +28,5 @@ jobs:
           DOCKER_DEV_CI_MODE: 1
           DOCKER_BUILDKIT: 0
           DOCKER_QUIET_BUILD: 1
-        run: make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env DOCKER_QUIET_BUILD=1 DOCKER_BUILDKIT=0 make all clobber'
\ No newline at end of file
+        run: |
+            DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env DOCKER_QUIET_BUILD=1 DOCKER_BUILDKIT=0 make all clobber'
```
g2flyer, line 32: Hmm, i think you have to add the DOCKER_BUILD_OPTS also to the second `env` clause of the OS choice will not be passed transitively to the docker-builds inside container. The dev container definitely should be built in correct OS but without defining the DOCKER_BUILD_OPTS also inside the container itself the build inside container would always build define 20.04 ?

bvavala: This is incredibly confusing. Let's try to clarify a few things.

The `make run` has some dependencies (on `dev`, `base-dev`, etc.), which build images and pass the `DOCKER_BUILD_OPTS` var. So, the built images should be as we intend them to be.

Then, `make run` enters the container to run what is in `DOCKER_DEV_OPTIONAL_CMD` -- essentially the `make all`.
Most importantly, it does not propagate `DOCKER_BUILD_OPTS`.
Question: does this trigger the (re-)build of the images? (If not, there's nothing to do.)

If so, the dockerfile will use the default value (i.e. 20.04 focal), and that's not intended.
Two solutions:
1. as suggested, pass `DOCKER_BUILD_OPTS` as part of env
2. let `make run` propagate `DOCKER_BUILD_OPTS`
Question: which one is more appropriate? (I'd argue the second one)

Comments?

g2flyer: > This is incredibly confusing. Let's try to clarify a few things.
> 
> The `make run` has some dependencies (on `dev`, `base-dev`, etc.), which build images and pass the `DOCKER_BUILD_OPTS` var. So, the built images should be as we intend them to be.

yes
>
> Then, `make run` enters the container to run what is in `DOCKER_DEV_OPTIONAL_CMD` -- essentially the `make all`.

yes (and it does so with the default environement as defined by default shell and (default) `config.mk` as well as the explicitly defined variables using the `env` prefix ...)

> Most importantly, it does not propagate `DOCKER_BUILD_OPTS`.

Yes as the docker container does not inherit the environment of the process calling the `make run`, i.e., any variable you want to inherit you have to explicitly pass using the `env` (which previously we did for the docker quiet and build options, but now you also would have to do for DOCKER_BUILD_OPTS. An alternative might be pass them also via docker-mounted `config.override.mk` or alike)

> Question: does this trigger the (re-)build of the images? (If not, there's nothing to do.)

`make` _always_ causes a `docker build` (it is nearly impossible in make to add dependencies which would figure out whether the build really can be skipped ...) but docker via caching tries to skip layers with dependencies which have not changed.   But also note that the `make run` from the github action only builds the `dev` container whereas `make all` builds a whole bunch of other ones ...

> If so, the dockerfile will use the default value (i.e. 20.04 focal), and that's not intended.
> Two solutions:
> 
> 1. as suggested, pass `DOCKER_BUILD_OPTS` as part of env
> 2. let `make run` propagate `DOCKER_BUILD_OPTS`
>    Question: which one is more appropriate? (I'd argue the second one)

Both are options, but we should do it uniformly also for the other options/flags and not only for `DOCKER_BUILD_OPTS`. So (1) in a way is the easier option while (2) is the cleaner but requires a bit more thinking what all has to be passed  ....

> Comments?

bvavala: In the interest of time, the easier is preferred. Now we should be good.

.github/workflows/test.yaml
```diff
@@ -16,4 +28,10 @@ jobs:
...
           DOCKER_BUILDKIT: 0
           DOCKER_QUIET_BUILD: 1
-        run: make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env DOCKER_QUIET_BUILD=1 DOCKER_BUILDKIT=0 make all clobber'
\ No newline at end of file
+        run: |
+            DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
+            # we pass again DOCKER_BUILD_OPTS since it's used in `make all` and not propagated by the initial `make run`
+            make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env DOCKER_QUIET_BUILD=1 DOCKER_BUILDKIT=0 \
+                                                              DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
+                                                              make all clobber
```
g2flyer, line 36: Slight rephrase and formatting tweak to point out symmetry of all three env vars 
```suggestion
            DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
            make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env DOCKER_QUIET_BUILD=1\
                                                                  DOCKER_BUILDKIT=0\
                                                                  DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}"\
                                                              make all clobber
            # we pass DOCKER_BUILD_OPTS, similar to DOCKER_QUITE_BUILD and DOCKER_BUILDKIT, to both 'make' invocations as 
            # they are required when building the containers as part of 'make run' and when building inside the container (as part of make all)
            # but make run does not automatically propagate environment variables to the container.
```

g2flyer (approved):
As far as i can tell from the raw logs it should do now what we want.  One small formatting/rephrasing tweak suggestion ...

BTW: one thing which we will have to be mindful, currently it builds all explicitly built images in the right config but wouldn't do so with docker-compose as we had earlier, e.g., for demo.  For that we probably would have to define variables for UBUNTU_VERSION and UBUNTU_NAME inside `config.mk`, always pass these variables as build-args for explicit docker builds and use the variable also inside relevant docker-compose files ....

README.md
```diff
@@ -304,9 +302,19 @@ Make sure that you have the following required dependencies installed:
 
 * [Go](https://golang.org/) 1.15.4 or higher
 
-* Docker 18.x and docker-compose (1.25.x or higher)
+* Docker 18.09 (or higher) and docker-compose 1.25.x (or higher)
+  Note that version from Ubuntu 18.04 is not recent enough!  To upgrade, install a recent version following the instructions from [docker.com](https://docs.docker.com/compose/install/), e.g., for version 1.25.4 execute	
+  ```bash	
+  sudo curl -L "https://github.com/docker/compose/releases/download/1.25.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose	
+  sudo chmod +x /usr/local/bin/docker-compose
```
mbrandenburger, line 309:
```suggestion
  ```bash	
  sudo curl -L "https://github.com/docker/compose/releases/download/1.25.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose	
  sudo chmod +x /usr/local/bin/docker-compose
  ```	
```

README.md
```diff
@@ -304,9 +302,19 @@ Make sure that you have the following required dependencies installed:
...
+    * To install docker-componse 1.25.4 from [docker.com](https://docs.docker.com/compose/install/), execute
+        ```bash
+        sudo curl -L "https://github.com/docker/compose/releases/download/1.25.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+        sudo chmod +x /usr/local/bin/docker-compose
+        ``` 
+
+* yq v3.x (newer versions, v4.x and higher, are currently *not* supported!)
   You can install `yq` v3 via `go get`.
   ```bash
   GO111MODULE=on go get github.com/mikefarah/yq/v3
```
mbrandenburger, line 320: seems codeblock formatting is broken here :/

mbrandenburger (changes requested):
LGTM except of a small issue in the README.md. I guess the wrong formatting was somehow introduced using some click-and-apply suggestion :/

mbrandenburger (approved):
I pushed a fix for the README

PR #556, opened 2021-04-01
Enable coverage report in CI

bvavala:
This PR enables test coverage tracking in FPC repo, in particular:
* it enables tracking through gcc coverage instrumentation and reporting to codecov.io
* instrumentation/reporting is configurable through environment variables; importantly, it is disabled by default, and enabled during CI build
* it instruments some c/cpp libraries (e.g., logging, attestation-api, etc.); others can/will be added later


build.mk
```diff
@@ -14,13 +14,22 @@ ifeq (${SGX_MODE}, HW)
 endif
 GO := $(GO_CMD) $(GOFLAGS)
 
+# in CI build, enable code coverage; disable it otherwise
+ifeq (${IS_CI_RUNNING}, true)
+	#this enables coverage in c/cpp code, libraries and linked binaries
+	export CODE_COVERAGE_ENABLED ?= true
+else
+	export CODE_COVERAGE_ENABLED ?= false
+endif
```
mbrandenburger, line 23: Would that also fit into `config.mk` where also other exports are?

ecc/chaincode/enclave/enclave.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 // #cgo CFLAGS: -I${SRCDIR}/ecc-enclave-include -I${SRCDIR}/../../../common/sgxcclib
-// #cgo LDFLAGS: -L${SRCDIR}/ecc-enclave-lib -lsgxcc
+// #cgo LDFLAGS: -L${SRCDIR}/ecc-enclave-lib -lsgxcc -lgcov
```
mbrandenburger, line 23: is this necessary to to compile ecc now?

mbrandenburger (commented):
Thanks Bruno, this is pretty cool. A few questions below ...

Moreover, can you give some more details on how coverage test is performed? What is the coverage output? How is this data shipped to codecov.io? Maybe we a few lines description how this works inside `report.sh` would be good?! 

Also how is codecov.io integrated in the github PR workflow? I saw that the Fabric SDK Go uses codecov.io as well and they have a [.codecov.yaml](https://github.com/hyperledger/fabric-sdk-go/blob/main/.codecov.yaml) in there repo.

Once this PR is merged we should extend it to also run our go tests with coverage test when `CODE_COVERAGE_ENABLED` is enabled.

mbrandenburger:
Also can you please add `*.gco` to `.gitignore` and ideally also some cleanup mechanism?. As I said in an offline discussion, this code coverage tool should also be used by FPC devs and not only CI. So if someone runs this locally by enabling it (`export CODE_COVERAGE_ENABLE=true`) we should make sure the coverage files are A) ignored by git and B) easy to remove.

build.mk
bvavala: I moved it there

ecc/chaincode/enclave/enclave.go
bvavala: If code coverage is enabled, then yes, because the libraries will contain gcov symbols, so the final executable must be linked accordingly.

If code coverage is disabled, then no, and in this case the compiler is intelligent enough not to add any gcov-related thing to the binary.

bvavala:
> Also can you please add `*.gco` to `.gitignore` and ideally also some cleanup mechanism?. As I said in an offline discussion, this code coverage tool should also be used by FPC devs and not only CI. So if someone runs this locally by enabling it (`export CODE_COVERAGE_ENABLE=true`) we should make sure the coverage files are A) ignored by git and B) easy to remove.

That's precisely what happens (at least "on my machine"). Any generated file falls in a `_build` folder, and your (A) and (B) points already apply for any such folder.

Also, I agree that developers might want to run this locally. And again, any new or existing artifact should be created in the `_build` folders. Do you have in mind any scenario showing otherwise, and which the FPC build should take into account?

bvavala:
> > Also can you please add `*.gco` to `.gitignore` and ideally also some cleanup mechanism?. As I said in an offline discussion, this code coverage tool should also be used by FPC devs and not only CI. So if someone runs this locally by enabling it (`export CODE_COVERAGE_ENABLE=true`) we should make sure the coverage files are A) ignored by git and B) easy to remove.
> 
> That's precisely what happens (at least "on my machine"). Any generated file falls in a `_build` folder, and your (A) and (B) points already apply for any such folder.
> 
> Also, I agree that developers might want to run this locally. And again, any new or existing artifact should be created in the `_build` folders. Do you have in mind any scenario showing otherwise, and which the FPC build should take into account?

Actually, for the local case, the script was missing a flag and support for the repo-token. I have added both.

mbrandenburger (approved):
Thanks Bruno! LGTM!

PR #560, opened 2021-04-07
Fix bug in preparation of write set

bvavala:
One of the nanopb data structures was not initialized properly during the preparation of the write-set at response time, ending up   in a memory corruption issue.
This PR adds a test which exposes the issue and provides a fix for that.


mbrandenburger (approved):
Tested it locally and and proposed fix seems to work well.

PR #565, opened 2021-04-15
Update Client SDK example with multi-org support

### LINKED ISSUE #562: Init enclaves for different organizations.
For example in integration/test-network, we got a network with two organizations org1 and org2.
And we need to initialize enclave for chaincode before calling transactions. 
And here https://github.com/hyperledger-labs/fabric-private-chaincode/blob/37de010921cf28195942078c561d0bc4c0c41570/client_sdk/go/sample/main.go#L168-L172 we initialize enclave for org1.
It works well when I calling transactions as a member in org1. But if I act as a member in org2, I can't call transactions.(And when I want to initialize  enclave for org2,  i got error messages that enclave already initialized)

But if I just initialize enclave for org2 at beginning, I can call transactions as a  member of org2 instead of org1.

So do we need to init enclaves for different organizations separately or do something to deal with this situation?

When I want to init enclave for org2. I got his message.
image
And here I use member in different orgs to call transactions.
User1 for org1. It worked will.
image
User1 for org2. Got error message.
image

mbrandenburger:
- Upgrade to Fabric SDK Go v1.0.0
- Refactor client_sdk/go/sample/main.go to easily switch between orgs via env vars
- Provide a sample connection file
- Update test-network tutorial to use the sample app with different orgs
- update-connection.sh now consolidates all peers that can host FPC enclaves

Fixes #562

```

### DIFF
--- a/client_sdk/go/sample/sample_connection.yaml
+++ b/client_sdk/go/sample/sample_connection.yaml
@@ -1,0 +1,110 @@
+name: test-network-org2
+version: 1.0.0
+client:
+  organization: Org2
+  connection:
+    timeout:
+      peer:
+        endorser: '300'
+organizations:
+  Org2:
+    mspid: Org2MSP
+    peers:
+      - peer0.org2.example.com
+    certificateAuthorities:
+      - ca.org2.example.com
+    cryptoPath: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/test-network/fabric-samples/test-network/organizations/peerOrganizations/org2.example.com/msp
+    users:
+      Admin:
+        cert:
+          path: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/test-network/fabric-samples/test-network/organizations/peerOrganizations/org2.example.com/users/Admin@org2.example.com/msp/signcerts/Admin@org2.example.com-cert.pem
+        key:
+          path: /project/src/github.com/hyperledger-labs/fabric-private-chaincode/integration/test-network/fabric-samples/test-network/organizations/peerOrganizations/org2.example.com/users/Admin@org2.example.com/msp/keystore/priv_sk
+peers:
+  peer0.org2.example.com:
+    url: grpcs://localhost:9051
+    tlsCACerts:
+      pem: |
+        -----BEGIN CERTIFICATE-----
+        MIICWDCCAf6gAwIBAgIRAJNPOYGJsy9KS5pXHfE3YqAwCgYIKoZIzj0EAwIwdjEL
+        MAkGA1UEBhMCVVMxEzARBgNVBAgTCkNhbGlmb3JuaWExFjAUBgNVBAcTDVNhbiBG
+        cmFuY2lzY28xGTAXBgNVBAoTEG9yZzIuZXhhbXBsZS5jb20xHzAdBgNVBAMTFnRs
+        c2NhLm9yZzIuZXhhbXBsZS5jb20wHhcNMjEwNDE0MDg1OTAwWhcNMzEwNDEyMDg1
+        OTAwWjB2MQswCQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UE
+        BxMNU2FuIEZyYW5jaXNjbzEZMBcGA1UEChMQb3JnMi5leGFtcGxlLmNvbTEfMB0G
+        A1UEAxMWdGxzY2Eub3JnMi5leGFtcGxlLmNvbTBZMBMGByqGSM49AgEGCCqGSM49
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by mbrandenburger** [comment]
@TrueAbc can you please have a look at this?

File: integration/test-network/README.md [hunk 1]
```diff
@@ -109,21 +110,22 @@ In order to initiate the FPC Chaincode enclave and register it with the FPC Encl
 
 ```bash
 cd ${FPC_PATH}/client_sdk/go/sample
-CC_ID=echo go run . -withLifecycleInitEnclave
+CC_ID=echo ORG_NAME=Org1 go run . -withLifecycleInitEnclave
 ```
+Note that we execute the go app as `Org1` and thereby issue to create and registery the FPC Chaincode enclave at `peer0.org1.example.com`.
```
bvavala, line 115:
```suggestion
Note that we execute the go app as `Org1`, thereby creating and registering the FPC Chaincode enclave at `peer0.org1.example.com`.
```

bvavala, line 115: Should we add that, alternatively, we could do the same at Org2?

integration/test-network/README.md
```diff
@@ -109,21 +110,22 @@ In order to initiate the FPC Chaincode enclave and register it with the FPC Encl
...
 
-Afterwards you can run the application without the flag
+Afterwards you can run the application without the `withLifecycleInitEnclave` flag and play with multiple organizations.
```
bvavala, line 117: Actually, if by chance somebody runs "with" initEnclave, it will fail. So, it is mandatory to use it once and only once.
```suggestion
Afterwards you _must_ run the application without the `withLifecycleInitEnclave` flag and you can play with multiple organizations.
```

client_sdk/go/sample/sample_connection.yaml
```diff
@@ -0,0 +1,110 @@
...
+        AwIDSAAwRQIgDybw8Bn6AAc5curctlFjhyOVX1zsrQUiBk3wgfZc5uYCIQCQH1nV
+        +cEVbS6SoV2jlB0K026S+kJWXzZOf8ThVyVMvA==
+        -----END CERTIFICATE-----
+    grpcOptions:
+      ssl-target-name-override: peer0.org2.example.com
+      hostnameOverride: peer0.org2.example.com
+  peer0.org1.example.com:
+    url: grpcs://localhost:7051
+    tlsCACerts:
+      pem: |
```
bvavala, line 49: I am a bit confused here.
The main reason for the connection profile that I see here is to specify the certificate of the peer endpoint. As we are using TLS, this will allow the handshake to succeed.
Two questions:
1. (I guess we quickly talked about this) in theory the client should know the currently valid root certificates in the channel and, additionally, the client should be able to extract the name/id (e.g., `peer0.org2.example.com`) from the certificate received by the server (i.e., `peer0`, following the previous case). Isn't there any such mechanism?
2. the `url` field seems to indicate a sort of DNS translation. Shouldn't this occur as the client tries to connect to (say) `peer0.org1.example.com`? Or perhaps `peer0.org1.example.com` is really a string and not a valid url?

mbrandenburger: The main reason to have the connection profile here is to provide an example, so the FPC Client SDK end-users can easily see what "minimum" configuration is required to use our SDK (Lifecycle and Gateway).

1) In theory this sounds correct. In practise you need to specify at least (that is my interpretation) one peer in the connection profile in order to connect to the network. With the help of that peer you (I mean the Fabric SDK) can then leverage discovery and get endpoints and TLS certs to connect to other peers. All the existing sample configuration files I saw for the Fabric SDK where using this form of configuration - I haven't see a configuration where the client just uses the org root cert to validate peer certs. Maybe there is such a mechanism but I haven't find it yet. @andrew-coleman WDYT?

2) In the Fabric SDK configuration, `peer0.org1.example.com:` is a key string that maps to an entry in the `peers:` section. In this example we just use `url: grpcs://localhost:9051` as I took the connection from the test-network, which uses docker-compose to run the network on your local machine.

bvavala (approved):

integration/test-network/README.md
```diff
@@ -92,50 +92,50 @@ The FPC Chaincode is now up and running, ready for processing invocations!
...
-For diagnostics, you can run the following to see logs for peer0.org1 ...
+For diagnostics, you can run the following to see logs for `peer0.org1.example.com`.
 ```bash
 docker logs -f peer0.org1.example.com
 docker logs -f ercc.peer0.org1.example.com
 docker logs -f ecc.peer0.org1.example.com
 ```
 
-To interact interactively with the peer, run the following
+To interact interactively with the peer using the `peer CLI`, run the following
```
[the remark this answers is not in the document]
mbrandenburger:
```suggestion
To interact with the peer using the `peer CLI`, run the following
```

mbrandenburger:
Squashed and rdy for merge

PR #569, opened 2021-04-20
Remove unused docker build dependencies (cc builder & peer)

bvavala:
This PR removes the cc_builder and peer from the docker build (but maintains the docker files as they might be useful in the future), thereby speeding up both local and CI build. 


mbrandenburger (approved):
LGTM!

PR #571, opened 2021-04-21
Remove some make dependencies

mbrandenburger:
- Move docker build from default `build` to `all` target
- Remove build-dep on test and linter to give the developer the freedom to just do the stuff they want (make all, or just make, still performs make build test)

Fixes #

```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
           DOCKER_QUIET_BUILD: 1
         run: |
             DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
-            make -C utils/docker build run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
+            make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
                                                                   DOCKER_QUIET_BUILD=1 \
                                                                   DOCKER_BUILDKIT=0 \
                                                                   DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
           DOCKER_QUIET_BUILD: 1
         run: |
             DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
-            make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
+            make -C utils/docker build run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
                                                                   DOCKER_QUIET_BUILD=1 \
                                                                   DOCKER_BUILDKIT=0 \
                                                                   DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ license:
 	@echo "License: Running licence checks.."
 	@scripts/check_license.sh
 
-linter: gotools build
+linter: gotools
 	@echo "LINT: Running code checks for Go files..."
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: .github/workflows/test.yml [hunk 1]
```diff
@@ -31,7 +31,7 @@ jobs:
           DOCKER_QUIET_BUILD: 1
         run: |
             DOCKER_BUILD_OPTS="--build-arg UBUNTU_VERSION=${{ matrix.os-version }} --build-arg UBUNTU_NAME=${{ matrix.os-name }}" \
-            make -C utils/docker run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
+            make -C utils/docker build run DOCKER_DEV_OPTIONAL_CMD='env IS_CI_RUNNING=true \
```
bvavala, line 34: What's the reason for this?
I think this changes the original build flow rationale in the CI, namely to do the whole build from docker.

Makefile
```diff
@@ -27,7 +27,7 @@ license:
 	@echo "License: Running licence checks.."
 	@scripts/check_license.sh
 
-linter: gotools build
+linter: gotools
```
bvavala, line 30: +1

build.mk
```diff
@@ -18,7 +18,6 @@ GO := $(GO_CMD) $(GOFLAGS)
 all: build test ci_report checks # keep checks last as license test is brittle ...
 
 .PHONY: test
-test: build
```
bvavala, line 21: Not sure this is going in the right direction. The dependency allows to rebuild after changes.
Actually, we should make sure that the dependencies work, namely: they force a rebuild when there are changes; they do not rebuild anything if there are no changes.

mbrandenburger: I agree for all cpp related subdirectories, this is the correct thing todo. However, go tests work differently as they do not require to run `go build` before `go test`. In particular, I removed this dependency here on the top level as all makefiles are including `build.mk` and thereby inherent this dependency. I suggest that we add this dependency where needed, eg. in `common/` and `ecc_enclave`

.github/workflows/test.yml
mbrandenburger: I added this here to also run the build for `ccenv` but let me integrate this better using the `IS_CI_RUNNING` flag

build.mk
mbrandenburger: I just checked `common/crypto` and `common/logging` already have a `build` dependency on `test`. `ecc_enclave` has currently no unit tests defined via cmake.

mbrandenburger: another reason why I don't like the dependency defined in `build.mk` is that a `make` triggered from the root dir will first go through all subdirs and call `make build`, once this is done, it goes through all subdirs again calling `make test`, which will call `build` dependency in each subdir.

We currently also have certain subfolders where `make build` is not smart enough (or our way how to use make) to detect that nothing has changed and just runs the build again. Proto generate is an example.

bvavala (changes requested):

build.mk
```diff
@@ -15,14 +15,17 @@ endif
...
-all: build test ci_report checks # keep checks last as license test is brittle ...
-
-.PHONY: test
-test: build
+# if we run CI, we add some additional targets on the default make
+ifeq (${IS_CI_RUNNING}, true)
+all: build test docker ci_report checks # keep checks last as license test is brittle ...
+else
+all: build test checks # keep checks last as license test is brittle ...
+endif
```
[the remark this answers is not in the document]
mbrandenburger: You are right (ii) is not ideal. Adding `docker` target to `all` should be the default. That is, by calling a simple `make` will do the entire build program, ... build, test, docker, (report if ci), and checks ...

As a developer I can just call`make build` and this should avoid calling the `docker target`. I don't like a `build-no-docker` target.

build.mk
bvavala: >another reason why I don't like the dependency defined in build.mk is that a make triggered from the root dir will first go through all subdirs and call make build, once this is done, it goes through all subdirs again calling make test, 
which will call build dependency in each subdir.

Per se, this is not an issue. Ideally, the second `make build`  would simply return with a "nothing to do".

> We currently also have certain subfolders where make build is not smart enough (or our way how to use make) to detect that nothing has changed and just runs the build again. Proto generate is an example.

Yes, I agree. Even in the CI, the `make test` causes (for example) `libshim` to be rebuilt. So, in general, there is some duplication of build -- even if you just run `make build` twice.

I am fine removing it for now but, again, it's `make build` that should be made smarter.
Regarding the go part, avoiding the go build/test dependency (assuming we restore the build/test dependency) adds unnecessary complexity since the go build part takes less than a handful of seconds, and in the end you always have to build for end-to-end tests.

bvavala (approved):
This is more "palatable".

In the future we may want to:
1) make `docker` a top-folder, since now its build is separate from `utils`' build
2) restore the `test` dependency on `build`. This however requires adding the dependencies on built artifacts in order to avoid unnecessary rebuilds.

PR #572, opened 2021-04-21
Make fabric patch optional

mbrandenburger:
- Update readme and remove statements telling the reader to patch fabric
- Remove fabric patch from default build path; patch targets are still available for custom use;
- Docker dev image still checkouts fabric repo
- By default fabric binaries are fetched into `fabric/_internal` and set in `config.mk`

Fixes #

```

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -78,6 +78,12 @@ export FPC_PATH=$(abspath $(TOP))
 # re-defined to point to the FPC path as seen by the docker daemon
 export DOCKERD_FPC_PATH ?= $(FPC_PATH)
 
+# For testing Fabric binaries are needed; you can customize these via the following
+# env variable. By default we fetch the binaries into $(FPC_PATH)/fabric/_internal/bin
+export FABRIC_BIN_DIR ?= $(FPC_PATH)/fabric/_internal/bin
+# In case you want to your custom fabric bins, just point to your fabric path
+# export FABRIC_BIN_DIR ?= $(FABRIC_PATH)/build/bin
+
 # Additional SGX related settings
 #--------------------------------------------------
 export SGX_CREDENTIALS_PATH ?= $(FPC_PATH)/config/ias
--- a/config.mk
+++ b/config.mk
@@ -78,11 +78,10 @@ export FPC_PATH=$(abspath $(TOP))
 # re-defined to point to the FPC path as seen by the docker daemon
 export DOCKERD_FPC_PATH ?= $(FPC_PATH)
 
-# For testing Fabric binaries are needed; you can customize these via the following
+# Fabric binaries are needed for testing; you can customize these via the following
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: config.mk [hunk 1]
```diff
@@ -78,6 +78,12 @@ export FPC_PATH=$(abspath $(TOP))
 # re-defined to point to the FPC path as seen by the docker daemon
 export DOCKERD_FPC_PATH ?= $(FPC_PATH)
 
+# For testing Fabric binaries are needed; you can customize these via the following
+# env variable. By default we fetch the binaries into $(FPC_PATH)/fabric/_internal/bin
+export FABRIC_BIN_DIR ?= $(FPC_PATH)/fabric/_internal/bin
+# In case you want to your custom fabric bins, just point to your fabric path
+# export FABRIC_BIN_DIR ?= $(FABRIC_PATH)/build/bin
```
bvavala, line 85: Leftover?

fabric/Makefile
```diff
@@ -11,8 +11,24 @@ FABRIC_PEER_BIN = $(FABRIC_PATH)/build/bin/peer
 FABRIC_PATCHED = $(FABRIC_PATH)/.fpc_patched
 FABRIC_CURRENT_DIFF_COMMITS_CMD = git log $(FABRIC_VERSION_TAG)..HEAD --oneline
 
-build: native
+FABS = _internal
```
bvavala, line 14: I guess here we should move `bin` to `scripts`, and `_internal` to `bin`.

config.mk
mbrandenburger: this was on purpose but I rephrased it now

fabric/Makefile
mbrandenburger: mh, I was thinking the same. But then decided that for me it makes more sense when we provide our peer wrapper in `fabric/bin` as this is something we give to the FPC users.

It seems that current 'scripts' only contain helper for testing and ci. That said, it would make sense for me to move `ledger_*` scripts to `scripts`. The other wrapper files (configtxgen and orderer) are also not really needed. 

In testing, can't we just define leverage the PATH to run the peer, orderer, and configtxgen commands?

We could say if FPC is used: do something like `PATH=$FPC_PATH/fabric/bin:$FPC_PATH/fabric/_internal/bin:$FPC`

fabric/README.md
```diff
@@ -12,6 +12,16 @@ cd $FPC_PATH/fabric
 make
 ```
 
+# Build custom Fabric binaries
+
```
bvavala, line 16:
```suggestion
The following assumes that the `FABRIC_PATH` environment variable contains the path to Fabric's source code.
```

bvavala (changes requested):
minor useful changes

bvavala (approved):

PR #584, opened 2021-05-18
Add hybrid request encryption

mbrandenburger:
This introduces a hybrid encryption scheme for fpc requests. Before FPC
requests were encrypted using the chaincode encryption key using
RSA-based encryption. This had the limitation that the maximum payload
size was restricted to ~200bytes. They new scheme uses symmetric
encryption to encrypt the actual request and uses the chaincode
encryption key to encrypt the key used to encrypt the FPC request. The
encrypted request key is added to the request protos. Unit test for go
cryto package and support for mock_ecc were added.

Fixes #

```

### DIFF
--- a/common/enclave/cc_data.h
+++ b/common/enclave/cc_data.h
@@ -45,9 +45,8 @@ public:
     ByteArray get_state_encryption_key();
     std::string get_enclave_id();
     bool sign_message(const ByteArray& message, ByteArray& signature) const;
-    bool decrypt_cc_message(const ByteArray& encrypted_message, ByteArray& message) const;
-    bool encrypt_message(
-        const ByteArray key, const ByteArray& message, ByteArray& encrypted_message) const;
+    bool decrypt_key_transport_message(
+        const ByteArray& encrypted_key_transport_message, ByteArray& key_transport_message) const;
 };
 
 extern cc_data* g_cc_data;
--- a/common/enclave/cc_data.h
+++ b/common/enclave/cc_data.h
@@ -42,11 +42,13 @@ public:
         uint32_t credentials_max_size,
         uint32_t* credentials_size);
 
-    ByteArray get_state_encryption_key();
     std::string get_enclave_id();
     bool sign_message(const ByteArray& message, ByteArray& signature) const;
     bool decrypt_key_transport_message(
         const ByteArray& encrypted_key_transport_message, ByteArray& key_transport_message) const;
+    bool decrypt_state_value(const ByteArray& encrypted_value, ByteArray& value) const;
+    bool encrypt_state_value(const ByteArray& value, ByteArray& encrypted_value) const;
+    static int estimate_encrypted_state_value_length(const int value_len);
 };
 
 extern cc_data* g_cc_data;
--- a/ecc/chaincode/enclave/mock_enclave.go
+++ b/ecc/chaincode/enclave/mock_enclave.go
@@ -23,6 +23,7 @@ import (
 	"github.com/hyperledger/fabric-private-chaincode/internal/protos"
 	"github.com/hyperledger/fabric-protos-go/ledger/rwset/kvrwset"
 	"github.com/hyperledger/fabric/protoutil"
+	"github.com/pkg/errors"
 )
 
 type MockEnclaveStub struct {
@@ -107,7 +108,7 @@ func (m *MockEnclaveStub) GetEnclaveId() (string, error) {
 	return strings.ToUpper(hex.EncodeToString(hash[:])), nil
 }
 
-func (m *MockEnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, chaincodeRequestMessage []byte) ([]byte, error) {
+func (m *MockEnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, chaincodeRequestMessageBytes []byte) ([]byte, error) {
 	logger.Debug("ChaincodeInvoke")
 
 	signedProposal, err := stub.GetSignedProposal()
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

**Comment by mbrandenburger** [comment]
@g2flyer do you have a chance to look at this as I touched mainly the code you added. Thank you

**Comment by mbrandenburger** [comment]
something is wired with github actions :/

File: ecc/chaincode/enclave/mock_enclave.go [hunk 1]
```diff
@@ -122,15 +123,26 @@ func (m *MockEnclaveStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, chai
 		return nil, err
 	}
 
+	encryptedRequestEncryptionKey := chaincodeRequestMessageProto.GetEncryptedRequestEncryptionKey()
+	if encryptedRequestEncryptionKey == nil {
+		return nil, fmt.Errorf("no encrypted request")
```
mbrandenburger, line 128:
```suggestion
		return nil, fmt.Errorf("no encrypted request key")
```

bvavala:
This PR implements the second solution suggested in #478 , namely: _RSA-OAEP with random key and AES-256-GCM_

.gitignore
```diff
@@ -8,3 +8,5 @@ cscope.*
 # generate protos
 common/protos
 internal/protos
+coverage.out
```
bvavala, line 11: You had the double star in the other branch.

g2flyer (changes requested):

g2flyer (changes requested):
General idea of solving issue 478 certainly the right way (adding it to PDO crypto as initially thought would be the more generic solution but as long we need (large) symmetric messages exactly and only for the cc request/response, then doing it here is almost cleaner ...).

That said, see my concern regarding having two RSA encryptions: merging them into a single key-transport RSA encryption not only is more secure but also conceptually cleaner (e.g., symmetric encrypted messages are really only the payloads) and should also lead to more symmetric APIs in cc_data.* ...

internal/crypto/client_encryption.go
```diff
@@ -121,13 +52,14 @@ func (e EncryptionProviderImpl) NewEncryptionContext() (EncryptionContext, error
...
-		resultEncryptionKey:    resultEncryptionKey,
+		requestEncryptionKey:   requestEncryptionKey,
+		responseEncryptionKey:  resultEncryptionKey,
 		chaincodeEncryptionKey: ccEncryptionKey,
 	}, nil
 }
 
 // EncryptionContext defines the interface of an object responsible to encrypt the contents of a transaction invocation
-// and decrypt the corresponding response.
+// and DecryptMessage the corresponding response.
```
mbrandenburger, line 62:
```suggestion
// and decrypt the corresponding response.
```

protos/fpc/fpc.proto
```diff
@@ -99,8 +99,11 @@ message CleartextChaincodeRequest {
...
 message ChaincodeRequestMessage {
-    // an RSA-encryption of the serialization of CleartextChaincodeRequest with the chaincode encryption key
+    // an encryption (using the request encryption key (aes128-gcm)) of the serialization of CleartextChaincodeRequest with the chaincode encryption key
     bytes encrypted_request = 1;
+
+    // an RSA-encryption of the request encryption key (aes128-gcm) with the chaincode encryption key
+    bytes encrypted_request_encryption_key = 2;
 }
 
 message CleartextChaincodeResponse {
```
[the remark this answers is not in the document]
mbrandenburger: I like the idea of having the `KeyTransportMessage` for key encapsulation. Will push a fixup.

common/enclave/cc_data.h
```diff
@@ -45,7 +45,9 @@ class cc_data
     ByteArray get_state_encryption_key();
     std::string get_enclave_id();
     bool sign_message(const ByteArray& message, ByteArray& signature) const;
-    bool decrypt_cc_message(const ByteArray& encrypted_message, ByteArray& message) const;
+    bool decrypt_cc_message(const ByteArray& encrypted_message_key,
+        const ByteArray& encrypted_message,
+        ByteArray& message) const;
     bool encrypt_message(
         const ByteArray key, const ByteArray& message, ByteArray& encrypted_message) const;
```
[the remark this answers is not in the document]
mbrandenburger: good question, that question should have been asked when this method was initially written :P

With introducing a `KeyTransportMessage`, `cc_data.h` should only need a `decrypt_key_transport_message` as this requires the use of `cc_data.cc_decryption_key_`, which is private in `cc_data`. This also makes decrypt/encrypt_cc_message obsolete as they are only wrapper around `pdo::crypto::skenc::EncryptMessag` and `pdo::crypto::skenc::DecryptMessag`.

protos/fpc/fpc.proto
```diff
@@ -93,14 +93,22 @@ message InitEnclaveMessage {
...
-    bytes return_encryption_key = 3;
 }
 
 message ChaincodeRequestMessage {
-    // an RSA-encryption of the serialization of CleartextChaincodeRequest with the chaincode encryption key
+    // an encryption (symmetric) of the serialization of CleartextChaincodeRequest with KeyTransportMessage.request_encryption_key
     bytes encrypted_request = 1;
+
+    // an encryption (asymmetric) of the serialization of request KeyTransportMessage with AttestedData.chaincode_ek
+    bytes encrypted_key_transport_message = 2;
```
g2flyer, line 103: would seem more natural to revert the order (but obviously a matter of personal taste ...)

protos/fpc/fpc.proto
```diff
@@ -93,14 +93,22 @@ message InitEnclaveMessage {
...
+}
+
+message KeyTransportMessage {
+    // key to decrypt CleartextChaincodeRequest
+    bytes request_encryption_key = 1;
+
+    // key to encrypt CleartextChaincodeResponse
```
g2flyer, line 110: worth mentioning in comment what type of key it is? (AES-GCM-128?)

protos/fpc.options
```diff
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
-fpc.CleartextChaincodeRequest.return_encryption_key type:FT_POINTER
-
 fpc.ChaincodeRequestMessage.encrypted_request type:FT_POINTER
+fpc.ChaincodeRequestMessage.encrypted_key_transport_message type:FT_POINTER
+
+fpc.KeyTransportMessage.request_encryption_key type:FT_POINTER
+fpc.KeyTransportMessage.response_encryption_key type:FT_POINTER
```
g2flyer, line 7: These should in principle be fixed size?

g2flyer (approved):
Looks good to me (but haven't done any testing or alike ...)

protos/fpc/fpc.proto
mbrandenburger: I removed the details about RSA and AES keys as they may change with whatever crypto lib is used (even though I don't see this changed soon).

protos/fpc.options
mbrandenburger: True, but keeping this variable allows to decide the crypto lib the size of the keys.

common/enclave/cc_data.h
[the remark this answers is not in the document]
bvavala: > btw: why is the encrypt_message function not called encrypt_cc_message for symmetry reason?
> good question, that question should have been asked when this method was initially written :P

The answer is (was!) in the implementation: the former is for generic encryption, the latter uses the cc key.

ecc_enclave/enclave/enclave.cpp
```diff
@@ -115,10 +145,11 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
...
 
         {  // encrypt response
-            b = g_cc_data->encrypt_message(return_encryption_key,
-                ByteArray(b64_response.c_str(), b64_response.c_str() + b64_response.length()),
-                encrypted_response);
-            COND2LOGERR(!b, "cannot encrypt message");
+            ByteArray response =
+                ByteArray(b64_response.c_str(), b64_response.c_str() + b64_response.length());
+            CATCH(b, encrypted_response =
+                         pdo::crypto::skenc::EncryptMessage(response_encryption_key, response));
```
bvavala, line 151: The decryption and the encryption operations were in `cc_data` 
* to keep them in a single place, together with the crypto keys
* and to confine the pdo crypto usage to `cc_data`

This breaks the abstraction, and the pdo libs are accessible only because they had to be exposed in `cc_data.h`, otherwise they would not be available.
This part should be reverted.

mbrandenburger: I am not yet convinced about reverting this part.

I like the abstraction that all the enclave keys are not exposed outside `cc_data`. However, this abstraction is already broken in `shim.cpp` where the `state_encryption_key` is used with the pdo crypto functions. Also `enclave.cpp` used `pdo::crypto::constants::SYM_KEY_LEN` before.

The methods to decrypt and encrypt the request and the response messages using the keys from the `keyTransportMessage` are, I believe, not related to the enclave state (cc_data), that is, they are not consuming any state of `cc_data`, and thus, for me `cc_data` is the wrong place to provide such an abstraction for the pdo crypto.

In contrast, to limit the scope of the chaincode_decryption_key used to decrypt KeyTransportMessage it is OK for me to have `decrypt_key_transport_message` as part of `cc_data`. Simply because this function depends on the state.

A clean solution, I believe, would be to have a `crypto.cpp/h` in `common/crypto` which offers all primitives/constants needed in FPC, so we can remove all references to `pdo::crypto` elsewhere. I also think this is beyond the scope of this PR.

mbrandenburger: Oh we already have `pdo-crypto-c-wrapper.h/cpp`. Can't we just use this?

bvavala: > Oh we already have pdo-crypto-c-wrapper.h/cpp. Can't we just use this?

I don't see why we can't, but in my opinion we should not.
As I said above, this wrapper additionally makes a translation between object-oriented and procedural abstractions.
This comes with some overhead at the edge due to data serialization/deserialization and object build/destruction.

mbrandenburger: ok, I moved the static (generic) methods for message encryption/decryption to `enclave/crypto.cpp` and extended cc_data with methods to encrypt/decrypt state values without returning the state_encryption_key to `shim.cpp`. That is, `cc_data` not provides all crypto functions that are related to it's state. `cc_data` still uses pdo crypto directly. Stateless, crypto functions such as message encryption/decryption/hashing are not available in `crypto.cpp` and can be used wherever needed.

Eventually, the dependency to pdo crypto can be removed also from `cc_data` by wrapping pdo methods inside `crypto.cpp`. I think we have now a cleaner solution and did a bit more of refactoring as needed to solve the actual problem - but this is also not a bad thing. Let's aim to finish this PR and conclude this discussion now as it is growing.

bvavala (approved):

PR #604, opened 2021-06-16
Create pure go crypto impl

### LINKED ISSUE #601: Go client SDK: crypto library in plain Go
Currently, the Go client SDK requires a signature scheme whose implementation is not in pure go. 
This complicates the integration of client SDK into the Fabric Smart Client (https://github.com/hyperledger-labs/fabric-smart-client/issues/23).

Therefore, the purpose of this request is to add support for all required crypto schemes in plain Go.

mbrandenburger:
~This is WIP~. 
- This PR relies on https://github.com/hyperledger-labs/private-data-objects/pull/323 ✅
- Attestation conversion tested with HW mode and IAS ✅

Changes:

- updates PDO to the latest version, including RSA key size bump to 3072bit and flexible elliptic curve parameters
- Introduces a crypto provider service (csp) which comes with a PDO-based
  crypto implementation and a pure go implementation.
- Existing crypto test suite is extended with test cases using both csp
  implmentations
- CSP impl compatibility test ensures both implementation are working
  smoothly together
- FPC client SDK uses go-based CSP by default
- Introduce a attestation converter that supports different attestation types.
- Add support for simulated and epid attestation
- Use new attestation converter in FPC Client SDK and peer-cli-assist by default
- Create attestation conversion compatibility test with PDO-based conversion

Fixes #601 and #478

```

### CODE REVIEW

**Comment by mbrandenburger** [comment]
Build breaks due to missing PDO crypto patch.

**Comment by mbrandenburger** [comment]
Note when testing manually make sure that `common/crypto/pdo` HEAD is at latest commit ` 7546768`. Otherwise testing may fail with `unsupported eliptic curve` ...

**Review by bvavala — COMMENTED** [review]
(partial review, still in progress)

File: internal/attestation/test_cmd/main.go [hunk 1]
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// this tool is meant to be used in `$FPC_PATH/common/crypto/attestation-api/test` to ensure compatibility
+// with the attestation conversion implementation based on PDO crypto.
```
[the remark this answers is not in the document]
mbrandenburger: updated comment

internal/attestation/test_cmd/main.go
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// this tool is meant to be used in `$FPC_PATH/common/crypto/attestation-api/test` to ensure compatibility
```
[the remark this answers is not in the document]
mbrandenburger: ok, moved it

internal/attestation/epid.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "could not read %s", path)
+	}
+
+	if len(data) == 0 {
+		return "", errors.Errorf("empty file %s", path)
+	}
+
+	return strings.TrimSuffix(string(data), "\n"), nil
+}
```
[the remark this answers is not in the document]
mbrandenburger: I would like to keep this implementation in the internal folder as it seems to be the right place from a "go" repository perspective as it is consumed by the FPC client SDK and the cli helper. 
I also agree that the conversion implementation is related to the shell-based implementation and therefore it also makes sense for me to have both implementation close. I see our common folder a good place for common c++ code, not so much for go code. I think this is one of our general problems having a multi-language repository. In the future, we may split our repo into multiple. If there was a FPC sdk repo, the code would be already in the right place.

internal/crypto/crypto.go
```diff
@@ -7,278 +7,19 @@ SPDX-License-Identifier: Apache-2.0
...
-		encryptedMessageLen,
-		&encryptedMessageActualLen,
-	)
-	if ret == false {
-		return nil, fmt.Errorf("encryption failed")
-	}
-
-	return C.GoBytes(encryptedMessagePtr, C.int(encryptedMessageActualLen)), nil
+// CSP (Crypto Service Provider) offers a high-level abstract of cryptographic primitives used in FPC
+type CSP interface {
```
bvavala, line 11: Nice refactoring here!

scripts/golinter.sh
```diff
@@ -62,8 +62,9 @@ if [ -n "$OUTPUT" ]; then
 fi
 
 echo "Checking with go vet"
+GOTAGS=WITH_PDO_CRYPTO
 PRINTFUNCS="Debug,Debugf,Print,Printf,Info,Infof,Warning,Warningf,Error,Errorf,Critical,Criticalf,Sprint,Sprintf,Log,Logf,Panic,Panicf,Fatal,Fatalf,Notice,Noticef,Wrap,Wrapf,WithMessage"
-OUTPUT="$(go vet -all -printfuncs "$PRINTFUNCS" $1/...)"
+OUTPUT="$(go vet -all -tags "$GOTAGS" -printfuncs "$PRINTFUNCS" $1/...)"
```
bvavala, line 67: This uses the specified tag everywhere, doesn't it?

Also, I think it works only because the code vetted with the build tag is a super set of the code vetted without the build tag.
So, if anybody implements a Go file with `// +build !WITH_PDO_CRYPTO`, this is left out.

Should we maybe duplicate the `go vet` (i.e., run w/ and w/o gotags)?

bvavala (commented):
(second and last part)

scripts/golinter.sh
mbrandenburger: good point! yes, here we tell go vet to use `WITH_PDO_CRYTPO` tag everywhere.

Without the tag, all the files with `// +build WITH_PDO_CRYPTO` are excluded from vetting. This also causes go vet to throw an error for ERCC, since the implementation for [NewVerifier](https://github.com/hyperledger/fabric-private-chaincode/blob/8edde2f06061dfa5088aab0a7b7665708029c05d/ercc/attestation/verifier.go#L23) is missing. As ERCC is currently relying on PDO crypto, it must be built with the `WITH_PDO_CRYPTO` tag nevertheless. Having a ERCC without PDO crypto would be an alternative, but requires to write the verification code in go as well. Maybe a TODO for the future to also give SDK users the ability to verify the attestations. Technically, within ERCC there is no technical need to use the `WITH_PDO_CRYPTO` tag, in contrast to the Client SDK related components, however, I think this tag is a good indicator that this files should only be compiled if PDO crypto is available on the platform.

Ideally we could tell go vet to use the `WITH_PDO_CRYPTO` tag only in the scope of ERCC. Not sure how to do that easily. But if we do that, the PDO-based CSP implementation is also left out from vetting. This becomes complex very quickly. I suggest we see the `WITH_PDO_CRYPTO` tag not as a toggle. That is, you are not supposed to write go files using `!WITH_PDO_CRYPTO` in order to use a different implementation for something.

Alternatively, we could run go vet per component to better respect individual build options. This may require a bigger refactoring of the `golinter.sh`. I don't think it is worth the efforts at the moment, and it does not harm to run go vet just with the `WITH_PDO_CRYPTO` flag. Do you have any idea for a lightweight solution?

It seems that gotags and platform specific code in general add this complexity when using go vet. See [here](https://stackoverflow.com/questions/63745232/how-to-run-go-vet-on-all-code-regardless-of-platform).

bvavala: > Alternatively, we could run go vet per component to better respect individual build options. This may require a bigger refactoring of the golinter.sh. I don't think it is worth the efforts at the moment, and it does not harm to run go vet just with the WITH_PDO_CRYPTO flag. Do you have any idea for a lightweight solution?

Right, the right approach seem to be per-component. Not worth it now.
I though the lightweight approach was the "double" execution, but unfortunately it does not work.
It's ok as it is for now.

bvavala (approved):
Nice!

PR #607, opened 2021-07-12
Improve test-network sample to accept any chaincode

### LINKED ISSUE #605: Improve test-network sample to accept any chaincode
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

The current test-network sample only accepts the echo chaincode. It could explain how to also use the helloworld chaincode or any other chaincode. To be able to do so the user would have to be able to configure the chaincode name and the path to the code itself.
**Describe alternatives you've considered**

The solution could be done in 2 steps. 
1. Standardize the use of CC_ID because in some parts TEST_CC_ID is used, add CC_PATH as configurable, update the Readme to reflect this change. 
2. Update just the readmes of the test-network and the helloworld samples to reflect the steps needed to use the helloworld chaincode in the test-network.
**Additional context**

munapower:

Solves step one in issue 605. 

It changes the Test-Network sample readme
```

### DIFF
--- a/samples/deployment/test-network/Makefile
+++ b/samples/deployment/test-network/Makefile
@@ -7,10 +7,10 @@ TOP = ../../..
 include $(TOP)/build.mk
 
 ifeq ($(CC_ID),)
- $(error CC_ID is not set);
+ $(error CC_ID is not set)
 endif
 ifeq ($(CC_PATH),)
- $(error CC_PATH is not set);
+ $(error CC_PATH is not set)
 endif
 
 build: ercc-container ecc-container
--- a/samples/deployment/test-network/Makefile
+++ b/samples/deployment/test-network/Makefile
@@ -7,10 +7,10 @@ TOP = ../../..
 include $(TOP)/build.mk
 
 ifeq ($(CC_ID),)
- $(error CC_ID is not set) && exit 1;
+ $(error CC_ID is not set);
 endif
 ifeq ($(CC_PATH),)
- $(error CC_PATH is not set) && exit 1;
+ $(error CC_PATH is not set);
 endif
 
 build: ercc-container ecc-container
--- a/samples/deployment/test-network/Makefile
+++ b/samples/deployment/test-network/Makefile
@@ -6,9 +6,12 @@
 TOP = ../../..
 include $(TOP)/build.mk
 
-
-CC_ID ?= echo
-CC_PATH ?= ${FPC_PATH}/samples/chaincode/${CC_ID}
+ifeq ($(CC_ID),)
+ $(error CC_ID is not set) && exit 1;
+endif
+ifeq ($(CC_PATH),)
+ $(error CC_PATH is not set) && exit 1;
+endif
 
 build: ercc-container ecc-container
 
--- a/samples/deployment/test-network/Makefile
+++ b/samples/deployment/test-network/Makefile
@@ -7,8 +7,8 @@ TOP = ../../..
 include $(TOP)/build.mk
 
 
-TEST_CC_ID ?= echo
-TEST_CC_PATH ?= ${FPC_PATH}/samples/chaincode/${TEST_CC_ID}
+CC_ID ?= echo
+CC_PATH ?= ${FPC_PATH}/samples/chaincode/${CC_ID}
 
 build: ercc-container ecc-container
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by munapower** [comment]
@ikegawa-koshi this would be the first stage, if you would like to review.
TIA!

File: samples/deployment/test-network/Makefile [hunk 1]
```diff
@@ -7,8 +7,8 @@ TOP = ../../..
 include $(TOP)/build.mk
 
 
-TEST_CC_ID ?= echo
-TEST_CC_PATH ?= ${FPC_PATH}/samples/chaincode/${TEST_CC_ID}
+CC_ID ?= echo
+CC_PATH ?= ${FPC_PATH}/samples/chaincode/${CC_ID}
```
mbrandenburger, line 11: It would be nice to check if `CC_ID` and `CC_PATH` are set; otherwise abort make, similar to what we are doing in `installFPC.sh`.

samples/deployment/test-network/installFPC.sh
```diff
@@ -15,16 +15,24 @@ if [[ -z "${FPC_PATH}"  ]]; then
...
 
 PEERS=("peer0.org1.example.com" "peer0.org2.example.com")
 
 ERCC_EP="OutOf(2, 'Org1MSP.peer', 'Org2MSP.peer')"
 ECC_EP="OutOf(2, 'Org1MSP.peer', 'Org2MSP.peer')"
 
-CC_ID="echo"
-CC_VER="$(cat ${FPC_PATH}/samples/chaincode/${CC_ID}/_build/lib/mrenclave)"
+#CC_ID?=echo
+#CC_PATH?="$(cat ${FPC_PATH}/samples/chaincode/${CC_ID}"
```
mbrandenburger, line 34: let remove this instead of commenting it out

samples/deployment/test-network/README.md
```diff
@@ -79,11 +79,12 @@ cd $FPC_PATH/samples/deployment/test-network
...
 
 Now we have the FPC Chaincode installed on the Fabric peers, but we still need to start our chaincode containers.
-Make sure you have set `TEST_CC_ID` to the same chaincode ID as used in the earlier step when building the chaincode.
+Make sure you have set `CC_ID` to the same chaincode ID as used in the earlier step when building the chaincode. Also confirm that `CC_PATH` is set to the location of the FPC chaincode code.
 
 ```bash
 # Start FPC container
-export TEST_CC_ID=echo
+export CC_ID=echo
+export CC_PATH=${FPC_PATH}/samples/chaincode/echo
```
mbrandenburger, line 87: maybe the explicit export instruction in the beginning of this readme is enough and we can save those two lines here. The ext above also tells the user to check that `CC_ID` and `CC_PATH` is set accordingly.

mbrandenburger (commented):
This looks already good. Have not tested it yet, will do once this PR moves from draft to review status.

ikegawa-koshi:
@munapower
I think so too; these changes are very good!

samples/deployment/test-network/installFPC.sh
munapower: hahah I guessed you would say that

samples/deployment/test-network/README.md
munapower: Yeah, I also thought of that ...

samples/deployment/test-network/Makefile
munapower: Done!

samples/deployment/test-network/installFPC.sh
munapower: Done!

samples/deployment/test-network/README.md
munapower: Done!

samples/deployment/test-network/Makefile
munapower: @mbrandenburger I ended up doing it a bit differently than in other parts but it works. Please can you review.

samples/deployment/test-network/Makefile
```diff
@@ -6,9 +6,12 @@
...
 
-
-TEST_CC_ID ?= echo
-TEST_CC_PATH ?= ${FPC_PATH}/samples/chaincode/${TEST_CC_ID}
+ifeq ($(CC_ID),)
+ $(error CC_ID is not set) && exit 1;
+endif
+ifeq ($(CC_PATH),)
+ $(error CC_PATH is not set) && exit 1;
+endif
```
mbrandenburger, line 14:
```suggestion
ifeq ($(CC_ID),)
	$(error CC_ID is not set)
endif
ifeq ($(CC_PATH),)
 	$(error CC_PATH is not set)
endif
```

mbrandenburger: it seems that the `exit 1` is not needed as the error command already exits make according to https://www.gnu.org/software/make/manual/html_node/Make-Control-Functions.html

mbrandenburger (changes requested):
This looks good and works for me. Just a minor change request and then this is ready to merge.

samples/deployment/test-network/Makefile
munapower: You are right. Made the change and tested.

PR #618, opened 2021-08-21
Unittest for ECC

mbrandenburger:
Adding missing unit tests for ECC. This PR also includes some refactoring for better testing; fixed typos, go docs, etc.


bvavala (approved):
* lgtm
* tested in HW mode
* the integration tests worked without Fabric's source code -- hence it does not reintroduce the dependency

PR #635, opened 2021-10-22
Demo FSC integration using the IRB demo

mbrandenburger:
This PR adds a new sample on how to use FPC with the Fabric Smart Client (FSC) by demonstrating our IRB demo, which was presented at HLGF21. To simplify the demo and focus on the FPC and FSC usage, the demo was stripped a lot. Web services, UI, and patient data pdf transformer have been removed and experiment worker uses a pure python implementation (and thus maybe breaking graphene compatibility).


bvavala (commented):
looks good - (not tested)

PR #639, opened 2021-11-12
Fix sgx device location for dcap machines

bvavala:


mbrandenburger (approved):
Thanks Bruno! LGTM

PR #640, opened 2021-12-07
Add del_state support

mbrandenburger:
This PR adds `del_state` to the FPC shim.

### DIFF
--- a/ecc_enclave/enclave/shim.cpp
+++ b/ecc_enclave/enclave/shim.cpp
@@ -170,6 +170,7 @@ void put_public_state(const char* key, uint8_t* val, uint32_t val_len, shim_ctx_
     // save write -- only the last one for the same key
     ctx->write_set.erase(key);
     ctx->write_set.insert({key, ByteArray(val, val + val_len)});
+    ctx->del_set.erase(key);
 
     ocall_put_state(key, val, val_len, ctx->u_shim_ctx);
 }
@@ -248,6 +249,15 @@ void get_public_state_by_partial_composite_key(
     }
 
     unmarshal_values(values, (const char*)json, len);
+}
+
+void del_state(const char* key, shim_ctx_ptr_t ctx)
+{
+    ctx->write_set.erase(key);
+    ctx->del_set.insert(key);
+
+    ocall_del_state(key, ctx->u_shim_ctx);
+    return;
 }
 
 int get_string_args(std::vector<std::string>& argss, shim_ctx_ptr_t ctx)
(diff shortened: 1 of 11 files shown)


mbrandenburger:
DO NOT merge until issue in #641 is discovered

mbrandenburger:
Need to check what is the problem here. Will continue after holidays :/

integration/client_sdk/go/stress_test/stress_test.go
```diff
@@ -73,15 +73,15 @@ var _ = Describe("Stress tests", func() {
...
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(res).Should(Equal([]byte("OK")))
 
-					res, err = echoContract.EvaluateTransaction("get_state", key)
+					res, err = contract.EvaluateTransaction("get_state", key)
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(res).Should(Equal([]byte(value)))
 
-					_, err = echoContract.EvaluateTransaction("del_state", key)
+					res, err = contract.SubmitTransaction("del_state", key)
```
mbrandenburger, line 84: This fixes the failing integration test.

ecc_enclave/enclave/shim.cpp
```diff
@@ -250,6 +251,15 @@ void get_public_state_by_partial_composite_key(
     unmarshal_values(values, (const char*)json, len);
 }
 
+void del_state(const char* key, shim_ctx_ptr_t ctx)
+{
+    ctx->write_set.erase(key);
+    ctx->del_set.insert(key);
+
+    ocall_del_state(key, ctx->u_shim_ctx);
```
bvavala, line 259: (this is unrelated to the PR)

Why do we do this ocall? -- we do the same for the `put_state`.

The ocall for reading is necessary to grab the data. However, the ocalls for putting and deleting keys essentially update the transaction's rwset in Fabric, which we are not using (if I recall correctly). In fact, FPC maintains its own rwset, which is validated and replayed in the second step (of the two-step transaction commit procedure).

There may still be one reason for these ocall: checking for errors, and returning any error to the enclave.
The big problem is that any feedback/error would not be trustworthy. This is likely a showstopper for such error handling.

bvavala (approved):
looks good - not tested locally

ecc_enclave/enclave/shim.cpp
mbrandenburger: very good observation. We should look into that in ideally remove the ocall for `put_state` and `del_state` if really not needed. This may positively chaincode execution performance.

PR #641, opened 2021-12-07
Add stress test for many invocations and fix memory leak

mbrandenburger:
Add another stress test to detect memory leaks.

### DIFF
--- a/ecc_enclave/enclave/crypto.cpp
+++ b/ecc_enclave/enclave/crypto.cpp
@@ -16,8 +16,6 @@
 #include "pdo/common/crypto/crypto.h"
 #include "sgx_trts.h"
 
-typedef std::unique_ptr<BIO, void (*)(BIO*)> BIO_ptr;
-
 int get_random_bytes(uint8_t* buffer, size_t length)
 {
     /* WARNING WARNING WARNING */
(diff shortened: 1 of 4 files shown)


mbrandenburger:
For some reason my test fails but did not fail the entire CI test ... 🤔

mbrandenburger:
I suspect the changes made in #637 causing the enclave crash. Running the test without a949ca9 seems to pass.

mbrandenburger:
The stress test runs by default with 1000 invocations (`defaultInvocationCount = 1000`).
Via `export STRESS_TEST_INVOCATION_COUNT=10000` you can set the invocation count when testing locally with larger numbers.

mbrandenburger:
> EDITED: I suspect the changes made in #637 causing the enclave crash (earlier). ~Running the test without [a949ca9](https://github.com/hyperledger/fabric-private-chaincode/commit/a949ca90c7e38b9a3de6c7391db88773cc0a17a0) seems to pass.~

As mentioned earlier, the changes in #637 introduces a memory leak when decoding the creator certificate. Additionally, some `pb_release` calls were missing. Note that also some `pb_release` calls where missing, in particular those two:
```
    pb_release(fpc_CleartextChaincodeRequest_fields, &cleartext_cc_request);
    pb_release(fpc_KeyTransportMessage_fields, &key_transport_message);
```

mbrandenburger:
This PR not contains a fix for the memory leaks that causes the FPC enclave to crash after many invocations. @bvavala please have a look. Once this PR is ready and merged, we can look into PR #640.

mbrandenburger:
Just squashed fixups so this PR is ready for merge.

ecc_enclave/enclave/crypto.cpp
```diff
@@ -111,37 +109,46 @@ std::string extract_encoded_public_key(const ByteArray cert_pem)
     EVP_PKEY* pubkey = X509_get_pubkey(cert);
     EC_KEY* eckey = EVP_PKEY_get1_EC_KEY(pubkey);
 
-    BIO_ptr pkBio(BIO_new(BIO_s_mem()), BIO_free_all);
-    PEM_write_bio_EC_PUBKEY(pkBio.get(), eckey);
-    int keylen = BIO_pending(pkBio.get());
+    BIO* pkBio = BIO_new(BIO_s_mem());
```
bvavala, line 112: The return value is not checked. Should doublecheck whether it fails properly later in case of error.

ecc_enclave/enclave/crypto.cpp
```diff
@@ -111,37 +109,46 @@ std::string extract_encoded_public_key(const ByteArray cert_pem)
...
+    BIO_free_all(certBio);
+
     return std::string((const char*)pem_str.data(), pem_str.size());
 }
 
 std::string extract_subject_from_cert(const ByteArray cert_pem)
 {
-    BIO_ptr certBio(BIO_new(BIO_s_mem()), BIO_free_all);
-    BIO_write(certBio.get(), cert_pem.data(), cert_pem.size());
+    BIO* certBio = BIO_new(BIO_s_mem());
```
bvavala, line 134: The return value is not checked. Should doublecheck whether it fails properly later in case of error.

ecc_enclave/enclave/crypto.cpp
```diff
@@ -111,37 +109,46 @@ std::string extract_encoded_public_key(const ByteArray cert_pem)
...
+    BIO_write(certBio, cert_pem.data(), cert_pem.size());
 
-    X509* cert = PEM_read_bio_X509(certBio.get(), NULL, NULL, NULL);
+    X509* cert = PEM_read_bio_X509(certBio, NULL, NULL, NULL);
     if (!cert)
     {
         throw std::runtime_error("cannot parse cert");
     }
 
     char* subj = X509_NAME_oneline(X509_get_subject_name(cert), NULL, 0);
```
bvavala, line 143: The return value is not checked. Should doublecheck whether it fails properly later in case of error.

ecc_enclave/enclave/enclave.cpp
```diff
@@ -370,6 +377,8 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
 
     // release dynamic allocations (TODO:release in case of error)
     pb_release(fpc_ChaincodeRequestMessage_fields, &cc_request_message);
+    pb_release(fpc_CleartextChaincodeRequest_fields, &cleartext_cc_request);
```
bvavala, line 380: Good catch!

ecc_enclave/enclave/crypto.cpp
mbrandenburger: added

ecc_enclave/enclave/crypto.cpp
mbrandenburger: added

ecc_enclave/enclave/crypto.cpp
mbrandenburger: added

PR #657, opened 2022-02-25
fabric-private-chaincode/samples/deployment/test-network/Readme.md bugs

b3n3d17:
Fixes the test-network README so it works again ;)

see [issue](https://github.com/hyperledger/fabric-private-chaincode/issues/656)

Thanks to  Marcus and Koshi for helping me solve my issues.

I do not expect that to happen.

However, please whoever knows how the simple-go client works. Check out why it is breaking with the test-network.
I was not able to figure that out.


samples/deployment/test-network/README.md
```diff
@@ -130,27 +131,9 @@ cd $FPC_PATH/samples/deployment/test-network
...
-```
-Note that we execute the go app as `Org1`, thereby creating and registering the FPC Chaincode enclave at `peer0.org1.example.com`. Alternatively, we could run this as `Org2` to initiate the enclave at `peer0.org2.example.com`.
-
-Afterwards you _must_ run the application without the `withLifecycleInitEnclave` flag and you can play with multiple organizations.
-```bash
-cd $FPC_PATH/samples/application/simple-go
-CC_ID=echo ORG_NAME=Org1 go run .
-CC_ID=echo ORG_NAME=Org2 go run .
-```
-
```
mbrandenburger, line 150: I believe in order to fix this tutorial issue, we need to touch `$FPC_PATH/samples/application/simple-go` as well. 

I would prefer not removing this, and create a separate PR.

samples/deployment/test-network/README.md
```diff
@@ -130,27 +131,9 @@ cd $FPC_PATH/samples/deployment/test-network
...
 ### How to use simple-cli-go
 
-You can also use `$FPC_PATH/samples/application/simple-cli-go` instead of the simple-go application.
+You can use `$FPC_PATH/samples/application/simple-cli-go` to interact with the echo chaincode.
```
mbrandenburger, line 136:
```suggestion
You can use `$FPC_PATH/samples/application/simple-cli-go` to interact with the chaincode.
```

mbrandenburger (changes requested):
Thanks for this PR! I suggest that we keep the "use simple-go" section and fix it in another PR. WDYT?

samples/deployment/test-network/README.md
mbrandenburger: I've created PR #660 to fix this part of the tutorial.

b3n3d17: sure I'll remove that commit

b3n3d17: I also squashed the commits.
It should be ready.

mbrandenburger (approved):
Thanks! LGTM now!

Issue #656, opened 2022-02-25
fabric-private-chaincode/samples/deployment/test-network/Readme.md bugs


b3n3d17:
**Describe the bug**
* git clone https://github.com/hyperledger/fabric-samples
    * The current fabric-samples repo changed its structure and therefore breaks the Readme.md
* The way simple-go was described breaks
* An envrionmentvariable was mission to make simple-cli-go work

**To Reproduce**
Try and do the Readme.md with the exact steps described.
I used SGX_MODE=HW and used the containerized dev environemnt

**Expected behavior**
* folder strucutres are different in fabric-samples and docker-compose file networkname is different
* simple-go cannot find the wallet path (Org1 should be lower case in the command org1)
   * If that issue is fixed this error appears upon using simple-go commands "key transport message decryption failed"(cc_data.cpp:270)  
 
**Log-filesScreenshots**
not needed

**Environment (please complete the following information):**
 - Ubuntu 20.04
 - https://github.com/hyperledger/fabric-private-chaincode/commit/a0fe00b27df083a55ecaf23c84abe48d53a8a8b4)
 - Azure Confidential Computing vm: size=DCsv2

PR #660, opened 2022-03-02
Fix configuration path for case-sensitive fs

mbrandenburger:
The client SDK needs to load the configuration file from the file
system. In this sample app we use the configurations provided in
fabric-samples/test-network. On a case-sensitive fs (Linux), the
constructed path (ccpPath) in this demo app is not correct. We fix
this by demo by using strings.ToLower as we know the path already.

The problem is described in #656. This PR fixes the problem with simple-go.

### DIFF
--- a/samples/application/simple-go/main.go
+++ b/samples/application/simple-go/main.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/hyperledger/fabric-private-chaincode/client_sdk/go/pkg/client/resmgmt"
 	fpc "github.com/hyperledger/fabric-private-chaincode/client_sdk/go/pkg/gateway"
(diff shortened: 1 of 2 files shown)


mbrandenburger:
@b3n3d17 could you please look into this and see if that fixes the problem you described on Discord:
```
The only remaining problem that still exists is that using simple-go does still not work:
First the command:
CC_ID=echo ORG_NAME=Org1 go run . -withLifecycleInitEnclave
does not work because ORG_NAME is capitalized -> it therefore does not find the wallet
Error:
 [sdk-test] main -> FATA 001 failed to create sdk: failed to initialize configuration: unable to load config backend: loading config file failed: /project/src/github.com/hyperledger/fabric-private-chaincode/samples/deployment/test-network/fabric-samples/test-network/organizations/peerOrganizations/Org1.example.com/connection-Org1.yaml: open /project/src/github.com/hyperledger/fabric-private-chaincode/samples/deployment/test-network/fabric-samples/test-network/organizations/peerOrganizations/Org1.example.com/connection-Org1.yaml: no such file or directory
-------------------------------------------------------------------------------------------------------------------------------------------------------
Changing the ORG_NAME to org1 works until:

CC_ID=echo ORG_NAME=org1 go run . -withLifecycleInitEnclave

[sdk-test] main -> INFO 001 --> Invoke LifecycleInitEnclave
[sdk-test] main -> FATA 002 Failed to get network: Failed to create new channel client: event service creation failed: could not get chConfig cache reference: QueryBlockConfig failed: QueryBlockConfig failed: target(s) required
exit status 1
-------------------------------------------------------------------------------------------------------------------------------------------------------
Same happens with the other commands : 
CC_ID=echo ORG_NAME=org1 go run . 
and 
CC_ID=echo ORG_NAME=org12go run .
```

bvavala (approved):

mbrandenburger:
Let's wait for @b3n3d17 to confirm.

b3n3d17:
can confirm in 2h

mbrandenburger:
> **One thing is missing** the file [README.md](https://github.com/hyperledger/fabric-private-chaincode/blob/main/samples/deployment/test-network/README.md) needs the following addition to line 140 image
> 
> ```shell
> export SGX_CREDENTIALS_PATH=$FPC_PATH/config/ias
> ```

Good catch! Added it! Thanks

PR #669, opened 2022-04-11
Fix get_*_partial_state

mbrandenburger:

When we call a `get_*_partial_state`, we fetch the data outside the enclave and wrap the entire result in a json message with the format [`key=value,...]`. Inside the enclave, we translate this json to `std::map<string, string>` and return it to the chaincode.

With non-public data, the values are encrypted and base64 encoded. That is, the json to map translation works as expected. However, for public data, where we handle arbitrary data, it seems that a value which contains a json string, will corrupt the json message and cannot be transformed into the map.

To fix this problem we do the following:
When creating the json message outside the enclave, we just take the value and encode it with base64 no matter if it is encrypted or not. After the translation into `std::map`, we just decode the value again. In case of encrypted values, we encode/decode twice and thereby increase payload size.

Thanks @lebdron for spotting this bug! The full discussion about this error can be found on [discord](https://discord.com/channels/905194001349627914/943059293765267497/962989270329602078).

### DIFF
--- a/ecc/chaincode/enclave/shim.go
+++ b/ecc/chaincode/enclave/shim.go
@@ -12,6 +12,7 @@ package enclave
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"sync"
 	"unsafe"
@@ -162,7 +163,7 @@ func get_state_by_partial_composite_key(comp_key *C.char, values *C.uint8_t, max
 		buf.WriteString("{\"key\":\"")
 		buf.WriteString(utils.TransformToFPCKey(item.Key))
 		buf.WriteString("\",\"value\":\"")
-		buf.Write(item.Value)
+		buf.WriteString(base64.StdEncoding.EncodeToString(item.Value))
 		if iter.HasNext() {
 			buf.WriteString("\"},")
 		} else {
--- a/ecc_enclave/enclave/shim.cpp
+++ b/ecc_enclave/enclave/shim.cpp
@@ -192,7 +192,8 @@ int unmarshal_values(
     {
         JSON_Object* pair = json_array_get_object(pairs, i);
         const char* key = json_object_get_string(pair, "key");
-        const char* value = json_object_get_string(pair, "value");
+        const char* b64value = json_object_get_string(pair, "value");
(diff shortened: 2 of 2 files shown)


lebdron:
Works great, thank you @mbrandenburger !

bvavala (commented):
Good catch!
The problem is most likely due to the quotes. It makes sense to wrap a string, but if that's a json blob, then that tricks parson.

The b64 encoding definitely solves the issue, but it add lots of overhead and payload size.

Instead, escaping the value (on write, and reverting that when getting the string) should be sufficient. Also, this would only impact the public values, since there's nothing to escape in the b64 strings.

mbrandenburger:
> The b64 encoding definitely solves the issue, but it adds lots of overhead and payload size.
> 
> Instead, escaping the value (on write, and reverting that when getting the string) should be sufficient. Also, this would only impact the public values, since there's nothing to escape in the b64 strings.

Note that we need this encoding (or escaping) when we get the data via `get_*_partial_state` calls. The data will be stored on the ledger in the original format as submitted by the chaincode. We only need to handle the case when we collect the data (in `ecc`), pack it in json array with key/value pairs and give it to the enclave (`ecc_enclave`).

While I like the idea of escaping to keep the overhead moderate, I think this approach makes it just more complex. That is, do we need to detect if the value is a string that needs to be escaped? Do we already have this escaping reverting code available in `ecc_enclave`? @bvavala Do you want to give it a try and provide a patch using escaping?

bvavala:
> Good catch! The problem is most likely due to the quotes. It makes sense to wrap a string, but if that's a json blob, then that tricks parson.
> 
> The b64 encoding definitely solves the issue, but it adds lots of overhead and payload size.
> 
> Instead, escaping the value (on write, and reverting that when getting the string) should be sufficient. Also, this would only impact the public values, since there's nothing to escape in the b64 strings.

We just confirmed that the handled values are actually not of type string but rather bytes.
Hence, escaping is not a good option, and base64 is recommended.

In addition, there is a type mismatch between the getstate/putstate API, which work with bytes, and the getpartialstate which returns a map of strings.

PR #672, opened 2022-05-19
Fixing irb experiment Make file and removing tf dependence

berserkr:

Fixes #

```


bvavala:
Hi @berserkr , thanks for the PR.
Would you please elaborate on
* the "fix" to the Makefile
* why the dependency should be removed?

Also, if a removal is indeed necessary, then the lines should actually be removed, rather than commented.

berserkr:
> Hi @berserkr , thanks for the PR. Would you please elaborate on

Sure, np :)

> 
> * the "fix" to the Makefile

The Makefile does not have a `build` tag, so the container is not being built. The Make file has a `docker` tag instead. The fix is to rename the tag to `build`.

> * why the dependency should be removed?
> 


Yes, so TF is not actually used through the demo. This is a PyTorch demo with no calls to TF other than the `__version__` property.  I was getting TF failures on my end, for some reason it was complaining about TF GPU not being loaded properly. So, although it could be an environment issue, after debugging I realized that the TF dependency is not really needed..

> Also, if a removal is indeed necessary, then the lines should actually be removed, rather than commented.

Should've removed it, old habit to make sure I discuss it before removal :)

mbrandenburger:
@berserkr Thanks for creating this PR! Let me run this locally. I have the feeling that a few more changes are needed to make the demo working (again) :)

mbrandenburger:
@berserkr I was able to reproduce the issue. As already mentioned, there are a few more changes needed to make the IRB demo run smoothly again. I've pushed yet another commit to this PR to add all the changes. Please have a look and check if this demo now runs without any problems.

bvavala (approved):

PR #674, opened 2022-06-22
README_branchname_fix_peer_name_fix

berserkr:
This change covers two items:

1) Branch name in README does not match the expected branch.
2) Description of running peers does not match what is in the system. Name has been change to reflect that.


mbrandenburger (approved):
Thanks for catching this!

PR #678, opened 2022-07-04
Fix README.md for make docker

ikegawa-koshi:

When building FPC, We need to enter the fpc-development-main container and run `make`.
In that case, We need to run `make docker` before running `make`, otherwise We will get an error, so I modified the README.

None

No

Signed-off-by: ikegawa-koshi <koshi.ikegawa.mf@hitachi.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,11 +363,13 @@ Note by default we build FPC with SGX simulation mode. For SGX hardware-mode sup
 To build all required FPC components and run the integration tests run the following:
 ```bash
 cd $FPC_PATH
+make docker
 make
  ```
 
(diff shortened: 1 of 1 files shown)


mbrandenburger:
> When building FPC, We need to enter the fpc-development-main container and run `make`.

I think this is not an accurate problem description. The `fpc-ccenv` image is actually missing.

ikegawa-koshi:
Indeed, this problem description is indeed incorrect, since it is not about entering a docker container. I will fix it.

ikegawa-koshi:
Thanks for your advice.
I have modified the description.

bvavala (approved):

PR #683, opened 2022-08-01
Fix path to SGX hardware

ikegawa-koshi:

An error occurred when performing a FPC build in the go-support-preview branch on a DCsv2 instance of Microsoft Azure.
I fixed it because the way the path to the SGX hardware in `integration/go_chaincode/utils/utils.go` was written was not an absolute path.

None

None

Signed-off-by: ikegawa-koshi <koshi.ikegawa.mf@hitachi.com>


mbrandenburger (approved):
Good catch! Thanks

PR #687, opened 2022-08-16
New fabric-samples demo

mbrandenburger:
This PR adds a new sample that shows how to port one of the existing fabric-samples chaincode to FPC.
To enable existing go chaincodes, support for `peer.Response` proto messages was added in FPC go and c/c++ chaincode.


internal/utils/proto.go
```diff
@@ -232,3 +232,18 @@ func GetChaincodeRequestMessageFromSignedProposal(signedProposal *pb.SignedPropo
...
+
+// UnwrapResponse unmarshalls the given serialized peer.Response message and returns the Payload field if Status is 200;
+// otherwise, the Message field is returned as an error
+func UnwrapResponse(responseBytes []byte) (payload []byte, err error) {
+	clearResponse, err := protoutil.UnmarshalResponse(responseBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal peer.Response message")
+	}
+
+	if clearResponse.Status != 200 {
```
bvavala, line 244: I'm wondering if the status should be checked and removed here, or if this could be propagated further.

ecc_enclave/enclave/enclave.cpp
```diff
@@ -232,25 +233,60 @@ int ecall_cc_invoke(const uint8_t* signed_proposal_proto_bytes,
...
+            ret = memcpy_s(protoResponse.payload->bytes, protoResponse.payload->size, response,
+                response_len_out);
+            COND2LOGERR(ret != 0, "cannot encode field");
+        }
+        else
+        {
+            // set error
+            protoResponse.status = 500;
+            // TODO set response.message with proper error message as reported by user
+        }
```
bvavala, line 259: I think it's conceptually cleaner -- one b64 encoding less.

bvavala (approved):
lgtm

PR #740, opened 2023-10-20
Remove loading docker config inside dev container

### LINKED ISSUE #715: Docker commands are not found inside the development environment
**Describe the bug**

Once Docker environment is updated to 4.21.x you can start the development environment but if you try to run the `make build` command you get the following message `Cannot connect to the Docker daemon at unix:///Users/XXX/.docker/run/docker.sock. Is the docker daemon running?`

**To Reproduce**

1- Start dev environment with `make -C utils/docker build build-dev run-dev`
2- from inside the container run `make build` and at one point you will get the following:
```
FABRIC_IMAGES: peer orderer ccenv tools baseos
===> Pulling fabric Images
====>  docker.io/hyperledger/fabric-peer:2.5.3
Cannot connect to the Docker daemon at unix:///Users/XXX.docker/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///Users/XXX/.docker/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///Users/XXX/.docker/run/docker.sock. Is the docker daemon running?
```

**Expected behavior**

User should be able to run the make build and the container find the docker commands.
**Log-filesScreenshots**

- Mac OS 13.5
- FPC: main github branch
- Docker for Desktop: 4.21.1
**Additional context**

I had added the docker_host in the config override but it didn't use it
I saw it listed with docker context
```
docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                     KUBERNETES ENDPOINT   ORCHESTRATOR
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux *     moby                Docker Desktop                            unix:///Users/XXX/.docker/run/docker.sock
```
UNTIL I ran `docker context use default` and voila, the `make clean `inside the container did a lot more than it had previously
and the `make build` worked!

mbrandenburger:
With recent versions of Docker Desktop (and others Rancher Desktop), the docker configuration loaded into the dev container can not resolve the path to the docker socket. We let users the option to mount the docker configuration inside the container if needed.

This PR fixes #715

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ A few more notes:
   DOCKER_BUILD_OPTS=--build-arg UBUNTU_VERSION=18.04 --build-arg UBUNTU_NAME=bionic
   ```
 * If you run behind a proxy, you will have to configure the proxy,
-  e.g., for docker (`~/.docker/config.json`). See
+  e.g., for docker (`~/.docker/config.json`) and load the configuration inside the dev container by setting `DOCKER_DEV_RUN_OPTS += -v "~/.docker":"/root/.docker"` in `$FPC_PATH/config.override.mk`. See
   [Working from behind a proxy](#working-from-behind-a-proxy) below for more information.
 * If your local host is SGX enabled, i.e., there is a device `/dev/sgx/enclave` or
   `/dev/isgx` and your PSW daemon listens to `/var/run/aesmd`, then the docker image will be sgx-enabled and your settings from `./config/ias` will be used. You will have to manually set `SGX_MODE=HW` before building anything to use HW mode.
--- a/README.md
+++ b/README.md
@@ -226,8 +226,8 @@ A few more notes:
   DOCKER_BUILD_OPTS=--build-arg UBUNTU_VERSION=18.04 --build-arg UBUNTU_NAME=bionic
   ```
 * If you run behind a proxy, you will have to configure the proxy,
-  e.g., for docker (`~/.docker/config.json`) and load the configuration inside the dev container by setting `DOCKER_DEV_RUN_OPTS += -v "~/.docker":"/root/.docker"` in `$FPC_PATH/config.override.mk`. See
-  [Working from behind a proxy](#working-from-behind-a-proxy) below for more information.
+  e.g., for docker (`~/.docker/config.json`) and load the configuration inside the dev container by setting `DOCKER_DEV_RUN_OPTS += -v "$HOME/.docker":"/root/.docker"` in `$FPC_PATH/config.override.mk`.
+  See [Working from behind a proxy](#working-from-behind-a-proxy) below for more information. Also note that with newer docker versions (i.e., docker desktop), the docker socket is located on the host in `~/.docker/`. This may cause issues when using docker inside the FPC dev container as the docker client is not able to access the docker socket at the path of the host system. You may try to switch the docker context to use `/var/run/docker.sock`. We do not recommend this approach and happy for suggestions.
 * If your local host is SGX enabled, i.e., there is a device `/dev/sgx/enclave` or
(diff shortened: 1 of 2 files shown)


mbrandenburger:
@munapower This should resolve Issue #715! Can you please test it.

README.md
```diff
@@ -226,7 +226,7 @@ A few more notes:
   DOCKER_BUILD_OPTS=--build-arg UBUNTU_VERSION=18.04 --build-arg UBUNTU_NAME=bionic
   ```
 * If you run behind a proxy, you will have to configure the proxy,
-  e.g., for docker (`~/.docker/config.json`). See
+  e.g., for docker (`~/.docker/config.json`) and load the configuration inside the dev container by setting `DOCKER_DEV_RUN_OPTS += -v "~/.docker":"/root/.docker"` in `$FPC_PATH/config.override.mk`. See
```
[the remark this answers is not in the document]
munapower: @mbrandenburger have you seen the comments above?

README.md
```diff
@@ -226,7 +226,7 @@ A few more notes:
...
   [Working from behind a proxy](#working-from-behind-a-proxy) below for more information.
```
munapower, line 230: please review previous comments

munapower (changes requested):

README.md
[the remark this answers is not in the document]
mbrandenburger: ah we should propose to use `"$HOME/.docker"` instead of '~'.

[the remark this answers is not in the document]
mbrandenburger: anyway, the second error message is expected if you don't switch the docker context. 

Maybe we need to update the wording of the readme saying that this might introduce additional complications?!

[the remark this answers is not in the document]
munapower: let me try with `$HOME/.docker`

README.md
munapower: Let me re phrase my question. Is my configuration considered to qualify as "behind a proxy"?

mbrandenburger: Well, I don't know - do you access the internet via a proxy?

mbrandenburger: I would the the "normal" use case is no proxy ...

munapower: right, so I don't have a proxy defined in my browser settings so I consider that my computer isn't "behind a proxy" so I shouldn't have applied the configurations?

munapower: and since the issue is because of docker versions and not the fact that you are behind or not a proxy, should the comment be there or should it be another note?

README.md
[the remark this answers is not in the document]
munapower: @mbrandenburger I had to use DOCKER_DEV_RUN_OPTS += -v $$HOME/.docker:/root/.docker for it to work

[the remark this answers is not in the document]
munapower: so I also tried without having this configuration set and everything worked, including running `make docker` once inside the container

README.md
mbrandenburger: well, with this PR we won't have the issue anymore. It will be only a problem if someone loads the host docker configuration into the dev container. The only reason why someone would do it is because they need extra configuration for instance of proxy use.

README.md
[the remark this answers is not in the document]
munapower: So, finally tested without this configuration, and after having changed docker desktop configuration to use the legacy implementation for the file sharing I was able to follow thu with the readme.
If I use this configuration, once I reach the Build my first chaincode and try to execute `make docker` I get the error that is solved changing the docker context used.

README.md
munapower: view comment in other thread.

munapower (approved):

PR #747, opened 2023-11-28
Add echo-go sample

mbrandenburger:
This PR adds another go chaincode sample based on the existing c++ echo chaincode


munapower (approved):
Tested and it works

PR #748, opened 2023-12-05
Fix testing on non-Intel CPU

bvavala:
This PR upgrades the SGX SDK and SSL packages and enables the SKIP_INTELCPU_CHECK flag.
SGX SSL aborts when it's not used on Intel CPUs, even in SIM mode.
However, more recent versions (which require an SDK upgrade) allow to build the package by excluding that check.
Additionally, the PR enables SGX SSL tests to spot these issues early on, since these are only executed when the docker containers need to be rebuilt.


utils/docker/base-rt/Dockerfile
```diff
@@ -30,7 +30,7 @@ ARG UBUNTU_VERSION
 ARG UBUNTU_NAME
 
 # config/build params (part 2)
-ARG SGX=2.12
+ARG SGX=2.16
```
mbrandenburger, line 33: Is this version the most recent one we can get without "compatibility" issues / api changes?

PR #759, opened 2024-04-29
Add Dependabot

mbrandenburger:
Helps us to keep track up of updates.


mbrandenburger:
Before this one can be merged we need to find a way to keep `github.com/hyperledger/fabric-samples/asset-transfer-basic/chaincode-go` within `go.mod` while dependabot runs `go mod tidy`

munapower (approved):

PR #766, opened 2024-06-11
Feature/solution skvs

### LINKED ISSUE #484: Rollback protection Extension (aka Trusted Ledger)
**Is your feature request related to a problem? Please describe.**

Extension of FPC Lite / 1.0 to provide rollback protection ...
See [milestone 44](https://github.com/hyperledger-labs/fabric-private-chaincode/milestone/44) for container containing related issues. See also non-fpc-lite part of the specification in `docs/design/fabricv2+` for related protocols

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

chenchanglew:

Implement a Rollback attack protection solution for FPC: SKVS.

Single Key-Value Storage (SKVS) is a naive approach for rollback attacks. All key-value pairs are encapsulated and stored in this approach with a single call to put_state(). During execution, the enclave must load the entire state before accessing individual key-value pairs. While this approach prevents the rollback attack, applications with large states and multiple writers will experience bad performance, as the use of a single key-value pair will cause transactions to fail due to concurrent write issues.

A user can use it by changing the chain code to SVKS chaincode
ex: `skvsChaincode := fpc.NewPrivateChaincode(secretChaincode, fpc.WithSKVS())`

for example we can test with skvs/main.go in secret keeper 
`make ECC_MAIN_FILES=cmd/skvs/main.go with_go env docker`

Fixes #484 

Loom demonstration video: [Watch here](https://www.loom.com/share/4790cd0cba9e4433879083b41158d42d?sid=a0f367e7-dc49-4065-926d-df7b2fbf55a1)


ecc_go/chaincode/enclave_go/skvs_stub.go
```diff
@@ -0,0 +1,111 @@
...
+type skvsStub struct {
+	*EnclaveStub
+}
+
+func NewSkvsStub(cc shim.Chaincode) *skvsStub {
+	enclaveStub := NewEnclaveStub(cc)
+	return &skvsStub{enclaveStub}
+}
+
+func (e *skvsStub) ChaincodeInvoke(stub shim.ChaincodeStubInterface, chaincodeRequestMessageBytes []byte) ([]byte, error) {
```
mbrandenburger, line 29: I have the feeling that we should remove this wrapper and instead just inject a provider function for the stub that we can set externally.

chenchanglew: no sure about this one, might need to ask more insight

munapower:
> @munapower can you also please have look at this PR. Thanks!

@mbrandenburger @chenchanglew not sure what I am supposed to test. @chenchanglew did you resolve the conflict Marcus mentions? What am I supposed to test? Should I follow the Readme or what?

ecc_go/chaincode/singleKVS.go
```diff
@@ -0,0 +1,29 @@
...
+func NewSkvsChaincode(cc shim.Chaincode) *chaincode.EnclaveChaincode {
+	logger.Info("Creating new SKVS Chaincode")
+	ecc := &chaincode.EnclaveChaincode{
+		Enclave:   enclave_go.NewSkvsStub(cc),
+		Validator: endorsement.NewValidator(),
+		Extractor: &chaincode.ExtractorImpl{},
+		Ercc:      &ercc.StubImpl{},
+	}
+	return ecc
+}
```
mbrandenburger, line 29: I am wondering if it would be better to add an option to the `NewPrivateChaincode` method to instantiate the with SKVS rather than creating this new constructor.

It could look like ... 
```go
func NewPrivateChaincode(cc shim.Chaincode, options ...func(*chaincode.EnclaveChaincode)) *chaincode.EnclaveChaincode {
  ecc := &chaincode.EnclaveChaincode{
		Enclave:   enclave_go.NewSkvsStub(cc),
		Validator: endorsement.NewValidator(),
		Extractor: &chaincode.ExtractorImpl{},
		Ercc:      &ercc.StubImpl{},
	}
  for _, o := range options {
    o(ecc)
  }
  return ecc
}

func WithSKVS() func(*chaincode.EnclaveChaincode) {
  return func(ecc *chaincode.EnclaveChaincode) {
    ecc.Enclave = enclave_go.NewSkvsStub(cc)
  }
}
```

and in the chaincode `main.go`, the developer would write something like that ...
```go
// naive
chaincode := fpc.NewPrivateChaincode(secretChaincode)
// with SKVS
chaincode := fpc.NewPrivateChaincode(secretChaincode, fpc.WithSKVS)
```

See this article https://golang.cafe/blog/golang-functional-options-pattern.html

WDYT?

mbrandenburger (changes requested):

ecc_go/chaincode/private.go
```diff
@@ -14,14 +14,24 @@ import (
...
 )
 
+type BuildOption func(*chaincode.EnclaveChaincode, shim.Chaincode)
+
 // NewPrivateChaincode creates a new chaincode! This is for go support only!!!
-func NewPrivateChaincode(cc shim.Chaincode) *chaincode.EnclaveChaincode {
+func NewPrivateChaincode(cc shim.Chaincode, options ...BuildOption) *chaincode.EnclaveChaincode {
 	ecc := &chaincode.EnclaveChaincode{
-		Enclave:   enclave_go.NewEnclaveStub(cc),
+		Enclave:   enclave_go.NewSkvsStub(cc),
```
mbrandenburger, line 22: I believe here we want to keep the "normal" NewEnclaveStub

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -62,22 +62,18 @@ func (s *SkvsStubInterface) InitSKVS() error {
 		}
 	}
 
-	logger.Warningf("SKVS Init finish, allDataOld: %s, allDataNew: %s", s.allDataOld, s.allDataNew)
```
mbrandenburger: I would suggest that we remove these loggers as they may print sensitive data.

ecc_go/chaincode/enclave_go/skvs_stub.go
```diff
@@ -0,0 +1,21 @@
...
+
+package enclave_go
+
+import (
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+)
+
+func NewSkvsStub(cc shim.Chaincode) *EnclaveStub {
+	logger.Warning("==== SKVS NewSkvsStub ====")
```
mbrandenburger, line 15:
```suggestion
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+
+type SkvsStubInterface struct {
+	*FpcStubInterface
+	allDataOld map[string][]byte
+	allDataNew map[string][]byte
+	key        string
+}
+
+func NewSkvsStubInterface(stub shim.ChaincodeStubInterface, input *pb.ChaincodeInput, rwset *readWriteSet, sep StateEncryptionFunctions) *SkvsStubInterface {
+	logger.Warning("==== Get New Skvs Interface =====")
```
mbrandenburger, line 25:
```suggestion
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+	fpcStub := NewFpcStubInterface(stub, input, rwset, sep)
+	skvsStub := SkvsStubInterface{fpcStub, map[string][]byte{}, map[string][]byte{}, "SKVS"}
+	err := skvsStub.InitSKVS()
+	if err != nil {
+		logger.Warningf("Error!! Initializing SKVS failed")
```
mbrandenburger, line 30: If we have an error ... shouldn't we better through a panic? We cannot really go on and continue, right?

mbrandenburger, line 30:
```suggestion
		panic("Initializing SKVS failed")
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return &skvsStub
+}
+
+func (s *SkvsStubInterface) InitSKVS() error {
+	logger.Warningf(" === Initializing SKVS === ")
```
mbrandenburger, line 36:
```suggestion
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// get current state, this will only operate once
+	encValue, err := s.GetPublicState(s.key)
+	if err != nil {
+		return nil
+	}
+
+	if len(encValue) == 0 {
+		logger.Warningf("SKVS is empty, Initiating.")
```
mbrandenburger, line 45: should we also handle this error differently? Can we actually continue if `encValue` is empty?

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		logger.Warningf("Error!! Initializing SKVS failed")
+	}
+	return &skvsStub
+}
+
+func (s *SkvsStubInterface) InitSKVS() error {
+	logger.Warningf(" === Initializing SKVS === ")
+
+	// get current state, this will only operate once
```
mbrandenburger, line 38: Do you think it makes sense to protect the entire Init method with `sync.Once`? 
https://pkg.go.dev/sync#Once

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if len(encValue) == 0 {
+		logger.Warningf("SKVS is empty, Initiating.")
+	} else {
+		value, err := s.sep.DecryptState(encValue)
+		if err != nil {
+			return err
+		}
+		logger.Warningf("SKVS has default value, loading current value.")
```
mbrandenburger, line 51:
```suggestion
		logger.Debug("SKVS has default value, loading current value.")
```

mbrandenburger (changes requested):
Thanks Zac for the refactoring. I've found a few things we should still fix. Please have a look.

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
chenchanglew: yes, we can continue if the encValue is empty.

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
chenchanglew: I think this function is safe even it runs multiple time, since the 
`encValue, err := s.GetPublicState(s.key)` will always fetch the latest value, if someone try to do something malicious by calling this init function they will just get the latest public state.
Plus I already change the init function into a private function thus I think it should be okay without the sync.Once.

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,116 @@
...
+	allDataNew map[string][]byte
+	key        string
+}
+
+func NewSkvsStubInterface(stub shim.ChaincodeStubInterface, input *pb.ChaincodeInput, rwset *readWriteSet, sep StateEncryptionFunctions) *SkvsStubInterface {
+	fpcStub := NewFpcStubInterface(stub, input, rwset, sep)
+	skvsStub := &SkvsStubInterface{
+		FpcStubInterface: fpcStub,
+		allDataOld:       map[string][]byte{},
+		allDataNew:       map[string][]byte{},
```
mbrandenburger, line 31:
```suggestion
		allDataOld:       make(map[string][]byte),
		allDataNew:       make(map[string][]byte),
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.Errorf("SKVS Json unmarshal error: %s", err)
+		return err
+	}
+	return nil
+}
+
+func (s *SkvsStubInterface) GetState(key string) ([]byte, error) {
+	value, found := s.allDataOld[key]
+	if !found {
+		return nil, errors.New("skvs allDataOld key not found")
```
mbrandenburger, line 77:
```suggestion
		return nil, fmt.Errorf("skvs allDataOld key=%s not found", key)
```

mbrandenburger: Mhh, shouldn't we return `nil, nil` in that case?
```
	// If the key does not exist in the state database, (nil, nil) is returned.
	GetState(key string) ([]byte, error)
```
https://github.com/hyperledger/fabric-chaincode-go/blob/main/shim/interfaces.go#L80C1-L81C38

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
```diff
@@ -0,0 +1,116 @@
...
+*/
+
+package enclave_go
+
+import (
+	"encoding/json"
+
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/pkg/errors"
```
mbrandenburger, line 14:
```suggestion
```

ecc_go/chaincode/enclave_go/skvs_stub_interface.go
mbrandenburger, line 77:
```suggestion
		return nil, nil
```

PR #772, opened 2024-09-05
Fix IRB container package

mbrandenburger:
This PR re-enables the IBR tests and fixes existing issues.


samples/demos/irb/Makefile
```diff
@@ -23,11 +23,11 @@ ercc:
 build: ercc
```
munapower, line 23: if when you call build you execute ercc then why do you have it in all?

mbrandenburger: good catch! Let me clean this up.

mbrandenburger: I've updated the Makefile. Note that we call the `all` target implicitly via a `make test` as triggered from the root directory. This means that, we do not run the irb `build` target when we call `make build` from the root, thus, we lazy build irb only when we test it.

PR #774, opened 2024-10-01
Add fixup check to gh actions

mbrandenburger:
This PR introduces another github action to check if a PR contains fixup commits. The check does not prevent us from merging but highlights the needs of squashing before merging a PR.

I've created a test PR that contains this commit and an empty fixup commit. As shown in #775, the gh action indicates that there is a fixup commit.

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -1,0 +1,20 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+
+name: Git checks
+
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


munapower:
I was reading the documentation for the github action that is added. So for small PRs it is ok, but for larger PRs, like the recent one where we upgraded the samples ... it suggests a process. Maybe we should reference that somewhere so our collaborators are aware?

mbrandenburger:
Thanks @munapower

The "squash and merge" option will always help to get rid of the fixup commits eventually. As you pointed out correctly, for larger PRs, where a "squash and merge" seems wrong, a process is required. We actually describe this process for FPC [here](https://docs.google.com/document/d/1sR7YV3pSYN3NEFiW-2fUqtpsJeJrpC0EWUVtEm0Blcg/edit?usp=sharing).

munapower (approved):

PR #781, opened 2024-10-23
Remove unused code

mbrandenburger:
This function may have issues with memory leaks. However, as it is not used at all, we can safely remove it.


bvavala:
> This function may have issues with memory leaks. However, as it is not used at all, we can safely remove it.

I guess it's a bit more than a memory leak! ;)

bvavala (approved):

PR #788, opened 2025-03-30
Fix unmarshal values

mbrandenburger:
This PR fixes an issue where a malicious key is used that might cause a null pointer dereferencing in
[unmarshal_values](https://github.com/hyperledger/fabric-private-chaincode/blob/88b7c21cc398ed7273d807976f6dffe5e69bd18c/ecc_enclave/enclave/shim.cpp#L195), which may crash the enclave.

### DIFF
--- a/ecc_enclave/enclave/shim.cpp
+++ b/ecc_enclave/enclave/shim.cpp
@@ -192,7 +192,17 @@ int unmarshal_values(
     {
         JSON_Object* pair = json_array_get_object(pairs, i);
         const char* key = json_object_get_string(pair, "key");
+        if (key == NULL)
+        {
+            LOG_ERROR("Shim: Cannot parse json: key does not exists");
+            return -1;
+        }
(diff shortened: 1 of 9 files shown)


chenchanglew:
Double-checked the pipeline — the new test case has been executed and passed successfully.
```
=== RUN   TestMustNotCrash  
    crash_test.go:17: Use channel: mychannel, chaincode ID: crash  
[fabsdk/core] 2025/03/30 19:10:01 UTC - cryptosuite.GetDefault -> INFO No default cryptosuite found, using default SW implementation  
--- PASS: TestMustNotCrash (4.09s)  
PASS  
ok  	github.com/hyperledger/fabric-private-chaincode/integration/crashtest/unmarshal_values/test	4.108s  
```

mwlik:
Hi @mbrandenburger, can I have a response on my report concerning this [bug](https://hackerone.com/bugs?report_id=2810471)? it's been months!

PR #836, opened 2025-09-02
feature(samples): confidential escrow example - LFX25

PsychoPunkSage:
This PR is meant to introduce a new example (using FPC & cc-tools) called `confidential escrow`.

NIL

TBA


samples/chaincode/confidential-escrow/chaincode/transactions/digitalAsset.go
```diff
@@ -0,0 +1,700 @@
...
+	},
+
+	Args: []transactions.Argument{
+		{
+			Tag:         "uuid",
+			Label:       "UUID",
+			Description: "UUID of the Digital Asset to read",
+			DataType:    "string",
+			Required:    true,
+		},
```
samuelvenzi, line 150: Instead of receiving the uuid section of the @key of the asset you could receive directly a reference to the asset. The argument could be:
```
		{
			Tag:         "digitalAsset",
			Label:       "Digital Asset",
			Description: "Digital Asset to read",
			DataType:    "->digitalAsset",
			Required:    true,
		},
```

This allows you to call the transaction with two options of body:
```
{
    "@assetType": "digitalAsset",
    "@key": "digitalAsset:[UUID]",
}
```
Or
```
{
    "@assetType": "digitalAsset",
    "symbol": "[TokenSymbol]",
}
```

This would the "CC-Tools" way of doing it.

samuelvenzi: Also, the argument would "arrive" in the transaction routine already validated by CC-Tools.

samples/chaincode/confidential-escrow/Makefile
```diff
@@ -0,0 +1,7 @@
+TOP = ../../..
```
mbrandenburger: I think we need to add the `confidential-escrow` in this makefile so all tests are triggered via the GH actions. 
https://github.com/hyperledger/fabric-private-chaincode/blob/main/samples/chaincode/Makefile#L9

go.mod
```diff
@@ -23,7 +23,8 @@ replace google.golang.org/grpc => google.golang.org/grpc v1.29.1
 require (
 	github.com/client9/misspell v0.3.4
 	github.com/golang/protobuf v1.5.3
-	github.com/hyperledger/fabric v1.4.0-rc1.0.20230405174026-695dd57e01c2
+	github.com/hyperledger-labs/cc-tools v1.0.2
```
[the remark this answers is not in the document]
mbrandenburger: It might be the case the the version upgrade made in root's go.mod cause the current integration tests to fail.

mbrandenburger (changes requested):
Thank you @PsychoPunkSage for working on this awesome example of FPC+CC-Tools! I highly appreciate all the work you put into this PR already.

I've left you a few comments below. Let's see if we can make the GH Actions tests happy and then get this merged.

samples/chaincode/confidential-escrow/test.sh
```diff
@@ -0,0 +1,500 @@
...
+    cd $FPC_PATH/samples/deployment/test-network
+    run_cmd "./setup.sh" "Running initial setup"
+}
+
+# Setup test network
+setup_network() {
+    log_info "=== SETTING UP NETWORK ==="
+    cd $FPC_PATH/samples/deployment/test-network/fabric-samples/test-network
+
+    run_cmd "./network.sh down" "Bringing down network"
```
munapower: I had to comment the network down and network up ca for it to work for me ....

samples/chaincode/confidential-escrow/README.md
```diff
@@ -0,0 +1,141 @@
...
+docker exec -it fpc-development-main /bin/bash
+cd samples/chaincode/confidential-escrow
+
+# Interactive menu
+./multi_user_dashboard.sh
+
+# Setup Bob using Option 4.
+```
+
+#### 4. In 3rd terminal window - Docker Environment (`Monitor`)
```
munapower: wouldn't this be the 4th terminal?

samples/chaincode/confidential-escrow/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### 5. Run Tests
+
+```bash
+# Run all basic tests
+./multi_user_dashboard.sh
+
+# Chosing Option 7.
+```
+
```
munapower: Maybe a section explaning that the you should first create a wallet for Alice and one for Bob, and that these activities can be monitored from the 5th terminal? and that after that either alice or bob can proceed to create an escrow? and what options to use to work with it? Maybe an explanation of the difference between release and refund?

samples/chaincode/confidential-escrow/Makefile
mbrandenburger: @PsychoPunkSage thanks for adding this.

I can see now that the CI runs indeed the `build` and `test` targets here. However, we currently do not have a `test` target defined.

I believe we should define a test target which calls your go unit tests.

For example:
```
test:
  go test ./...
```

`make test` should then run all go unit tests in all subdirectories. What do you think?

munapower (approved):

